PR #51, opened 2017-03-05
Close #50

### LINKED ISSUE #50: Memory leakage due to allocation and not freeing vars in initialise_qpoases()
I found a memory leakage problem today in the script file ocp_qp_condensing_qpoases.c, then Robin fixed: 
- Move the creator inside initialise_qpoases() to ocp_qp_condensing_qpoases(), so that the memory allocation can be freed. 
Lines: 334-338, file ocp_qp_condensing_qpoases.c.
To be move to the beginning of ocp_qp_condensing_qpoases()
- And then free the memory for primal_solution and dual_solution at the end of ocp_qp_condensing_qpoases().
Uncomment lines 310-311.

roversch:
Got rid of initialise_qpoases() as it became unnecessary. Everything inside ocp_qp_condensing_qpoases is free'd and malloc'd as it should be.


dkouzoup:
or better said, how it should be "for now" ;)

PR #64, opened 2017-04-17
Python interface for OCP NLP

roversch:
Finished a first draft implementation of the SWIG interface for OCP NLPs.


roversch:
Somehow, my push does not build, but the PR does. After trying to fix that for a few hours, I'm just gonna accept it and see if it works.

PR #78, opened 2017-05-10
Add code coverage to acados

roversch:
This complements the unit tests, as we can see now how much of our code (in percentages) is actually run.

For now, coverage should only check the acados library.


roversch:
codecov complains that there is no baseline to compare to, but that's normal as codecov is not in the main repo yet.

PR #83, opened 2017-05-31
Turn package install script into a complete installation script, flesh out the installation instructions in the readme

schlagenhauf:
The script allows to install acados on Ubuntu 16.04 (maybe more, please test).
The readme now contains more elaborate installation instructions with some hints for pitfalls during installation.
The environment variable setting is moved outside, to remove dependency to the install script.


roversch:
Thanks a lot, @schlagenhauf ! This however breaks our current Travis setup, as we first need to install all the dependencies, and then build `acados` for different build types: `Test`, `Debug`, `Release` etc. Moreover, Travis does not support all `apt-get` packages you listed; Travis works with a [whitelist](https://github.com/travis-ci/apt-package-whitelist/blob/master/ubuntu-trusty#L4108).

Also, I don't really see why it's better to do the `PATH` tinkering outside of the shell script.

Possible solution: have two scripts: one for Travis, one for users?

schlagenhauf:
Woops, totally disregarded Travis.
What is the canonical way of installing dependencies that are not in the whitelist? Alternatively two scripts don't sound too bad, as long as Travis still builds?

Regarding the paths: I guess i'm not seeing the intended usage, but since all `export` calls are not persistent, the user has to set the paths manually afterwards, right? In this case i'd say it provides some explicitness.

roversch:
> Are there good options to locally test if Travis builds?

just test if Travis works on your remote (`schlagenhauf/acados`) by pushing to it, no? You might have to activate Travis first. 

Alternatively, you can mimic a Travis build locally with `docker`: look [here](https://github.com/travis-ci/travis-ci/issues/6655) for a step-by-step procedure. Caveat: use `ci-garnet` as the [docker image](https://quay.io/repository/travisci/ci-garnet?tab=tags) instead. This takes some time to get to work though..

> since all export calls are not persistent, the user has to set the paths manually afterwards, right?

Not if you do `source install.sh` instead of `./install.sh`.

schlagenhauf:
I now restored the original install script as ".travis_install.sh", next to the new one. Travis builds again, so this might be an acceptable temporary solution.
The environment variables are again set inside the install script as before, in addition i left the statements in the "Getting Started" section for transparency.

PR #93, opened 2017-06-09
Update experimental scripts for ESP32

doanminhdang:
Some examples are collected for ESP32:
chen_nmpc_qpoases
mass_spring_hpmpc
mass_spring_qpoases
pendulum_partial_tightening (currently requires too large memory to fit in ESP32, mem_size = 8157608 bytes)


giaf:
@doanminhdang this PR is now impossible to merge, since it refers to a much older version of acados. I would close it, and you can send a new one if it is still relevant.

PR #94, opened 2017-06-13
CMake for qpOASES

roversch:
In order to provide Windows support


rienq:
Wow, this looks already amazing! Do I understand it correctly that you managed to make a CMake build system for qpoases and you are now in the progress of fixing the one for qpdunes, which depends on qpoases? Based on my limited experience from experimenting with Visual Studio, I agree with Robin that CMake is crucial for providing Windows support! 
However, what will we do with hpmpc and/or blasfeo then @giaf @zanellia?

zanellia:
I have used CMake for blasfeo and hpmpc before, but we decided with @giaf
not to go down that road. Howerber, we recently revived the discussion and,
if I got it right, he seemed to be willing to use CMake there too ;)

PR #107, opened 2017-08-09
HPIPM big merge; condensing complete rewrite

roversch:
acados with HPIPM runs all the test problems on my machine; lint and maybe some other travis errors may occur.


rienq:
A bit late for this question, but does this new implementation also allow one to eliminate the initial state just like the original implementation? (by default, this does not seem to be the case and it can strongly affect computation times)

giaf:
Well, the condensing itself is not removing x0. If it is there, it stays there. If there is not, there is not... So that would be another step to be performed before or after.

rienq:
Yea, I saw a chain mass example on how to eliminate x0 before condensing, but is there also an example somewhere on how to do it afterwards? (which would be important, for example in order to update an already condensed QP with a new initial state value)
This seems complicated by the strange order of variables at the moment in the condensed QP:
[u_0 u_1 ... u_{N-2} x_0 u_{N-1}]
--> is this really the intended order of the condensed variables (it seems to be based on the bound values) and why is that?

PR #129, opened 2017-09-12
'create' facilities in ocp_qp

roversch:
Instead of using the old 'allocate' functions, we now offer 'create' facilities, that calculate the size of a module and allocate it all at once.


roversch:
@rienq This PR removes all 'allocate' functions (rationale: better to offer only one way of doing something). Is that breaking your code?

rienq:
By my code, you mean PR #121? This is still broken because of a minor bug somewhere that I did not have the time yet to trace but I will try to do that soon.. I would also need to make a DAE based OCP example or better even a unit test.

dkouzoup:
Hey @rienq . I think @roversch was referring to your statement once that you use the allocate_qp function for your own projects. This pull request gets rid of this function as we discussed a few days ago.

rienq:
Hmm, I thought this allocate functionality would still be there some way but then based on the functions that return the memory size needed to store the struct and the functions that put the pointers in place?
About my code being broken: for sure, but that's why I will probably locally not be pulling acados for a while at least until the upcoming paper deadlines ;-)

dkouzoup:
yes of course it's there, just the syntax and the underlying memory allocation changed.

rienq:
It looks great! :+1:

PR #131, opened 2017-09-22
Minor changes to build system

nielsvd:
# Only changes to CMake-files
- casadi method getVersion() is deprecated, replace with version().
- allow unit test build with own version of casadi for octave, replace error for warning.
- fix timing, MEASURE_TIMINGS macro wasn't consistent among build of acados and examples (leading to undefined behavior because of type mismatch).


nielsvd:
Updated version-check syntax for casadi is apparently not compatible with casadi 3.1.1. This change is reverted and a new pull-request is issued.

PR #143, opened 2017-10-09
Add Travis build for Raspberry Pi

doanminhdang:
The build for ARM architecture in Raspberry Pi needs to be emulated with qemu, running on a docker container, because Travis only provides x86 machines.

Since it takes much more time to compile with qemu, the test build on Rpi is separated as a new build, with only support for Python. To save build time in order to fit in Travis limitation (50 minutes), steps in building swig and casadi on ARM Rpi was done before, and saved as a pullable docker image.

The Raspbian OS running with qemu introduced a segment fault when importing Casadi in Python (it could be something related to multi-threading stuff, but i don't know how to fix it). Hence, the Debian base image for Rpi was used instead (it is from Resin, they said to have some fix related to that segment fault issue, and Casadi on Python worked).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -82,7 +82,7 @@ script:
 # prepare qemu
 - docker run --rm --privileged multiarch/qemu-user-static:register --reset
 # build image
-- docker build -t rpi-raspbian .
+- docker build -t rpi-debian .
 # test image
 #- docker run rpi-raspbian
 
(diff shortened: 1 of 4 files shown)


doanminhdang:
I updated with the newest version of travis file. Test builds passed. One codecov failed, it reported decreasing coverage, might be because of the Dockerfile not covered by tests. :)

PR #164, opened 2017-11-27
fix casadi code generation (issue 162)

schoelst:
only very minor changes

### DIFF
--- a/swig/acados.i
+++ b/swig/acados.i
@@ -177,6 +177,7 @@ static std::string generate_vde_function(casadi::Function& model) {
     casadi::Function vde = casadi::Function(full_name, input, output);
     casadi::Dict opts;
     opts["with_header"] = casadi::GenericType(true);
+    opts["with_export"] = casadi::GenericType(false);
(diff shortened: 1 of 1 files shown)


swig/acados.i
```diff
@@ -177,6 +177,7 @@ static std::string generate_vde_function(casadi::Function& model) {
     casadi::Function vde = casadi::Function(full_name, input, output);
     casadi::Dict opts;
     opts["with_header"] = casadi::GenericType(true);
+    opts["with_export"] = casadi::GenericType(false);
```
jgillis: Out of curiosity; why do you write explicit type-conversion to ` casadi::GenericType`?

roversch: Good question, I was wondering the same. `true` should work as well, no?

schoelst: I remember getting an error at some point and just sticked to it.

roversch: I merged it and fixed it locally.

Just `true` instead of `GenericType(true)` works for me.

PR #168, opened 2017-11-29
added == assert after assign

zanellia:
Checking dimension of assigned memory region with '=='. @dimitris this does not  generate warnings when assert is switched off by CMake.


dkouzoup:
I think I have a partially satisfying solution ;)

dkouzoup:
merged in #172

PR #172, opened 2017-12-01
now possible to always have == assert (but size of structs not multiple of 64)

dkouzoup:


dkouzoup:
Only implemented for ocp_qp_in/out and ocp_nlp_in/out. Will stop there and at some point we can decide what's more important (structs multiples of 64 - note that currently it's not correctly implemented, or always asserts with ==)

PR #178, opened 2017-12-01
Irk integrator

chenyutao36:


rienq:
This looks really clean @chenyutao36 :+1: 
You might want to store, instead of recompute, the LU factorizations of the forward derivative propagation in order to do the backward derivative propagation though but I guess that was already on your mind as well? In addition, you can reuse the LU factorization of the previous integration step to do the Newton-type iterations for the next integration step and additionally save a lot of computations

chenyutao36:
Yes, I will add some options to make what you said possible. By now it's an
exact irk integrator that may somehow gurantee the convergence(although
theoretically not)

2017年12月1日 17:14，"rienq" <notifications@github.com>写道：

Issue #188, opened 2017-12-18
qpDUNES problems in new_core


bnovoselnik:
There are two main issues with qpDUNES that I've not been able to figure out so far.

Is there a way to properly cold start it?
In mass spring examples a low quality solution in acquired (max KKT residual is around 1e-2)...

bnovoselnik:
That's correct, the termination depends on stationarity tolerance. Unfortunately, nothing happens when you set it to 1e-12 (the default is 1e-6).

dkouzoup:
hmm, which is the KKT residual that is so high?

bnovoselnik:
It's res_g, I think that's the stationarity residual.

dkouzoup:
and can you check how many iterations does qpDUNES do on this example the first time the problem is solved?

bnovoselnik:
this is from mass_spring_qpdunes:

inf norm res: 3.253027e-02, 8.881784e-16, 4.440892e-16, 1.144447e-15

***************************************************************
total time             =         4.610 ms      =        100.00 %
condensing time =         0.000 ms      =          0.00 %
QP time              =         4.601 ms      =         99.80 %
interface time     =         0.009 ms      =          0.20 %
misc                   =         0.000 ms      =          0.00 %
QP iter               =                6
***************************************************************

dkouzoup:
sanity check: to check that our choice of stationarity tolerance reaches qpDUNES, set it to a negative value. Should do max number of iterations in that case. Does it? Also, if you set NREP to 2, how many iterations are needed in the second (warm-started) run? Just 1 or more?

bnovoselnik:
Well if I put it to negative value then it fails to solve it.

With NREP 2 this is the output:

inf norm res: 5.329071e-15, 8.881784e-15, 4.440892e-16, 1.145113e-15

total time      =         0.276 ms      =        100.00 %
condensing time =         0.000 ms      =          0.00 %
QP time         =         0.268 ms      =         97.10 %
interface time  =         0.007 ms      =          2.54 %
misc            =         0.001 ms      =          0.36 %
QP iter         =             1

dkouzoup:
What do you mean it fails to solve it? It should run e.g. 100 iterations and give a maximum iterations return value (which btw we also have in acados but not properly implemented) and the solution of the QP should have high accuracy, no?

bnovoselnik:
Well, I mean exactly what I said :)
If I put the tolerance to -1 I get the following message:

[...]

PR #189, opened 2017-12-18
Adds a number of iterations of a QP solver to qp_out

bnovoselnik:
Also some small experiments with qpDUNES interface (see issue #188 )

### DIFF
--- a/acados/utils/mem.c
+++ b/acados/utils/mem.c
@@ -32,7 +32,7 @@
 
 // #define _USE_VALGRIND_  // uncomment to bypass assignment and do new memory allocation instead
 
-#define _USE_MALLOC_  // acados_malloc = malloc / acados_malloc = calloc
+//#define _USE_MALLOC_  // acados_malloc = malloc / acados_malloc = calloc
 
 void make_int_multiple_of(int num, int *size) {
     *size = (*size + num - 1) / num * num;
--- a/examples/c/mass_spring_cond_solver.c
+++ b/examples/c/mass_spring_cond_solver.c
@@ -112,6 +112,7 @@ int main() {
         ocp_qp_info *info = (ocp_qp_info *)qp_out->misc;
         ocp_qp_info min_info;
         min_info.total_time = min_info.condensing_time = min_info.solve_QP_time = min_info.interface_time = 1e10;
+        min_info.num_iter = 10000;
 
         for (int rep = 0; rep < NREP; rep++) {
             acados_return = ocp_qp_condensing_solver(qp_in, qp_out, arg, mem, work);
@@ -120,6 +121,7 @@ int main() {
             if (info->condensing_time < min_info.condensing_time) min_info.condensing_time = info->condensing_time;
             if (info->solve_QP_time < min_info.solve_QP_time) min_info.solve_QP_time = info->solve_QP_time;
             if (info->interface_time < min_info.interface_time) min_info.interface_time = info->interface_time;
+            if (info->num_iter < min_info.num_iter) min_info.num_iter = info->num_iter;
         }
 
         double time = acados_toc(&timer)/NREP;
--- a/examples/c/mass_spring_cond_solver.c
+++ b/examples/c/mass_spring_cond_solver.c
@@ -112,7 +112,6 @@ int main() {
         ocp_qp_info *info = (ocp_qp_info *)qp_out->misc;
         ocp_qp_info min_info;
         min_info.total_time = min_info.condensing_time = min_info.solve_QP_time = min_info.interface_time = 1e10;
-        min_info.num_iter = 10000;
 
         for (int rep = 0; rep < NREP; rep++) {
             acados_return = ocp_qp_condensing_solver(qp_in, qp_out, arg, mem, work);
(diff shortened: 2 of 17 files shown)


acados/utils/mem.c
```diff
@@ -32,7 +32,7 @@
 
 // #define _USE_VALGRIND_  // uncomment to bypass assignment and do new memory allocation instead
 
-#define _USE_MALLOC_  // acados_malloc = malloc / acados_malloc = calloc
+//#define _USE_MALLOC_  // acados_malloc = malloc / acados_malloc = calloc
```
dkouzoup: ideally calloc should not be the default option. But I will fix that later

examples/c/mass_spring_cond_solver.c
```diff
@@ -120,6 +121,7 @@ int main() {
             if (info->condensing_time < min_info.condensing_time) min_info.condensing_time = info->condensing_time;
             if (info->solve_QP_time < min_info.solve_QP_time) min_info.solve_QP_time = info->solve_QP_time;
             if (info->interface_time < min_info.interface_time) min_info.interface_time = info->interface_time;
+            if (info->num_iter < min_info.num_iter) min_info.num_iter = info->num_iter;
```
dkouzoup: the number of iterations should not change between repetitions. Instead of having the minimum, I would instead do an assert that for i>1, niter[i] = niter[i-1] to avoid having the solver warm started

examples/c/mass_spring_condensing_hpipm.c
```diff
@@ -81,6 +81,7 @@ int main() {
     ocp_qp_info *info = (ocp_qp_info *)qp_out->misc;
     ocp_qp_info min_info;
     min_info.total_time = min_info.condensing_time = min_info.solve_QP_time = min_info.interface_time = 1e10;
+    min_info.num_iter = 100000;
```
dkouzoup: same goes here

examples/c/mass_spring_condensing_qore.c
```diff
@@ -85,6 +85,7 @@ int main() {
     ocp_qp_info *info = (ocp_qp_info *)qp_out->misc;
     ocp_qp_info min_info;
     min_info.total_time = min_info.condensing_time = min_info.solve_QP_time = min_info.interface_time = 1e10;
+    min_info.num_iter = 10000;
```
dkouzoup: and here

examples/c/mass_spring_condensing_qpoases.c
```diff
@@ -85,6 +85,7 @@ int main() {
     ocp_qp_info *info = (ocp_qp_info *)qp_out->misc;
     ocp_qp_info min_info;
     min_info.total_time = min_info.condensing_time = min_info.solve_QP_time = min_info.interface_time = 1e10;
+    min_info.num_iter = 10000;
```
dkouzoup: and here

examples/c/mass_spring_pcond_solver.c
```diff
@@ -39,7 +39,7 @@
 #define ELIMINATE_X0  // NOTE(dimitris): not supported by qpDUNES
 #endif
 
-#define NREP 100
+#define NREP 1
```
dkouzoup: set this back to 100

examples/c/mass_spring_qpdunes.c
```diff
@@ -29,11 +29,12 @@
 #include "acados/ocp_qp/ocp_qp_common_frontend.h"
 #include "acados/ocp_qp/ocp_qp_qpdunes.h"
 #include "acados/utils/create.h"
+#include "acados/utils/print.h"
 #include "acados/utils/timing.h"
 #include "acados/utils/types.h"
 #include "acados/utils/mem.h"
 
-#define NREP 100
+#define NREP 1
```
dkouzoup: this one also

examples/c/mass_spring_qpdunes.c
```diff
@@ -125,7 +126,7 @@ int main() {
     compute_ocp_qp_res_nrm_inf(qp_res, res);
     double max_res = 0.0;
     for (int ii = 0; ii < 4; ii++) max_res = (res[ii] > max_res) ? res[ii] : max_res;
-    assert(max_res <= 1e6*ACADOS_EPS && "The largest KKT residual greater than 1e6*ACADOS_EPS");
+    // assert(max_res <= 1e6*ACADOS_EPS && "The largest KKT residual greater than 1e6*ACADOS_EPS");
```
dkouzoup: does qpDUNES fail this one without partial condensing? I thought it was working

examples/c/mass_spring_qpdunes.c
```diff
@@ -145,8 +146,10 @@ int main() {
 
     printf("\ninf norm res: %e, %e, %e, %e\n", res[0], res[1], res[2], res[3]);
 
-    printf("\nSolution time for %d dual Newton iterations, averaged over %d runs: %5.2e seconds\n\n\n",
-        -1, NREP, time);
+    ocp_qp_info *info = (ocp_qp_info *)qp_out->misc;
+
+    print_ocp_qp_info(info);
```
dkouzoup: does this function also print the number of iterations now? Would be nice

bnovoselnik: it indeed prints out the number of iterations

examples/c/mass_spring_qpdunes.c
bnovoselnik: Yep, it fails it because the KKT residuals is 1e-2 (after one call to the solver). If you call it at least twice, the problem "magically" goes away (because qpDUNES is not cold started).

examples/c/mass_spring_qpdunes.c
bnovoselnik: I left this so that the qpDUNES problem remains apparent. See the comment below.

PR #193, opened 2018-01-12
can include/exclude qpDUNES and qore from Makefile

dkouzoup:
Minor changes to make consistent with CMAKE pending

### DIFF
--- a/examples/c/mass_spring_example.c
+++ b/examples/c/mass_spring_example.c
@@ -30,10 +30,14 @@
 #include <acados/ocp_qp/ocp_qp_full_condensing_solver.h>
 #include <acados/ocp_qp/ocp_qp_hpipm.h>
 #include <acados/ocp_qp/ocp_qp_hpmpc.h>
+#ifdef ACADOS_WITH_QPDUNES
 #include <acados/ocp_qp/ocp_qp_qpdunes.h>
+#endif
 #include <acados/dense_qp/dense_qp_hpipm.h>
 #include <acados/dense_qp/dense_qp_qpoases.h>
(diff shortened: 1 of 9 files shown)


examples/c/mass_spring_example.c
```diff
@@ -118,8 +126,9 @@ int main() {
                     ((ocp_qp_partial_condensing_args *)((ocp_qp_sparse_solver_args *)args)->pcond_args)->N2 = N2;
                     ((ocp_qp_hpmpc_args *)((ocp_qp_sparse_solver_args *)args)->solver_args)->max_iter = 30;
                     break;
+                #if ACADOS_WITH_QPDUNES
```
roversch: It seems that you have this flag 2 times now, once in ocp_qp.c and once here. Can we not write it such that we only need it in the C interface?

examples/c/mass_spring_example.c
```diff
@@ -128,14 +137,17 @@ int main() {
                     #endif
                     ((ocp_qp_partial_condensing_args *)((ocp_qp_sparse_solver_args *)args)->pcond_args)->N2 = N2;
                     break;
+                #endif
                 case FULL_CONDENSING_HPIPM:
                     printf("\nFull condensing + HPIPM:\n\n");
                     // default options
                     break;
+                #if ACADOS_WITH_QORE
```
roversch: This #ifdef is superfluous I think.

dkouzoup: This one indeed, but only because we don't change any options of qore, was more like for the shake of completeness

PR #195, opened 2018-01-16
Copy Arguments

nielsvd:
The c-interface follows the following workflow:
1. configure solver
2. create ``arguments`` struct and set options
3. create self-contained ``solver`` struct
4. solve problem using ``solver`` struct and a ``qp_in`` struct

In step 3 we need a to copy the ``args`` struct created by the user. But since the structure can have an arbitrary lay-out, solver specific ``copy_args`` routines must be written.

## TODO
* [x] Prototype
* [x] Implementation for non-hpipm dependent solvers
* [x] Implementation for hpipm dependent solvers
* [x] Test +  debug

After discussion 
* [x] Move copy_args function to interface.

## Open questions
* [x] How to copy arguments structs created by HPIPM?


nielsvd:
Ready to merge.

dkouzoup:
Great! I will check it out tomorrow.

dkouzoup:
I don't quite understand yet how this works. What's the type of *source? Can you also change some of the C examples in order to see how copy args becomes useful? Cheers.

nielsvd:
``void *source`` is a pointer to the args that are being copied.

Atm it is used in the interface to copy the user-created ``args`` into the self-contained solver struct. See:
https://github.com/nielsvd/acados/blob/707c2a14252a44e2a6e07f8406ad5fdf7a90d505/interfaces/acados_c/ocp_qp.c#L227

dkouzoup:
Ok I see, it's related to what you had as "hack" before. I am not 100% familiar with the c interface yet, so I don't get why the user can't directly create the struct itself with some sort of constructor. Or is this in the direction of using strings to define options? It also feels a bit counter intuitive to add a function pointer in core, that is only used in the c interface, can't we move it there somehow? 

ps: we can also have a Skype later if you prefer to discuss the details (or next week, it's not urgent).

nielsvd:
``copy_args`` has moved to the interface-folder.

examples/c/.gdb_history
```diff
@@ -0,0 +1,139 @@
...
+c
+p args->A_mat
+n
+p args->A_mat
+p args_
+p c_ptr
+s
+print source_
+print source_->A_mat
+quit
```
dkouzoup: @nielsvd, I thought we had completely removed qpDUNES-dev submodule from all possible places. Where did this come from?

nielsvd: @dkouzoup, it was still in my ``.git/index``-file and the ``external/qpDUNES-dev``-directory still existed. I therefore got errors when executing the ``git submodule update`` command. I had to do a ``git rm ..`` to make this error go away... I suppose it has *something* to do with the fact that my local git-repository was from before the ``qpDUNES-dev``-submodule was removed.

The ``.gitmodules`` file you link to is untouched:
https://github.com/acados/acados/blob/new_core/.gitmodules
https://github.com/nielsvd/acados/blob/new_core_copy_args/.gitmodules

dkouzoup: Yes, I had a quite similar mess when I removed it at some point, that's why I am asking. Actually, no submodules coming after could get updated because of this error. But I thought we had eliminated it from anywhere, so perhaps, it's just because you didn't have a fresh clone. Never mind ;)

dkouzoup:
merged

PR #198, opened 2018-01-24
Update/extend casadi wrapper.

nielsvd:
The ``casadi_wrapper`` currently used in the ``master`` branch to define NLP functions (constraints, objective) supports a larger class of generated code (including MX) than currently supported in the integrators. To this end, the idea is to merge ``casadi_wrapper`` and ``sim_casadi_wrapper``. TODOs include:
* [x] Update wrapper to ``new_core``-syntax.
* [x] Find interface that suits both the NLP definition and the integrators.
* [x] Update integrators to use new ``casadi_wrapper``.
* [x] Support parameters in integrators.
* [ ] Update NLP solver to use new ``casadi_wrapper``.


dkouzoup:
Sounds great! Just pointing out that @giaf is working on cleaning up the nlp interface atm (with blasfeo structs in nlp_in/nlp_out etc) so if there are major changes needed in nlp_common and ocp_nlp_gn_sqp let us know a bit in advance.

acados/sim/sim_erk_integrator.h
```diff
@@ -25,14 +25,50 @@ extern "C" {
...
+    double *b_vec;
+
+    bool sens_forw;
+    bool sens_adj;
+    bool sens_hess;
+
+    // Function pointers
+    external_function_fcn_ptrs *forward_vde;
+    external_function_fcn_ptrs *adjoint_vde;
+    external_function_fcn_ptrs *hess_vde;
```
dkouzoup: Isn't it more logical to have the model of the integrator in sim_in, as it was? I suppose you went for this way to have the nested implementation of submodules but that's not necessary to have (at least initially) for external functions I think. This also adds a lot of casadi-related code in the intergrators atm.

dkouzoup: On the other hand, it's only possible to change the model and use the same integrator only if you keep exactly the same dimensions, so I am not entirely convinced either what's the best place for the function pointers.

nielsvd: Indeed, it is constructed as such to maintain a nested structure of the ``args``, ``memory`` and ``workspace`` structs. Assuming that is the desired approach, it doesn't makes a lot of sense to put the function pointers in the input: Firstly, the in-struct is not an argument to the functions for creating args, memory, and workspace. Secondly, it would suggest to user it is possible to change function pointers on the fly, which isn't the case.

So far the integrator code is completely casadi-agnostic. It is aware that external functions are necessary to compute for instance the VDE sensitivities. The casadi_wrapper is just one (for now the only) tool to instantiate such external functions.

nielsvd:
Integrators are working with the new ``casadi_wrapper``. @giaf, @dkouzoup, @roversch : Should we immediately make use of the same ``external_function`` approach in ``ocp_qp_gn_sqp``? Atm, the PR branch is diverging from the implementation on blessed ``new_core``, since @giaf is working on a different ``external_function`` implementation.

dkouzoup:
Hey @nielsvd. Great that all integrators are working fine with your approach! After the discussion you had with @giaf during your visit, my understanding was that you both would prototype an external function the way you envisage it and then we will decide what makes more sense. Therefore, I wonder what is the point of having @giaf implementing it and the directly rejecting his approach? Shall we maybe open an issue and have a discussion there? i.e., where the two approaches differ, pros, cons etc? I do agree that we should make a decision before the end of this week though.

nielsvd:
Hi @dkouzoup,
see  #204.
Best, Niels

nielsvd:
See #204.

PR #203, opened 2018-02-05
Submodules-concept ACADOS-wide

nielsvd:
# Builds upon #198.
It's a small PR, but since it builds upon (the as of yet open) PR #198 it appears larger than it is.

The concept of submodules was initially introduced in #198 to support ``external_function``-modules inside the integrator.

**Main idea:** every module *can* but *doesn't need to* have submodules. Examples:
* https://github.com/nielsvd/acados/blob/00d5459d703620a426ef6f54eac3eb5fce2355e0/acados/sim/sim_erk_integrator.h#L89
but als:
* https://github.com/nielsvd/acados/blob/00d5459d703620a426ef6f54eac3eb5fce2355e0/acados/ocp_qp/ocp_qp_full_condensing_solver.h#L61

**Consequences:**
* Submodules need to be known at the time arguments are constructed, a pointer to the *module-specific* struct is passed (``NULL`` in case module doesn't have submodules): 
```c
int module_name_calculate_args_size(module name_dims *dims, void *submodules);
int module_name_assign_args(module_name_dims *dims, void *submodules, void *raw_memory);
```
* The ``ocp_qp_solver`` interface is slightly simplified, since general support for submodules makes the existence of either ``ocp_qp_solver_fcn_ptrs`` or ``ocp_qp_xcond_solver_fcn_ptrs`` redundant.


dkouzoup:
I like this idea in principle! I do not see yet how ocp_qp_solver_fcn_ptrs/ocp_qp_xcond_solver_fcn_ptrs are redundant though, because in this case we have two layers of submodules, i.e., for an SQP solver, the ocp_qp solver alone should be a submodule but not partial/full condensing, such that one can either directly put there the function pointers to the solver or the two-layered condensing+solver module. I know that at the moment this is not supported in the C interface, but it eventually should be, as the core also allows it. Am I right or did I understand your statement wrong?

dkouzoup:
or would you be able to treat this case using a NULL pointer for the condensing module?

nielsvd:
It's not the condensing routine that is considered as a submodule to the QP solver, but the QP solver is considered as a submodule to the *_condensing_solver.
* ``ocp_qp_full_condensing_solver`` has one submodule: a ``dense_qp``-solver. See [here](https://github.com/nielsvd/acados/blob/00d5459d703620a426ef6f54eac3eb5fce2355e0/acados/ocp_qp/ocp_qp_full_condensing_solver.h#L61).
* ``ocp_qp_sparse_solver`` (``ocp_qp_partial_condensing_solver`` may be a more appropriate name) has one submodule: an ``ocp_qp``-solver. See [here](https://github.com/nielsvd/acados/blob/00d5459d703620a426ef6f54eac3eb5fce2355e0/acados/ocp_qp/ocp_qp_sparse_solver.h#L60).
* The remaining solvers ``ocp_qp_hpipm``, ``ocp_qp_hpmpc``, ``ocp_qp_qpdunes`` have no submodules.

dkouzoup:
Yeah I am aware of that, since I wrote all of them :D (and I totally agree about the renaming)

My point above was that the sqp solver has a QP solver submodule which can either be function pointers to xcond_solver_ptrs (which has nested the functions to the actual qp solver) OR function pointers to an ocp_qp_solver_funs (which employs no condensing at all). Therefore, I do not see how your suggestion would make either of the two function pointer struct redundant

nielsvd:
We discussed already via Skype, for future reference here's the answer I gave.

If you consider that **all** modules have a submodules-struct pointer in their ``fcn_ptrs``-struct (as ``xcond_solver_ptrs`` has) you don't need to make the distinction between modules that have submodules and those that have not.

Of course, problems don't just vanish. In return the ``calculate_args_size()``- and ``assign_args()``-routines need to accept a possibly redundant argument. The solution here is that in case a certain module has no submodules, you can set the submodules-pointer to any value (such as ``NULL``) since it's not actually used.

nielsvd:
Implementation of recursive ``fcn_ptrs``-structs finished for all existing modules: ``dense_qp``, ``ocp_qp``, ``sim``, ``external_function``.

In addition, using the introduction of recursive ``fcn_ptrs``-structs, the c-interface now employs the ``fcn_ptrs`` struct as means to obtain the args- and solver- modules. The existing config-structs can be used to generate a ``fcn_ptrs``-struct, but in the future other module-design approaches can be explored.

PR #214, opened 2018-02-21
Delete cached ingored file

tmmsartor:
The header `chain_model.h` is auto-generated by CMake from the template` chain_model.h.in` .
It is ignored by git, but was still tracked.


dkouzoup:
@tmmsartor, the problem is that the Makefile system is actually using the .h file. So if it's not tracked, one should always first run the cmake to generate the file and then make. I guess that's also fine, do you maybe have a nicer solution to this?

giaf:
Yes the problem is that, when it is run from CMake, the absolute path is necessary...

A solution could be to replace the x0_nm2.txt and xN.nm2.txt (and all others) with .c files, containing only a static array of data giving the initial condition. In this way, the .c files could be included directly into the main .c file. And we would also avoid to read files :p

What do you think about that?

PR #219, opened 2018-03-02
restructured hpmpc interface

dkouzoup:
- fixed segfault
- avoided unnecessary copying of data
- using assign_memory properly
- broke partial tightening (to be fixed by @zanellia)

### DIFF
--- a/acados/ocp_qp/ocp_qp_hpmpc.c
+++ b/acados/ocp_qp/ocp_qp_hpmpc.c
@@ -41,17 +41,19 @@
 
 int ocp_qp_hpmpc_calculate_args_size(ocp_qp_dims *dims)
 {
-    int_t N = dims->N;
-    int_t size = sizeof(ocp_qp_hpmpc_args);
+    int N = dims->N;
+    int size = sizeof(ocp_qp_hpmpc_args);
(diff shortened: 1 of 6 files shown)


giaf:
When do you want to merge? Now or only after Andrea fixes partial tightening?

dkouzoup:
I'd leave @zanellia to decide. In principle there is quite some cleanup needed to have partial tightening working in a proper way so maybe it's better to remove it and directly implement it as a separate module based on hpipm?

zanellia:
It might be a good idea to have an implementation in hpmpc until we start
implementing this in hpipm. I can revive it in little time without
polluting too much the polished interface ;)

On Mon, Mar 5, 2018 at 9:05 AM, Dimitris Kouzoupis <notifications@github.com

dkouzoup:
Sounds good! Let's try today maybe since big merges are on their way ;)

dkouzoup:
@zanellia  status on this one? We should merge probably before merging with giaf_new_core, and the latter has to happen asap..

zanellia:
I am on it ;) We should merge https://github.com/dkouzoup/acados/pull/2 first.

dkouzoup:
addressed in 3561aafcdd8eab8331dfecbb8f1c5c506beb7912

PR #220, opened 2018-03-12
reverted changes in ocp_nlp example, added flag to handle Werror

dkouzoup:


dkouzoup:
@roversch do you compile with gcc-7? I hope I fixed all warnings and we can merge this PR

PR #226, opened 2018-03-19
check acados_c for sim & add ERK 1 and 2 & add sim_wt_model

giaf:


acados/sim/sim_erk_integrator.c
```diff
@@ -125,16 +131,59 @@ void sim_erk_opts_initialize_default(void *config_, sim_dims *dims, void *opts_)
 	opts->ns = 4; // ERK 4
     int ns = opts->ns;
 
-    assert(ns == 4 && "only number of stages = 4 implemented!");
+	assert( (ns==1 | ns==2 | ns==4) && "only number of stages = {1,2,4} implemented!");
```
dkouzoup: @giaf (future note): we should start using asserts only for things that should never happen and catch instead errors that are user-dependent. Asserts actually are deactivated in Release plus when they are triggered within Matlab, it gives a segfault which is not so nice

acados/sim/sim_erk_integrator.c
```diff
@@ -125,16 +131,59 @@ void sim_erk_opts_initialize_default(void *config_, sim_dims *dims, void *opts_)
...
+			break;
+		}
+		default:
+		{
+			// impossible
+			assert( (ns==1 | ns==2 | ns==4) && "only number of stages = {1,2,4} implemented!");
+		}
+	}
+
+    opts->num_steps = 2;
```
dkouzoup: Finally :)

acados/sim/sim_erk_integrator.c
giaf: It may not be that bad that all checks are gone in release ;)

Is there an example of catch somewhere to get inspired?

dkouzoup: Yes that's right, e.g. all the asserts on the memory allocation routines make sense to be active only in debug mode. But when it comes to possible mistakes by users we should find another way. At the moment we don't really do it the "right way" in many places, maybe we should introduced an acados_error_message function that may have different implementations depending on the CMAKE_BUILD_TYPE too. I'll look into qpDUNES that provides some nice error/warning functions.

PR #234, opened 2018-03-21
discrete model dynamics

giaf:


giaf:
@dkouzoup your casadi ERK4 works, but actually it looks quite a bit more accurate than acados ERK4, maybe there is a bug somewhere...

PR #240, opened 2018-03-26
implemented set_function for all integrators, cleaning up nonlinear chain example

dkouzoup:


giaf:
you don't need the cast to external_function_generic in model_set_function: you can always assign a void * to any pointer type

PR #251, opened 2018-03-30
Sim model functions renamings & added IRK to wt examples

giaf:
Also:
- some work on having a more systematic naming for the casadi functions in our model (done implicit models, some TODO in explicit models).
- Jonathan's work on merging jac for IRK. As of now there are only the merged ones, TODO add an opt to have the un-merged ones (for compatibility in case we have to work with black box models).

Can be merged.


dkouzoup:
I am a bit confused with the naming (trying to fix cmake):

I see:
wt_model_nx6/nx6p2/impl_**ode_jac**_u.o
wt_model_nx6/nx6p2/impl_**ode_fun_jac**_x_xdot.o
wt_model_nx3/impl_**ode_inc_J**_x_xdot_energy_balanced_model.o
wt_model_nx3/impl_**ode_J**_x_u_energy_balanced_model.o

what is exactly the new naming convention?

dkouzoup:
also I am not a huge fun of the order x - xdot - u but I guess it's too late now ;)

giaf:
The order x - xdot - u was the order used by Yutao in the IRK, and Jonathan left the same... Up to discussion!

The naming convention I would stick to is the one here
https://github.com/giaf/acados/blob/new_core/acados/sim/sim_common.h#L36
just used everywhere, also for the name of the files and casadi functions (which would also start with casadi_ ).
The one with _inc_ (for 'inclusive') and _J_ (for 'jacobian') was the one used by Jonathan, which apparently remains in some file names.

PR #254, opened 2018-04-05
external functions with argument {COLMAJ, BLASFEO_MAT, BLASFEO_VEC}

giaf:
And changed some of the external function calls.

The conventions are described in the file conventions_about_external_functions.txt, which HAS TO BE KEPT UPDATED when someone changes them!!!


giaf:
If everything is fine, I would merge it. 

eco4wind guys can manage that, and anyway if they use CasADi for the functions there is no difference in the argument type (besides the order of variables).

PR #255, opened 2018-04-07
Lifted integrators based on Yutao's integrators - WIP

zanellia:


zanellia:
I have fixed a big that caused the SQP iterations to converge to an inaccurate solution. I guess we could merge @giaf?

giaf:
Two questions:
1) what was the bug in the lifted IRK?
2) in update_vectors, I deleted these lines on purpose after fixing HPIPM. So either you forgot to submodule update, or I forgot to push HPIPM in acados ;)

dkouzoup:
2. is not a question :P

giaf:
the question is, what happened?

PR #259, opened 2018-04-10
soft constraints (WIP)

giaf:
changes so far:
- added QP example (mass spring)
- checked HPIPM interface (it works)
- made the other QPs return FAILURE instead of seg-faulting

### DIFF
--- a/acados/dense_qp/dense_qp_qore.c
+++ b/acados/dense_qp/dense_qp_qore.c
@@ -227,6 +227,14 @@ int dense_qp_qore(void *config_, dense_qp_in *qp_in, dense_qp_out *qp_out, void 
     int nvd = qp_in->dim->nv;
     int ngd = qp_in->dim->ng;
     int nbd = qp_in->dim->nb;
+    int nsd = qp_in->dim->ns;
+
+	if (nsd>0)
+	{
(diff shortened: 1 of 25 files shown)


nielsvd:
Out of curiosity, are there algorithmic/performance benefits in HPIPM declaring constraints as soft, over using extra control variables for slacks combined with hard constraints?

giaf:
Yes sure there are performance advantages, otherwise it would make no sense to make it ;)

Soft constraints are implemented using dedicated primal variables, the slacks. These do not enter the dynamic. 

In a tailored QP solver (like HPIPM for all dense_QP solver, OCP_QP solver and (partial) condensing), it is exploited the fact that their hessian of cost and jacobian of constraints are diagonal, and that they do not enter the dynamic. This means that they are eliminated before the Riccati recursion, which is the performance bottleneck, with complexity O(N(nu+nx)^3) ), and their processing has complexity O(N ns).

In an SQP, there is the additional advantage that, since they do not enter the dynamics, they do not increase complexity of simulation and sensitivity computation.

Overall, the soft constraints give negligible increase in complexity per iteration (they may increase the number of iterations in some cases).

giaf:
soft constraints are up and running, ready to be merged!

PR #261, opened 2018-04-16
added header file to wt_nx6p2 folder and modified wind_turbine_nmpc.c…

FreyJo:
… to clearify which model is used


giaf:
In theory the two headers should be identical, that's why I was using only one. Do they have any difference now?

FreyJo:
you are right, i just got confused when i read nx=8; in the example and saw
that the nx6_model header is included.
its just to make it more clear.

giaf:
ok good then I merge, it was just to be sure, I agree that it makes more clear

PR #262, opened 2018-04-17
fixed adjoint sensitivities in sim_irk_integrator

FreyJo:
removed the reuse of results from the forward sensitivity propagation in the adjoint sensitivity propagation - to increase readability and not have hidden bugs..


dkouzoup:
@FreyJo are we 100% sure that there is no practical case that one would need both forward and adjoint together? Because this PR makes the integrator slower in such a case.

giaf:
I guess exact Hessian is a case, but IRK is not implemented for that case, here no one knows how to do it, and when we will do it I would rather start from scratch than from a buggy implementation. So in that I agree with Jonathan.

A part from this case, if you have the full forward sensitivities, it is easy and cheap to compute the adjoint with a matrix-vector multiplication.
If you don't have the full forward sensitivities, it doesn't work anyway as of now all our integrators assume nf=nx+nu. So first I would fix that, and then be worried that we may need both forward and adjoint at the same time.

giaf:
@zanellia do you see any use of forward and adjoint sensitivities at the same time, maybe in any inexact scheme?

giaf:
About Make, since the guys prefer to be very conservative on CMake (i.e. BLASFEO generic, that would work also on non-Intel CPUs), I would leave the Makefile as the place to test on a larger user basis than myself some more top-notch features, like e.g. BLASFEO Hawell and now also the new AVX target in HPIPM ;)

dkouzoup:
I'm fine with HP and Sandy bridge on cmake btw, wasn't what we agreed upon a while back? Not sure when it changed exactly

giaf:
I guess you nailed it, it never changed ;)
https://github.com/acados/acados/blob/new_core/external/CMakeLists.txt#L26

dkouzoup:
so that was probably a "mental note" ;)

FreyJo:
thank you guys for the feedback.
I finished the cleaning so you can merge this one now :)

rienq:
Hey guys, the code looks great! :+1: 

And you would need a forward-backward propagation for exact Hessians, which is by the way described in my thesis and it has been implemented in ACADO code generation a long time ago:
http://cdn.syscop.de/publications/Quirynen2017.pdf
so this was indeed never implemented in acados but should not be too difficult, except for this issue of the reuse of the factorizations maybe.

Another example of when you would need this, is indeed for inexact Newton schemes where you have to generate a Jacobian approximation in the forward sweep and a gradient correction in the backward sweep, where you can again reuse factorizations like in the origin implementation in acados.

giaf:
Hey @rienq thanks for the clarification! An inexact Jacobian in the forward sweep would indeed be a case, since in this case the exact adjoint is still needed to compute the correct QP RHS.

By the way, you may have noticed that the code in new_core went through a major reimplementation, and I took the responsibility (and the shame :p ) of starting from a simple SQP scheme, commenting out the stuff related to inexact schemes. Once the code ripens and gets more stable, that stuff should be reimplemented, and it would be great to have your keeping an eye on the process ;)

giaf:
@FreyJo the code looks good, good job :)

giaf:
@FreyJo the memory of BLASFEO matrices has to be 64 bytes aligned. For a matrix, its memory size is multiple of 64 bytes, so if the first is aligned, the following are. For a vector, this is not true. So you should stack all matrices together, and all vectors together.

./sim_wt_model_nx3.out


sim solver: ERK

xn: 
-4.49402 -229.42317  0.00000 

S_forw_out: 
 0.10000  0.00000  0.00000  0.00000 -0.00119  0.00000  0.00000 
 0.10500  1.00000  0.00000  0.00000 -0.00003  0.00000  0.00000 
 0.00000  0.00000  1.00000  0.00000  0.00000  0.00000  0.00000 
time for 2748 simulation steps: 139.273000 ms (AD time: 137.215000 ms (98.66%))



sim solver: IRK
sim_wt_model_nx3.out: mem.c:195: assign_and_advance_blasfeo_dmat_mem: Assertion `(size_t)*ptr % 64 == 0 && "strmat not 64-byte aligned!"' failed.

zanellia:
@rienq just out of curiosity, what would the inexact scheme you have in
mind look like? A forward sweep to update only a few columns of the
Jacobian and a backward sweep for the gradient correction?

FreyJo:
@giaf  ah sure, I will fix this in the next hours

giaf:
@FreyJo I sent you a PR to fix the previous issue.

Now there is another one: the input corresponding to a blasfeo_dvec has to be a column vector in the CasADi model. Did you test all models after your changes? (and do you have assert disabled?)

./sim_wt_model_nx3.out


sim solver: ERK

xn: 
-4.49402 -229.42317  0.00000 

S_forw_out: 
 0.10000  0.00000  0.00000  0.00000 -0.00119  0.00000  0.00000 
 0.10500  1.00000  0.00000  0.00000 -0.00003  0.00000  0.00000 
 0.00000  0.00000  1.00000  0.00000  0.00000  0.00000  0.00000 
time for 2748 simulation steps: 135.475000 ms (AD time: 133.504000 ms (98.71%))



sim solver: IRK
sim_wt_model_nx3.out: external_function_generic.c:440: d_cvt_casadi_to_dvec_args: Assertion `sparsity_in[1]==1' failed.

FreyJo:
@giaf : no, you're right, I guess I should have tested them all.

> the input corresponding to a blasfeo_dvec has to be a column vector in the CasADi model.

So you think this bug should be fixed by regenerating the functions with a transposed input?
The generating function files, are not in the model folder, do you have/could you add them?

giaf:
@FreyJo I pushed in the blessed/new_core the updated models, now that example works (I am not sure I can push the casadi model for this example, so I have it locally on my PC).
The issue was that, in the casadi model, the outputs were stacked together, e.g. [impl_ode, jac_x, jac_u], that worked when they were all column-major, but now it has to be fixed.

Can you test and fix accordingly all other models?

examples/c/sim_wt_model_nx6.c
```diff
@@ -418,22 +418,58 @@ int main()
...
+		// debug adjoints
+		struct blasfeo_dmat S_forw_result;
+		struct blasfeo_dvec adjoint_seed;
+		struct blasfeo_dvec forw_times_seed;
+
+		int Sf_mem_size = blasfeo_memsize_dmat(nx, nx+nu);
+		int adj_s_mem_size = blasfeo_memsize_dvec(nx);
+		int check_mem_size = blasfeo_memsize_dvec(nx+nu);
+
+		void *Sf_mem = malloc(Sf_mem_size);
```
giaf: When you work with BLASFEO matrices, you need to provide aligned memory, e.g. using the function v_zeros_align in BLASFEO to allocate memory. Otherwise you get seg fault in certain architectures.

giaf:
Now all looks good.

@FreyJo just one question: why _t_ in the names J_t_x, J_t_u, J_t_xdot? They are the Jacobians, not the transposed of the Jacobians, right?

FreyJo:
@giaf  J_t_x, ... stands for temporary Jacobian w.r.t. x; renamed it to J_temp_x,...
Also removed unnecessary headers. This should be ready to merge now.

FreyJo:
@giaf now everything should be fine! can this be merged now?

giaf:
yep I merge now!

PR #264, opened 2018-04-18
SCQP

roversch:
Add SCQP facilities to acados

- add SCQP example
- new constraint model (called p)


giaf:
Hey Robin

- np is rather confusing in my opinion, I immediately thought about number of parameters... maybe there is some other letter still available?
- in your implementation they are clearly not independent, since having np>1 forces nh==1. What you did is build the p constraints _on_top_ of the nonlinear constraints, and not as another _independent_ kind of constraint. I really believe that this has to be changed. 
So, the p constraints have to be another kind of constraint, defined by the sizes np (number of nonlinear functions in p) and e.g. nqp (number of quadratic over p constraints, which in your current implementation is limited to 1). As a consequence, e.g. the total number of constraints is now ni=nb+ng+nh+nqp, and e.g. the DCt matrix in the QP has size (nu+nx)x(ng+nh+nqp). Clear enough?
- since p must not have to be a special case of h, you have to make sure that soft constraints work. Just look at how they work for g and h, and extend this to the new constraint type p.
- since p has to be independent from h, the external function p should have the needed number of outputs, and not more.
- ERK is limited by both accuracy and stability. Therefore, in my opinion a more conservative approach in the choice of num_steps should be preferred. The choice 2 has been discussed with Dimitris.
Especially, I think it is a very bad idea to change (and push) the default value in opts just for the sake of running a certain example, for the reason that the output of other examples may change as well. Actually I noted that something was going on because the output of other examples changed. This wrongly made me thing that there was a new bug somewhere.
About the fact that it was the easiest fix, well, we are the acados developers. There are two ways to change an opt (in decreasing order of difficulty)
1) https://github.com/acados/acados/blob/master/examples/c/wind_turbine_nmpc.c#L660
2) https://github.com/acados/acados/blob/master/examples/c/mass_spring_example.c#L170
Case 1) may be more "difficult" because it needs a cast. Case 2) is the way to go in the long run, and should be implemented for all modules. If something is not there yet, it would be a good idea to implement it: in the long run, everyone would benefit from it.
- managing make or cmake, or just adding updating the examples list as a new one is added are two completely different matters. I may be responsible for make, you may be responsible for cmake, but everyone adding a new example should be responsible to add it to the examples lists.

PR #267, opened 2018-04-19
Generalized integrator dimensions

FreyJo:
Each integrator has now its own dimension type.
The interface gives and takes this dimension structures as void pointers.

The following functions are now part of sim_solver_config:
 ```
    int (*dims_calculate_size)(void *config);
    void *(*dims_assign)(void *config, void *raw_memory);
    void (*extract_nx_nu)(void *dims_, int* nx, int* nu);
    void (*set_nx_nu)(void *dims_, int nx, int nu);
```
These functions have been implemented for all integrators in the master branch, and are set in sim_*integrator_name*_config_initialize_default.
Modified ocp_nlp_dynamics_cont to use this new interface.


giaf:
Hey @FreyJo 

everything seems to work fine :)

Fine also to pension the sim_crane_no_interface. We can revive it if we need it again.

I only have some concerns about naming: I would keep the convention used somewhere else in acados of having 'get' and 'set' functions, each getting/setting a single quantity at the time. E.g. see https://github.com/acados/acados/blob/master/acados/ocp_nlp/ocp_nlp_cost_nls.h#L119 (for get) and 
https://github.com/acados/acados/blob/master/acados/ocp_nlp/ocp_nlp_cost_nls.h#L121 (for set).

FreyJo:
@giaf sure I reworked the change sticking to this convention :)
The last commit was just to pass the ctest, the issue I created earlier today should still be solved properly.

giaf:
@FreyJo can you revert the last commit, and fix as stated in https://github.com/acados/acados/issues/269?

giaf:
Yep it is ok. Merging now!

giaf:
@FreyJo careful with chain_model.h, cmake modifies it with your local path (so you shouldn't add and commit it) https://github.com/acados/acados/commit/0ccb0c6a271a8cb82d7102aa206ae1dbd1142c90

PR #270, opened 2018-04-20
add TODO (opts->sens_adj = false)

zanellia:
no expansion step is performed if opts->sens_adj = false, i.e. multiple shooting + lifted integrators (Newton step) != direct collocation (Newton step) @giaf @dkouzoup @FreyJo

### DIFF
--- a/acados/sim/sim_lifted_irk_integrator.c
+++ b/acados/sim/sim_lifted_irk_integrator.c
@@ -209,7 +209,7 @@ void sim_lifted_irk_opts_initialize_default(void *config_, sim_dims *dims, void 
     opts->num_steps = 2;
     opts->num_forw_sens = dims->nx + dims->nu;
     opts->sens_forw = true;
-    opts->sens_adj = false;
+    opts->sens_adj = false; // TODO(andrea): this effectively disables the expansion step in lifted integrators !!!!!
     opts->sens_hess = false;
 
 	return;


rienq:
Instead of changing the default, it is probably cleaner to change the implementation itself though in order to do the expansions step regardless, i.e., 

you would remove the sens_adj from the step for the collocation variables here:
[](https://github.com/acados/acados/blob/master/acados/sim/sim_lifted_irk_integrator.c#L1108)
and you would add it to the Newton step for the multipliers here:
[](https://github.com/acados/acados/blob/master/acados/sim/sim_lifted_irk_integrator.c#L1145)

zanellia:
@rienq absolutely. I was suggesting to change the option, just pointing out that there is no expansion when opts->sens_adj = false ;)

zanellia:
I cannot merge, can somebody do it for me? It is just a comment, but an important one :P

dkouzoup:
weird, can you check how to give you this permission (if you like)?

zanellia:
I guess somebody with role ``owner'' has to upgrade my role to owner?

dkouzoup:
but Gianluca could accept PRs being a member (only made him an owner a couple of days ago) of the acados organization.. Or do you mean owner of the acados/acados project? Anyway I will make you an owner there anyway and hopefully that will solve the issue ;)

PR #276, opened 2018-04-24
Generalized Nonlinear Static Feedback (gnsf) integrator implemented

FreyJo:
Hey Team,

I implemented a structure exploiting IRK scheme using the Generalized Nonlinear Static Feedback (GNSF), which can be found in the attached PDF:
[gnsf_dynamic_system_structure](https://github.com/acados/acados/files/1958755/gnsf_dynamic_system_structure.pdf)

To show, how this integrator can be used, I created/modified following three examples:

- `sim_gnsf_crane.c` to have a relatively simple example, mainly used this for testing and developing
- `sim_wt_model_nx6.c` here one can see the differences using the GNSF integrator compared to other integrators: ERK, IRK, Lifted IRK
- `wind_turbine_nmpc.c` shows how to use the GNSF integrator within a NMPC scheme.

I did the necessary modifications in the (c)make files, in the interface (sim_common, sim_interface) and in the header files of the models and included the casadi generated C functions and some MATLAB files to generate these.
Documentation on the usage of external functions was added to convention_about_external_functions.txt.
Additional documentation of the integrator will be found in my master thesis.
Until then, please contact me for additional information.


acados/sim/sim_irk_integrator.c
```diff
@@ -567,20 +567,14 @@ int sim_irk(void *config_, sim_in *in, sim_out *out, void *opts_, void *mem_, vo
...
+                { 
                     acados_tic(&timer_ad);
                     model->impl_ode_fun->evaluate(model->impl_ode_fun, impl_ode_type_in, impl_ode_in, impl_ode_fun_type_out, impl_ode_fun_out);
                     timing_ad += acados_toc(&timer_ad);
-                    // fill in elements of rG
                 }
-
-                if ( (opts->jac_reuse & (ss==0) & (iter==0)) | (!opts->jac_reuse) )
-				{
-                    // compute the jacobian of implicit ode
```
dkouzoup: what are these changes about?

examples/c/sim_wt_model_nx6.c
```diff
@@ -336,6 +431,10 @@ int main()
 
 		int acados_return;
 
+		if (nss == 3) // for gnsf: perform precomputation
+			gnsf_precompute(gnsf_dim, in->model, opts, in->T);
```
dkouzoup: status on embedding this in another function?

examples/c/sim_wt_model_nx6.c
```diff
@@ -32,6 +32,8 @@
 #include "acados/sim/sim_erk_integrator.h"
 #include "acados/sim/sim_irk_integrator.h"
 #include "acados/sim/sim_lifted_irk_integrator.h"
+#include <acados/sim/sim_gnsf.h>
```
dkouzoup: detail: use " instead of < for consistency with other headers

dkouzoup: (in principle, one should only need the header of the c interface of the integrators)

examples/c/wind_turbine_nmpc.c
```diff
@@ -481,8 +528,9 @@ int main()
 	{
 		plan->nlp_dynamics[i] = CONTINUOUS_MODEL;
 //		plan->sim_solver_plan[i].sim_solver = ERK;
-		// plan->sim_solver_plan[i].sim_solver = IRK;
-		plan->sim_solver_plan[i].sim_solver = NEW_LIFTED_IRK;
+		plan->sim_solver_plan[i].sim_solver = IRK;
+		// plan->sim_solver_plan[i].sim_solver = NEW_LIFTED_IRK;
+		plan->sim_solver_plan[i].sim_solver = GNSF;
```
dkouzoup: only have _one_ uncommented sim plan to avoid confusion.

examples/c/wind_turbine_nmpc.c
```diff
@@ -667,8 +760,14 @@ int main()
 		}
 		else if (plan->sim_solver_plan[i].sim_solver == NEW_LIFTED_IRK)
 		{
-			sim_opts->ns = 4;
-			sim_opts->num_steps = 1;
+			sim_opts->ns = 5;
```
dkouzoup: why are those changed? Does the new_lifted_irk even work for num_steps > 1?

examples/c/wind_turbine_nmpc.c
```diff
@@ -667,8 +760,14 @@ int main()
...
+			sim_opts->num_steps = 4;
+		}
+		else if (plan->sim_solver_plan[i].sim_solver == GNSF)
+		{
+			sim_opts->ns = gnsf_num_stages;
+			sim_opts->num_steps = gnsf_num_steps;
+			sim_opts->jac_reuse = true;
```
dkouzoup: cool, did you already implemented that? ;)

examples/c/wind_turbine_nmpc.c
```diff
@@ -686,7 +785,6 @@ int main()
 		pcond_solver_opts->pcond_opts->N2 = 10;
 	}
 
-	// update after user-defined opts
```
dkouzoup: why removed?

examples/c/wind_turbine_nmpc.c
```diff
@@ -697,6 +795,34 @@ int main()
 
 	ocp_nlp_solver *solver = ocp_nlp_create(config, dims, nlp_opts);
 
+	/************************************************
+	* 	precomputation (after all options are set)
+	************************************************/
+
+	for (int i=0; i<NN; i++){
+		if (plan->sim_solver_plan[i].sim_solver == GNSF)
```
dkouzoup: Did you try to play around with using GNSF on only a subset of the shooting intervals and make sure that works too?

interfaces/acados_c/sim_interface.h
```diff
@@ -30,6 +30,7 @@ typedef enum {
     ERK,
     LIFTED_IRK,
     IRK,
+    GNSF,
     NEW_LIFTED_IRK
 } sim_solver_t;
```
dkouzoup: Try to fix the following warnings:

/Users/acados/acados/sim/sim_gnsf.c:219:9: warning: unused variable 'nuhat' [-Wunused-variable]
    int nuhat = dims->nuhat;
        ^
/Users/acados/acados/sim/sim_gnsf.c:524:9: warning: unused variable 'nx' [-Wunused-variable]
    int nx = dims->nx;
        ^
/Users/acados/acados/sim/sim_gnsf.c:591:13: warning: 12 enumeration values not handled in switch: 'EXPL_ODE_FUN',
      'EXPL_ODE_JAC', 'EXPL_ODE_HES'... [-Wswitch]
    switch (fun_type)
            ^
/Users/acados/acados/sim/sim_gnsf.c:591:13: note: add missing switch cases
    switch (fun_type)
            ^
/Users/acados/acados/sim/sim_gnsf.c:622:9: warning: unused variable 'nuhat' [-Wunused-variable]
    int nuhat      = dims->nuhat;
        ^
/Users/acados/acados/sim/sim_gnsf.c:689:9: warning: unused variable 'nuhat' [-Wunused-variable]
    int nuhat      = dims->nuhat;
        ^
/Users/acados/acados/sim/sim_gnsf.c:1016:12: warning: unused variable 'precomputation_time' [-Wunused-variable]
    double precomputation_time = acados_toc(&atimer) * 1000;
           ^
/Users/acados/acados/sim/sim_gnsf.c:1226:11: warning: unused variable 'sim_dim' [-Wunused-variable]
    void *sim_dim = (void *) dims;

acados/sim/sim_irk_integrator.c
FreyJo: i made this change to increase the readability of the code.
instead of having 2 complementary ifs; if+else is much more clear..

examples/c/sim_wt_model_nx6.c
FreyJo: no solution found yet. It has to be called after the `sim_gnsf_opts_update`. It could be made part of this function, if one adds another input type to the `opts_update` function in the `sim_solver_config`, namely `void *model` or `sim_in *in`.
In my opinion:
`    void (*opts_update) (void *config, void *dims, void *opts, void *model); `
would be a reasonable change.

examples/c/sim_wt_model_nx6.c
FreyJo: > in principle, one should only need the header of the c interface of the integrators
It is needed to set the additional dimensions, and to use the `gnsf_import_matrices` and `gnsf_precompute`.
I tried to use the interface as far as possible, but these things cannot be easily made part of it..

examples/c/wind_turbine_nmpc.c
FreyJo: As far as read into that, it should work. But I'll undo this change.

dkouzoup: The values you chose are anyway to high for practical applications. Ping @zanellia for what is currently supported in the new lifted IRK.

examples/c/wind_turbine_nmpc.c
FreyJo: jacobian reuse?
yes, i worked on this yesterday, see this commit [jac_reuse](0adb13badf9b5f55dce2ef3c948b396bce743469)

examples/c/wind_turbine_nmpc.c
FreyJo: replaced this with another comment a bit below
``	/************************************************
	* 	precomputation (after all options are set)
	************************************************/ ``

examples/c/wind_turbine_nmpc.c
FreyJo: nope. But should not be a problem. I can try this apart from the pull request.

interfaces/acados_c/sim_interface.h
FreyJo: all these warnings are fixed

examples/c/wind_turbine_nmpc.c
FreyJo: I tried it out. It works fine

giaf:
If you have a bit of patience I could have a look at that too. I am probably getting back to work at some point next week ;)

rienq:
Wow, @FreyJo, it looks like you have been very productive! :+1: 

How are the computational results compared to the preliminary results in the [IFAC paper](https://www.sciencedirect.com/science/article/pii/S2405896317300800)?

acados/sim/sim_gnsf.c
```diff
@@ -0,0 +1,1757 @@
...
+
+#include "acados/sim/sim_common.h"
+#include "acados/sim/sim_gnsf.h"
+#include "acados/sim/sim_irk_integrator.h"
+#include "acados/sim/sim_collocation_utils.h"
+
+// blasfeo
+#include "blasfeo/include/blasfeo_target.h"
+#include "blasfeo/include/blasfeo_common.h"
+#include "blasfeo/include/blasfeo_d_aux_ext_dep.h"
```
giaf: BLASFEO _ext_dep should not be used anywhere in acados core... so either this (and the i_aux_ext_dep) header is useless, or the code has to be fixed to not use these routines

acados/sim/sim_gnsf.c
```diff
@@ -0,0 +1,1757 @@
...
+#include "blasfeo/include/blasfeo_d_blas.h"
+#include "blasfeo/include/blasfeo_d_kernel.h"
```
giaf: I guess you don't use BLASFEO kernels yourself

acados/sim/sim_gnsf.c
```diff
@@ -0,0 +1,1757 @@
...
+    int ny = dims->ny;
+    int nuhat = dims->nuhat;
+    int exported_doubles = 0;
+
+    exported_doubles += (nx1 + nz) * (nx1 + nu + n_out + nx1+nz); // A, B, C, E;
+    exported_doubles += ny * (2*nx1 + nz); // L_x, L_xdot, L_z, L_u;
+    exported_doubles += nuhat * nu; // L_u;
+    exported_doubles += nx2*nx2; // A_LO;
+    // printf("exported_doubles %d= \n",exported_doubles);
+    double *export_in  = (double*) malloc(1*sizeof(double));
```
giaf: there should be no malloc and free in acados core

workspace and memory are here for this

acados/sim/sim_gnsf.c
```diff
@@ -0,0 +1,1757 @@
...
+}
+
+/************************************************
+* import functions
+************************************************/
+
+void gnsf_get_dims( sim_gnsf_dims *dims, casadi_function_t get_ints_fun)
+{
+    double *ints_out;
+    ints_out = (double*) calloc(10,sizeof(double));
```
giaf: there shoud be no m/calloc and free in acados core

acados/sim/sim_gnsf.c
```diff
@@ -0,0 +1,1757 @@
...
+
+    int nff = num_stages * n_out;
+    int nyy = num_stages * ny;
+    int nK1 = num_stages * nx1;
+    int nK2 = num_stages * nx2;
+    int nZ  = num_stages * nz;
+
+    // set up precomputation workspace
+    int pre_workspace_size = gnsf_pre_workspace_calculate_size(dims, opts);
+    void *pre_work_ = malloc(pre_workspace_size);
```
giaf: again no malloc

convention_about_external_functions.txt
```diff
@@ -248,3 +248,45 @@ outputs:
...
 2) [jac_xdot]   size: nx*nx     type: BLASFEO_DMAT
 3) [jac_u]      size: nx*nu     type: BLASFEO_DMAT
+
+/************************************************
+* sim_gnsf
+************************************************/
+
+/* nonlinearity function phi */
+
+external functions: phi_fun, phi_fun_jac_y, phi_jac_y_uhat
```
giaf: I would split these functions, and have separate inputs and outputs for each of them; I think it makes it more clear

giaf:
Can you explain all acronyms in gnsf_structure.pdf (e.g. NSF, DAE, LOS) and add it somewhere in the repo, e.g. in docs/reference?

FreyJo:
Hey @giaf 
thanks for your review,
The only thing left to fix are the mallocs.
I started with removing the most difficult one, the one in gnsf_precompute.
When this function is called in a MPC scheme, the pointer to the workspace of the integrator needs to be passed. I tried this as you can see here:
https://github.com/FreyJo/acados/blob/4521987988eefe541d871ef3212a3da930c74c94/examples/c/wind_turbine_nmpc.c#L840-L847
The problem is that the commented lines do not work, as `ocp_nlp_dynamics_cont_cast_workspace` is static and not part of the header file.
I solved it now by just passing the workspace of the ocp_nlp_solver, which large enough, and thus the example works fine. However, I do not think this problem is solved properly yet. Do you have a suggestion for a better fix?
I guess the cleanest would be to call it through the interface, by introducing a precompute function, which would need quite some changes of the interface and maybe is not worth it, as no one else is right now needing this.

dkouzoup:
actually @FreyJo  that's a good point. Since you know that nlp_workspace is always bigger than the integrator workspace you're good. That's kind of the purpose of the workspace anyway, to reuse it between modules. Just make sure that this size assumption always holds (e.g. if the total integrator workspace if is N*workspace of one integrator or if we take the maximum over the nodes)

FreyJo:
@dkouzoup cool, so this usage of the workspace is actually good style.
@giaf I removed all usages of (m/c)alloc.

examples/c/wind_turbine_nmpc.c
```diff
@@ -523,6 +577,37 @@ int main()
...
 	external_function_param_casadi_create_array(NN, impl_ode_fun_jac_x_xdot_u, np);
 	external_function_param_casadi_create_array(NN, impl_ode_jac_x_u, np);
+	// gnsf model
+	external_function_param_casadi_create_array(NN, phi_fun, np);
+	external_function_param_casadi_create_array(NN, phi_fun_jac_y, np);
+	external_function_param_casadi_create_array(NN, phi_jac_y_uhat, np);
+	external_function_param_casadi_create_array(NN, f_lo_jac_x1_x1dot_u_z, np);
+
+	int gnsf_num_stages = 5;
+	int gnsf_num_steps  = 2;
```
giaf: Choosing the same opts as NEW_LIFTED_IRK (ns=4 and num_steps=1), and even requiring on average less SQP iterations than it, in this example GNSF is quite slower that NEW_LIFTED_IRK (11 vs 8 ms on my PC, Sandy-Bridge BLASFEO target). Any idea why?

docs/reference/gnsf_dynamic_system_structure.tex
```diff
@@ -0,0 +1,62 @@
...
+
+\begin{align*}
+E \matr{c}{\dot{x}\uind{1} \\ z} & = A x\uind{1} + B u + C \phi(\lxdot\dot{x}\uind{1}+ \lx x\uind{1} + \lz z + \lu u)  	\\
+%	&= \bx x\uind{1} + \bu u + \bg g( x\uind{1}, z, u) 	\label{eq:dyn_nlsf_dae_b} \\
+\dot{x}\uind{2} &= \lo x\uind{2} + f(\dot{x}\uind{1}, x\uind{1}, z,u) 
+\end{align*}
+
+where we denote the states as $ x = \left[ {x\uind{1}}^\top, {x\uind{2}}^\top\right]^\top \in \R^{\nx} $, $ x\uind{1} \in \R^\nxone, x\uind{2}\in\R^\nxtwo $, the algebraic variables $ z \in \R^\nz $, the controls $ u \in \R^{n_u} $, the (nonlinear) functions $ \phi: \R^{\ninx + \ninu} \rightarrow \R^{\nout}, f: \R^{2\nxone + \nz + n_u} \rightarrow \R^{\nxtwo} $, which will be referred to as \defI{nonlinearity} and \defI{linear output function} respectively. Furthermore we use the model matrices $ \lxdot,\lx \in \R^{\ninx \times \nxone} $, $ \lz \in \R^{\ninx \times \nz} $, $ \lu \in \R^{\ninu \times n_u} $ (\defI{linear input matrices}) and the matrices $ A\in \R^{(\nxone + \nz) \times \nxone}$  $ B \in \R^{(\nxone + \nz) \times n_u},
```
giaf: What is the second term in the input size of $\phi$?

$L_u$ needs to have $n_y$ rows if you want to add $L_u u$ to the other terms forming $y$.

acados/sim/sim_gnsf.h
```diff
@@ -0,0 +1,253 @@
...
+    int ny;
+    int nuhat;
+} sim_gnsf_dims;
+
+
+typedef struct {
+    double *ff;
+    double *x0_1;
+    double *u_0;
+} gnsf_res_in;
```
giaf: This is used nowhere, right? Delete it?

acados/sim/sim_gnsf.h
```diff
@@ -0,0 +1,253 @@
...
+{
+    // external functions
+    // phi: nonlinearity function
+    external_function_generic *phi_fun;
+    external_function_generic *phi_fun_jac_y;
+    external_function_generic *phi_jac_y_uhat;
+    // f_lo: linear output function
+    external_function_generic *f_lo_fun_jac_x1_x1dot_u_z;
+
+    // precomputed matrices
```
giaf: Precomputed data should not belong to the model, but to the memory of the solver/integrator. 
@dkouzoup do you agree?

acados/sim/sim_gnsf.c
```diff
@@ -0,0 +1,1565 @@
...
+// void gnsf_import_precomputed(sim_gnsf_dims* dims, gnsf_model *model, casadi_function_t But_KK_YY_ZZ_LO_fun)
+// this function can be used instead of import_matrices + precompute (not working now, ATTENTION!!!)
+
+
+
+/************************************************
+* helpful functions
+************************************************/
+
+double minimum_of_doubles(double *x, int n){
```
giaf: this stuff doesn't really belong here... since it is used only in the examples, I would move it there

acados/sim/sim_gnsf.c
```diff
@@ -0,0 +1,1565 @@
...
+    {
+        if ( x[c] < min ) 
+        {
+           min = x[c];
+        }
+    }
+    return min;
+}
+
+void gnsf_neville(double *out, double xx, int n, double *x, double *Q){ // Neville scheme
```
giaf: since it is used only within this file, what about making it static?

acados/sim/sim_gnsf.h
```diff
@@ -0,0 +1,253 @@
...
+int sim_gnsf_memory_calculate_size(void *config, void *dims_, void *opts_);
+void *sim_gnsf_memory_assign(void *config, void *dims_, void *opts_, void *raw_memory);
+
+// interface
+void sim_gnsf_config_initialize_default(void *config_);
+
+// integrator
+int gnsf_simulate(void *config, sim_in *in, sim_out *out, void *opts, void *mem_, void *work_);
+
+// helpful functions
```
giaf: remove

acados/sim/sim_gnsf.h
```diff
@@ -0,0 +1,253 @@
...
+#include "blasfeo/include/blasfeo_d_blas.h"
+#include "blasfeo/include/blasfeo_d_kernel.h"
+#include "blasfeo/include/blasfeo_i_aux_ext_dep.h"
+#include "blasfeo/include/blasfeo_d_aux.h"
+
+typedef struct
+{
+    int nx;
+    int nu;
+    int num_stages;
```
giaf: Once the model contains only external functions and A, B, ..., L_x, ... and not pre-computed quantities, there is no need to have num_stages and num_steps in the dimensions, but they should be read from opts (same as for the other integrators). Right?

acados/sim/sim_gnsf.h
```diff
@@ -0,0 +1,253 @@
...
+    double *C;
+    double *E;
+    
+    double *L_x;
+    double *L_xdot;
+    double *L_z;
+    double *L_u;
+    double *A_LO;
+
+    // butcher table maybe remove
```
giaf: Again, after pre-computed quantities are gone from the model (into the memory), the butcher table should go as well (into the opts)

acados/sim/sim_gnsf.h
```diff
@@ -0,0 +1,253 @@
...
+int sim_gnsf_workspace_calculate_size(void *config, void *dims_, void *args);
+void *sim_gnsf_cast_workspace(void *config, void* dims_, void *raw_memory, void *args);
+int sim_gnsf_memory_calculate_size(void *config, void *dims_, void *opts_);
+void *sim_gnsf_memory_assign(void *config, void *dims_, void *opts_, void *raw_memory);
+
+// interface
+void sim_gnsf_config_initialize_default(void *config_);
+
+// integrator
+int gnsf_simulate(void *config, sim_in *in, sim_out *out, void *opts, void *mem_, void *work_);
```
giaf: rename sim_gnsf for naming consistency with the other integrators

docs/reference/gnsf_dynamic_system_structure.tex
FreyJo: thanks, forgot to update this...
I split the input of phi at some point, such that the integrator does not compute the derivatives w.r.t. the controls within the Newton iteration.
Basically, I just replaced the last `+` in line 54 with a `,`

acados/sim/sim_gnsf.c
FreyJo: I talked with @dkouzoup and moved it to `utils/math.c`.

FreyJo:
@giaf thanks again for the feedback!
I changed everything accordingly.

examples/c/sim_wt_model_nx6.c
```diff
@@ -316,13 +316,13 @@ int main()
 				gnsf_dim->nu = nu;
 				gnsf_dim->nx1= nx;
 				gnsf_dim->nx2= 0;
-				gnsf_dim->ny = nx;//nx + nu;
-				gnsf_dim->nuhat = 2;//nx + nu;
+				gnsf_dim->ny = 5;
+				gnsf_dim->nuhat = 0;
```
giaf: What does the change mean? Why was it larger before?

giaf:
@FreyJo a column-vector in CasADi must correspond to a BLASFEO_DVEC in the models.

 ===  USING SOLVER NUMBER 3 === 


sim solver: gnsf, ns=8, num_steps=3
sim_wt_model_nx6.out: external_function_generic.c:249: d_cvt_dvec_to_casadi: Assertion `sparsity_out[1]==1' failed.
Makefile:171: recipe for target 'run_sim_wt_model_nx6' failed

examples/c/sim_wt_model_nx6.c
FreyJo: Before, I was using y=x and uhat = u; which means that phi takes the whole state and control vector as an input.
To reduce the size of the matrix vector multiplications, I reformulated the model. Using only the entries of these vectors needed to evaluate phi.
Therefore I needed to make some changes in the generating .m files, which are also included in this commit.

FreyJo:
@giaf I tested the example only with cmake, which worked somehow.
Using make I get the same error you posted.
Debugging this, I found out, that it crashes, in the `sim_gnsf` in the function call `model->phi_fun_jac_y->evaluate` in `external_function_param_casadi_wrapper` in the first loop for `ii=1`, which means when converting the second input of this external function from dvec to casadi.
This input vector is of length 0, thus transposing it in the generating .m file will not fix the problem.
Do you have any idea what could be the problem? And why it works with cmake but not with make?

FreyJo:
@giaf  I tried different stuff to avoid this error, e.g. setting an empty column vector `uhat = MX.sym('uhat',1,0);` as input for the CasADi function.. But this does not change the sparsity_in[1].. 

It seems to be a general problem that one cannot use a blasfeo d_vec of length zero as in-/output of a external casadi function.

I guess one has to modify the casadi external function interface such that for input/output arrays of size 0 this assert is skipped. Thus I suggest to modify the function
`d_cvt_dvec_to_casadi(struct blasfeo_dvec *in, double *out, int *sparsity_out)`
to also take the size of the argument and look like:
`d_cvt_dvec_to_casadi(struct blasfeo_dvec *in, double *out, int *sparsity_out, int arg_size)`
and modify the assert to something like:
`assert( (sparsity_out[1]==1) |  (arg_size == 0));`

What do you think?

giaf:
ok everything looks good now

PR #278, opened 2018-04-26
Quickfix qpoases

zanellia:
Fixed buggy hotstart


dkouzoup:
out of curiosity, what was the bug?

PR #287, opened 2018-05-07
Travis status green

roversch:
This is a revival of Travis for acados

- manual formatting with clang-format
- all code should lint now
- unit tests are reintroduced (only ocp_qp for now)

Idea is to keep status green for all future PRs into master


dkouzoup:
I agree with 1 & 2, and don't care so much about 3! @giaf other than that, did you check the rest of the PR too to make sure we did not introduce any bugs in the code while fixing conflicts? If not we should certainly delegate it to someone asap as we should merge relatively quickly so huge PRs

dkouzoup:
Btw, is there a way to "undo" line breaks with clang format once we e.g. undo the spacings suggested above? Or one can only do this manually?

roversch:
@giaf I adapted the code to comply to your comment no. 2

For the others: you seem to care much more about style than I do. I have no automated way of changing these things. If you know one, feel free to tell me. I'm not going to go manually through all files and remove spaces.

Please don't forget that there is a discrepancy between lint and clang-format. I use clang-format to correct most linting errors automatically. But lint does not care if you write `nx[ii+1]` or nx[ii + 1]`, so this degree of freedom is already there.

@dkouzoup: yes, clang-format automatically 'squeezes' the lines to fit on the give line length.

dkouzoup:
Ok thanks for the update. Then I personally suggest:
(a) you de-activate the automatic spacing nx[ii+1] --> nx[ii + 1] from clang format to avoid such changes in the future (and eventually someone can help to massively revert those if necessary)
(b) remove the lint rule on the spacing of comment such that travis does not break, but it can be updated with clang format once in a while

does this make sense?

giaf:
Yep it makes sense! Totally agree with @dkouzoup 

I was not suggesting to undo changes (well, it is still great that you could easily undo the cast) but to remove the corresponding linting error (and de-activate the automatic spacing as @dkouzoup suggested) to have this extra bit of freedom from now on.

Once you tell me that all changes are done, I still want to check that all examples work before merging.

dkouzoup:
Unfortunately disabling the auto-spacing between the operators does not seem to be possible in clang format.. We could alternatively decide which spacing between operators we definitely not want and ask our web-hiwis to write a python script or something, to be executed after clang-format.

PR #290, opened 2018-05-10
Proof-of-concept of xxx_calculate_size_self and xxx_assign_self, to facilitate bottom-up approach

giaf:
This PR contains a proof-of-concept (for the ocp_nlp_dims) of a feature facilitating the bottom-up approach that @roversch plans to use in the python and matlab interfaces.

The idea to increase code-reuse between core and these interfaces it to create for each module xxx_calculate_size_self and xxx_assign_self, which handle all memory needed by the structure itself, excluding sub-modules. The functions xxx_calculate_size and xxx_assign internally call their _self counterparts, avoiding code duplication.

No changes are needed in the current acados solvers, C interface or examples, as they can still use the non_self functions as before.


giaf:
This PR also contains a little cleanup of unused junk.

PR #296, opened 2018-05-17
Intermediate PR for interfaces

roversch:
This PR introduces:

- bugfixes in c++ interface
- working ocp_nlp example in Python and MATLAB

This is an intermediate PR to keep the size in check.


dkouzoup:
Hey @roversch, I suggest to:

(a) make sure that when no SWIG_MATLAB/SWIG_PYTHON is ON, cmake does _not_ look for casadi.

(b) when either is ON and the right casadi version is in the "external" subfolder it should be automatically detected without setting any environment variables (using a e.g. a hint in cmake)

PR #297, opened 2018-05-18
Setters and getters for bounds in ocp_nlp

bnovoselnik:
Added setters and getters for bounds in ocp_nlp c interface.
Two versions of each function are provided, for example:

```cpp
int nlp_set_constraint_bounds_in_stage_with_offset(ocp_nlp_dims *dims, ocp_nlp_in *in, int stage, const char *identifier, int numel, int offset, double *values);
```

and 

```cpp
int nlp_set_constraint_bounds_in_stage(ocp_nlp_dims *dims, ocp_nlp_in *in, int stage, const char *identifier, double *values);
```

Both functions will insert values from the array ```values``` into stage constraints given by ```stage```. String ```identifier``` specifies which bounds are being inserted (e.g. "lb", "ub", "lg", "ug" and so on). If ```numel``` is greater than zero then only numel values will be inserted, otherwise the number of elements to be inserted is determined from dims. Offset can control the starting index of the insertion. For example, if one wants to insert only 3 bounds starting from bound 4 (i.e. one wants to update only bounds 4, 5 and 6), then numel=3 and offset=4. 

For convenience,  the second function (the one without offset) assumes that one wants to insert all the bounds (the number of bounds is determined from dims).

Getters are pretty much the same.


dkouzoup:
Thanks for the quick implementation @bnovoselnik! Two remarks:
- I would not allow the behavior: 0 in numel copies everything. 0 should indeed copy nothing, same as in blasfeo routines for example. Perhaps you can have the same behavior if numel is negative?
- What happens if the given offset and numel grow larger than dims? Do you make sure that you do not write outside the bounds of the current constraint, or even worse outside d?

bnovoselnik:
> I would not allow the behavior: 0 in numel copies everything. 0 should indeed copy nothing, same as in blasfeo routines for example. Perhaps you can have the same behavior if numel is negative?

OK, that's an easy fix.

> What happens if the given offset and numel grow larger than dims? Do you make sure that you do not write outside the bounds of the current constraint, or even worse outside d?

I don't check that, but I can do it.

bnovoselnik:
@giaf No problems, I'll leave it like this for now and we can discuss it on Monday :)

bnovoselnik:
I reworked setter and getter functions according to your suggestions.

```cpp
int nlp_bounds_bgh_set(ocp_nlp_constraints_bgh_dims *dims, ocp_nlp_constraints_bgh_model *model,
                                       const char *identifier, double *values);

int nlp_bounds_bgh_get(ocp_nlp_constraints_bgh_dims *dims, ocp_nlp_constraints_bgh_model *model,
                                       const char *identifier, double *values);
```

giaf:
Looks good

dkouzoup:
@roversch is this something you can work with?

PR #301, opened 2018-05-24
fixed wt nx3 inputs

FreyJo:
When setting the inputs, the forward seed was overwritten...
I guess this was not intended @giaf


dkouzoup:
@FreyJo this PR is also part of the larger PR on DAE's right? We can close it I guess. It was not merged automatically because of the conflict.

FreyJo:
yes, this was also part of #302

PR #302, opened 2018-05-25
Support for index-1 DAEs

FreyJo:
Hey guys,

this pull request enables acados to simulate index-1 DAEs with the `sim_irk` and the `sim_gnsf` integrators. Therefore some changes in the interface of the integrators were necessary.
The changes in a nutshell:

- ODEs are now treated as a special case with `nz = 0`
- all integrators have `dims->nz` now and corresponding get/set functions
- added `double *zn` to `sim_out` and `bool output_z` to `sim_rk_opts`, which indicates if the algebraic variables are just treated internally (`false`) or if they should be part of the output (`true`)
- Note: output value `zn` always corresponds to the beginning of simulation
- added `double *S_algebraic` to `sim_out` and `bool sens_algebraic` to `sim_rk_opts`, output of `S_algebraic` should contain the sensitivities of the output value `zn` w.r.t. `in->x, in->u`
- All currently existing examples are still working, without applying any changes
- Note: The DAE should come in the following (fully implicit) form: F(x, xdot, u, z) = 0
- Thus the external functions of true DAEs, i.e. `nz > 0`, take z as additional argument
- For IRK: The external functions, that generate also jacobians have jac_z as additional output (for true DAEs, i.e. `nz > 0`).
- I renamed the external functions within IRK only internally, such that the interface is not affected, changing
  * `impl_ode_jac_x_xdot_u` to `impl_ode_jac_x_xdot_u_z`
  * `impl_ode_fun_jac_x_xdot` to `impl_ode_fun_jac_x_xdot_z`
- added `sim_test_dae` with toy example; which runs IRK and GNSF with many different option configurations, computes the error of the output values (simulation, sensitivities (forward, adjoint, algebraic) and algebraic variables z)  w.r.t. a reference solution.
- errors of GNSF & IRK are super close -> as it should be :tada: 

For details look into the generating file of the test example, and the `sim_test_dae`.


FreyJo:
Hey Team,
From my side this PR is finished. I updated the initial description.
Any Feedback is welcome!

FreyJo:
@giaf thanks for your review
I forgot to mention the following quite important implementation decision:  The output value `zn` is computed by evaluating the polynomial through (t_i, z_i) (i=1,...,num_stages) at 0 (using Nevilles algorithm). The other option would be (if a higher accuracy is desired) to perform Newton iterations on the system F(x0, * , u, * ) = 0 to get more exact values of z(t_0) (and k(t_0)). 
As the second option is computationally much more expensive, we went for the first.
This could be additionally implemented, introducing an option `bool newton_zn`.
In the test we use Gauß-Legendre IRK methods, which have order P=2*num_stages, at the end point,  while the internal stages of any collocation method have an order of at least num_stages, [see Dissertation of Rien Quirynen p.62]. Thus it is not surprising that the accuracy of z and the algebraic sensitivities is lower.

On the other remarks
- In the `sim_test_*` first a reference solution is computed using many steps, a high order. The errors that get computed now are the absolute errors with the infinity norm. I wanted to change this soon to be the relative error.
- I will add an example of an inverted pendulum DAE model in my next pull request

giaf:
Ok let's merge it now
@FreyJo please address @dkouzoup last comment (about tol_z) in you next PR too.

giaf:
@FreyJo good job btw ;)

FreyJo:
@giaf thanks :) will do!

dkouzoup:
Thanks for the feedback @zanellia! @FreyJo are you on it?

FreyJo:
@dkouzoup yes!
The changes are part of my new PR: https://github.com/acados/acados/pull/305
@zanellia I implemented these changes, but number 2:
I think it is better to leave IMPL_ODE_FUN_JAC_X_XDOT ( instead of IMPL_ODE_FUN_JAC_X_XDOT_Z) for now. As these strings are also used to set the model functions for (new_)lifted_irk, which do not support the z (yet).

PR #303, opened 2018-06-11
SQP RTI

giaf:
This PR is a test for the NLP infrastructure, to check that it is possible to add another NLP solver.

The SQP RTI removes the SQP loop and the computation of NLP residuals. Therefore, by default it does not compute adjoint of dynamics and constraints, possibly slightly reducing computational time.

It would be possible to also include opts to avoid the computation of Lagrange multiplier in the QP sub-problems too, but this would requires some more changes, also to the HPIPM version (while Branimir is working with another new one (but not as new) in his PR on soft constraints.

What now? Clean and merge this PR or keep working on it?


dkouzoup:
I suggest to pull first the latest master with the fix of @bnovoselnik for travis (clang still failing) and make sure that at least the gcc builds succeed before we merge

giaf:
done

dkouzoup:
I had to restart the build but this bug is not related to this PR ;)

PR #310, opened 2018-06-20
Soft constraints support for interfaces other than HPIPM

bnovoselnik:
- provides functions to reformulate dense QPs and OCP QPs with soft constraints from structure exploiting formulation (used by HPIPM) to a more general formulation that can be used by other solvers as well
- slacks are explicitly stacked with optimization variables and problem matrices are extended accordingly
- QPOASES and QORE interfaces in ACADOS have been extended to use this reformulation when presented with a dense QP with soft constraints (tested in mass spring and wind turbine examples)
- OCP QP reformulation is not that usefull maybe, because QPDUNES cannot use it as it is (if there are soft constraints on states in the last stage for example) and HPMPC has some numerical troubles that I was not able to overcome (by tweaking HPMPC parameters). At this point, only HPIPM can solve this reformulated OCP (I used it just to test the correctness of the reformulation) but of course HPIPM doesn't need it :)

Should be noted: this PR uses a slightly newer version of HPIPM (bba890a).


examples/c/chain_model/chain_model.h
```diff
@@ -7,6 +7,24 @@
 extern "C" {
 #endif
 
+#define X0_NM2_FILE "/home/bnovoselnik/Software/acados/examples/c/chain_model/x0_nm2.txt"
```
giaf: This file got changed my cmake... We should find a better way of doing that, everyone added it to a commit at least once :p

bnovoselnik: I agree, it's really annoying :)

giaf:
@bnovoselnik travis is failing for all compilers, stating that d_ocp_qp has no member 'm'. So it looks like an older version of HPIPM is used.

bnovoselnik:
@giaf Yep, that's true. The problem is that the version bba890a also lacks one small bug fix from a newer version of HPIPM (4ee5ab4) which is relevant for soft constraints. I wanted to bump ACADOS to the newest version of HPIPM (which would include this bug fix) but I couldn't do it because of a problem with redeclaration of enums in HPIPM (see this issue I opened not loo long ago: https://github.com/giaf/hpipm/issues/30)

giaf:
Ok, yesterday I was working a bit on   the enums issue in HPIPM, today I didn't have any time. I'll try to finish it tomorrow.

PR #313, opened 2018-07-20
add parallelization to multiple shooting

giaf:
work in progress

### DIFF
--- a/acados/ocp_nlp/ocp_nlp_sqp.h
+++ b/acados/ocp_nlp/ocp_nlp_sqp.h
@@ -37,15 +37,16 @@ extern "C" {
 
 typedef struct
 {
-    int maxIter;
-    double min_res_g;
-    double min_res_b;
-    double min_res_d;
-    double min_res_m;
     void *qp_solver_opts;
     void **dynamics;     // dynamics_opts
     void **cost;         // cost_opts
     void **constraints;  // constraints_opts
+    double min_res_g;
+    double min_res_b;
+    double min_res_d;
+    double min_res_m;
+    int maxIter;
+	int reuse_workspace;
 } ocp_nlp_sqp_opts;
 
 //
--- a/acados/ocp_nlp/ocp_nlp_sqp.h
+++ b/acados/ocp_nlp/ocp_nlp_sqp.h
@@ -79,7 +79,10 @@ typedef struct
     ocp_nlp_memory *nlp_mem;
 
     int sqp_iter;
-    int dummy;  // NOTE(dimitris): make struct size multiple of 8
+
+	double time_qp_sol;
+	double time_lin;
+	double time_tot;
 } ocp_nlp_sqp_memory;
 
 //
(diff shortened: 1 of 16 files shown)


acados/ocp_nlp/ocp_nlp_sqp.h
```diff
@@ -77,7 +79,10 @@ typedef struct
     ocp_nlp_memory *nlp_mem;
 
     int sqp_iter;
-    int dummy;  // NOTE(dimitris): make struct size multiple of 8
```
dkouzoup: Is this not relevant any more? I vaguely remember that we should have an even number of ints in our structs to make sure that we are 8-byte aligned every += sizeof(struct) at the beginning of the assign routines.

examples/c/nonlinear_chain_ocp_nlp.c
```diff
@@ -1121,7 +1121,7 @@ int main()
 	ocp_nlp_solver_plan *plan = ocp_nlp_plan_create(NN);
 
 	// TODO(dimitris): not necessarily GN, depends on cost module
```
dkouzoup: maybe remove this comment now?

examples/c/wind_turbine_nmpc.c
```diff
@@ -518,10 +518,10 @@ int main()
...
 	{
 		plan->nlp_dynamics[i] = CONTINUOUS_MODEL;
-        // plan->sim_solver_plan[i].sim_solver = ERK;
-		// plan->sim_solver_plan[i].sim_solver = IRK;
-		// plan->sim_solver_plan[i].sim_solver = NEW_LIFTED_IRK;
-		plan->sim_solver_plan[i].sim_solver = GNSF;
+//		plan->sim_solver_plan[i].sim_solver = ERK;
+		plan->sim_solver_plan[i].sim_solver = IRK;
+//		plan->sim_solver_plan[i].sim_solver = NEW_LIFTED_IRK;
+//		plan->sim_solver_plan[i].sim_solver = GNSF;
```
dkouzoup: someone should clean up the examples since we test anyway all exotic combinations in the unit tests now..

acados/ocp_nlp/ocp_nlp_sqp.h
roversch: Good point. Don't we anyway always align to 8 bytes ?

dkouzoup: I think not every +sizeof(struct), coz it would be a bit tedious, but I haven't checked that since quite a while now..

examples/c/nonlinear_chain_ocp_nlp.c
roversch: 👍

dkouzoup:
Looks great! Can you either lint the code or delegate it to someone in order to merge soon?

examples/c/wind_turbine_nmpc.c
giaf: I would keep this stuff, this is how I found there was another bug in GNSF+BLASFEO

I don't trust unit tests alone...

giaf:
This PR was not intended to be reviewed yet, I just made the PR so that Jonathan could check what was going on with GNSF... ;)

giaf:
For example, it only works with make, someone should figure out how to do OpenMP with cmake

PR #319, opened 2018-08-13
Irk hessians

FreyJo:
Hey guys,

I mainly concerned myself with second order sensitivity propagation for IRK in the last time.
Here a short summary on what this PR includes:

- `sim_common`: `sim_out` now has enough memory to store full hessian, makes evaluation & packing much easier
- `external_function_generic`: fixed bug that occurred using `blasfeo_dmat` as input
- `sim_irk` implemented Forward Backward Hessian Propagation, added conditional memory allocation, some cleaning
- added Matlab Function to export `C` code for explicit models
- regenerated pendulum model using this file -> with recent naming convention & made use of these instead of old C files
- wrote `sim_test_hessian` where the pendulum model is used to compare the hessian results of `sim_irk` and `sim_erk`.
- Additionally a Hessian is generated using finite differences to check if the result is reasonable.
- deleted some old code
- `sim_erk` adjoint scheme contained some logical errors in if statements - fixed them as far as I could.
- There is definitely still a bug in the ERK Hessian propagation. The results are not completely off, but I do not know how to fix it, the whole file lacks in comments and should be reworked..


rienq:
Hi Jonathan, it's a great idea to compare the Hessian propagation result with an approximation based on finite differences: if you implement finite differences on top of an exact first order sensitivity propagation (either forward or backward), then you should obtain a Hessian result with a precision that equals the square root of the machine precision, i.e., around 1e-7 -- 1e-8 in double precision arithmetics.
Is that the case at the moment?

FreyJo:
Hi Rien,
in my test case it is a bit less accurate, the relative error I get is 1.653029e-05.
This should be fine, because the accuracy is very dependent on the nonlinearity within the model, I guess..

FreyJo:
@dkouzoup I dont know. I also can not compile the current master: I get
`/home/oj/Git/acados/examples/c/nonlinear_chain_ocp_nlp.c:1141:39: error: ‘PARTIAL_CONDENSING_OOQP’ undeclared (first use in this function)
  plan->ocp_qp_solver_plan.qp_solver = PARTIAL_CONDENSING_OOQP;`

dkouzoup:
I guess that's because you don't have OOQP locally. Indeed the examples should not be based on software that the user might not have. Does travis fail too though?

FreyJo:
still weird that the current master passed travis though :thinking:

dkouzoup:
The master branch has always OOQP available. Pull requests can only access OOQP if it happens to be still cached. I guess that's the reason. @giaf should have a look at this PR if you think it's ready to merge.

dkouzoup:
ps: pull again master and git submodule update on this one

FreyJo:
should be ready to be merged now!

dkouzoup:
@FreyJo, happy conflict solving :(

FreyJo:
@dkouzoup @giaf  seems fine to me now! Let me know, if something has to be fixed before merging

dkouzoup:
Cool, that was quick!

acados/ocp_nlp/ocp_nlp_dynamics_cont.c
```diff
@@ -527,24 +527,13 @@ void ocp_nlp_dynamics_cont_update_qp_matrices(void *config_, void *dims_, void *
...
-                hess_index++;
-            }
-        }
+        // unpack d*_d2u
+        blasfeo_pack_dmat(nu, nu, &work->sim_out->S_hess[(nx+nu)*nx + nx], nx + nu,
+                                     &mem->hes, 0, 0);
+        // unpack d*_dux: mem-hess: nx x nu
+        blasfeo_pack_dmat(nx, nu, &work->sim_out->S_hess[(nx + nu)*nx], nx + nu, &mem->hes, nu, 0);
+        // unpack d*_d2x
+        blasfeo_pack_dmat(nx, nx, &work->sim_out->S_hess[0], nx + nu, &mem->hes, nu, nu);
```
dkouzoup: Thanks for simplifying this! I did not check the details, but do you have a (general enough) test problem to confirm that the two pieces of code do the same thing? Also what about @giaf's comment on directly adding this to RSQrq instead? Maybe clarify with @roversch before we forget)

acados/sim/sim_erk_integrator.c
```diff
@@ -748,8 +755,21 @@ int sim_erk(void *config_, sim_in *in, sim_out *out, void *opts_, void *mem_, vo
         // store hessian
         if (opts->sens_hess)
         {
-            for (i = 0; i < nhess; i++)
-                S_hess_out[i] = adj_tmp[nx + nu + i];
+            // former line for tridiagonal export was
```
dkouzoup: Not sure I understand what's going on here but does this code produce the same results on the experiments of the acados paper that @roversch has tested? Is there anything else left to clarify?

acados/sim/sim_gnsf.c
```diff
@@ -1659,7 +1659,7 @@ int sim_gnsf(void *config, sim_in *in, sim_out *out, void *args, void *mem_, voi
             blasfeo_dgemv_n(nyy, nff, 1.0, &YYf, 0, 0, &ff_traj[ss], 0, 1.0, &yyss, nyy * ss,
                             &yy_traj[ss], 0);
 
-            if ((opts->jac_reuse & (ss == 0) & (iter == 0)) | (!opts->jac_reuse))
+            if ((opts->jac_reuse && (ss == 0) && (iter == 0)) || (!opts->jac_reuse))
```
dkouzoup: Ask @giaf why he likes & better ;)

acados/sim/sim_irk_integrator.h
```diff
@@ -40,48 +40,79 @@ typedef struct
 typedef struct
 {
     /* external functions */
```
dkouzoup: Good job! Do you check in the sim tests that all possible combinations of options (compute_hessian, compute_forward, etc) work properly?

acados/utils/external_function_generic.c
```diff
@@ -178,7 +178,7 @@ static void d_cvt_dmat_to_casadi(struct blasfeo_dmat *in, double *out, int *spar
         {
             for (idx = idxcol[jj]; idx != idxcol[jj + 1]; idx++)
             {
-                ptr[0] = BLASFEO_DMATEL(in, row[idx], nrow);
+                ptr[0] = BLASFEO_DMATEL(in, row[idx], jj);
```
dkouzoup: was this a bug?

interfaces/matlab/sim/generate_c_code_explicit_ode.m
```diff
@@ -0,0 +1,81 @@
+%   This file is part of acados.
```
dkouzoup: Do I get it right that this replaces all the custom scripts we had to generate every model?? Finally!! 👍 Maybe a short wiki entry with an example on how to use it?

acados/ocp_nlp/ocp_nlp_dynamics_cont.c
FreyJo: I tested with the `pendulum_sqp` example, in cooperation with @roversch 
@dkouzoup seems I missed that comment...
anyway the interface to hessians is cleaner now, and it can be used in the higher level moduls.

acados/sim/sim_erk_integrator.c
FreyJo: yes it does in combination with the code you commented on above.

acados/sim/sim_gnsf.c
FreyJo: I guess its not so high performing, but seems save..

acados/utils/external_function_generic.c
FreyJo: yes

acados/sim/sim_irk_integrator.h
FreyJo: Nope.
As I stated initially, ERK has some strange behaviour. I just checked again `sim_test_hessian`, trying to also turn forward and/or adjoint sensitivities off. And found:
1. `sens_hess = true, sens_adj = true, sens_forw = false`: produces adjoints that are off
2. `sens_hess = true, sens_adj = false, sens_forw =true `: produces Segmentation fault
3. `sens_hess = true, sens_adj = false, sens_forw = false `: produces Segmentation fault
4. `sens_hess = true, sens_adj = false, sens_forw = false `:  works but hessian and adjoint result are a bit off..

Only 4. is active in the test :arrow_right: test passes
I could instead remove ERK from the test and test all the combinations for IRK..

Maybe I should write an issue for that..

@roversch I would be really curious how your scheme for the paper works if you use the IRK hessians instead of the one from ERK!

interfaces/matlab/sim/generate_c_code_explicit_ode.m
FreyJo: Could also be extended to automatically generate a header file and the lines to use `external_function_casadi_create`, but might be complementary work to the Matlab interface...
@dkouzoup I dont have rights for the wiki..

dkouzoup: Sorry, I just invited you. Actually at the moment the code generator is in C++ and we are thinking for various reasons to move it to matlab. Therefore, if you implement it it will for sure be very useful!

acados/sim/sim_irk_integrator.h
dkouzoup: So IRK passes all 4? If yes I would say: run 1-4 for IRK, only 4 for ERK and more importantly: open an issue to fix 1-3 for ERK.

acados/ocp_nlp/ocp_nlp_dynamics_cont.c
giaf: The two pieces of code do not look like doing the same thing: in the first it looks like only the data of the lower-triangular matrices are stored (contiguously) in S_hess, while in the second it looks like the entire matrices are stored in S_hess (also with the unreferenced upper triangular part). Which one is the correct one? How is data stored in S_hess?

acados/sim/sim_common.c
```diff
@@ -133,15 +133,15 @@ int sim_out_calculate_size(void *config_, void *dims)
...
     size += sizeof(sim_info);
 
-    size += nx * sizeof(double);                   // xn
-    size += nx * NF * sizeof(double);              // S_forw
-    size += (nx + nu) * sizeof(double);            // S_adj
-    size += ((NF + 1) * NF / 2) * sizeof(double);  // S_hess
+    size += nx * sizeof(double);                // xn
+    size += nx * NF * sizeof(double);           // S_forw
+    size += (nx + nu) * sizeof(double);         // S_adj
+    size += (NF * NF) * sizeof(double);         // S_hess
```
giaf: Is this the answer to the above comment? The way data is stored in S_hess has changed?

giaf:
It looks good to me, I asked and answered myself a question, @dkouzoup do you want to have the honor to merge?

dkouzoup:
Thanks giaf! What about GNSF 2.0?

acados/sim/sim_erk_integrator.c
```diff
@@ -748,8 +755,21 @@ int sim_erk(void *config_, sim_in *in, sim_out *out, void *opts_, void *mem_, vo
...
+            //            for (i = 0; i < nhess; i++) S_hess_out[i] = adj_tmp[nx + nu + i];
+            int count_upper = 0;
+            for (int j = 0; j < nx + nu; j++) {
+                for (int i = 0; i < nx + nu; i++){
+                    // S_hess_out[i] = adj_tmp[nx + nu + i];
+                    if ( i >= j ) {
```
giaf: Why not have the inner for loop for(int i=j; i<nx+nu; i++) and remove the if ?

Also, in theory only the lower triangular part of the hessian is needed, so ti should not even be necessary to copy it on the upper triangular part.

giaf: Add a reference @FreyJo

Issue #320, opened 2018-08-16
GPL lincense in BLASFEO and HPIPMP: linking issues


giaf:
As of yesterday afternoon, both BLASFEO and HPIPM are GPLv3 + CE, instead of LGPLv2.1.

Therefore, their libraries can not be included into the libacados.a library (since acados is LGPL), but they should be stand alone. Linking is permitted at the instant of creating the executable.

Besides these legal issues, as BLASFEO and HPIPM developer I am against the option of having a single libacados.a incorporating libblasfeo.a and libhpipm.a (and libhpmpc.a too), as this removes the last bit of visibility (after git submodules take away most of it).

giaf:
So, if BLASFEO and HPIPM are not already stand alone libraries in both make and cmake build systems, this should be fixed before upgrading their version used in acados.

FreyJo:
I guess this issue is solved/outdated given the latest license changes to BSD-2 of acados, HPIPM and soon also BLASFEO.
https://github.com/acados/acados/pull/401

PR #322, opened 2018-08-21
Green build 🥇 

dkouzoup:
- updated to latest blasfeo/hpipm
- building shared libraries separately not to have any licensing issues (see #320) 
- temporarily disabled one travis command, that should be restored before merging

Let's see what appveyor says ;)


.travis.yml
```diff
@@ -81,20 +81,20 @@ install:
...
   - rm -r -f build
   - cmake -E make_directory build
-  - cmake -E chdir build cmake -DCMAKE_BUILD_TYPE=Debug -DUNIT_TESTS=ON -DSWIG_MATLAB=$SWIG_MATLAB -DMatlab_ROOT_DIR=$MATLAB_ROOT -DSWIG_PYTHON=$SWIG_PYTHON -DCOVERAGE=$COVERAGE -DCMAKE_INSTALL_PREFIX=$HOME/local ..
+  - cmake -E chdir build cmake -DCMAKE_BUILD_TYPE=Debug -DUNIT_TESTS=ON -DSWIG_MATLAB=$SWIG_MATLAB -DMatlab_ROOT_DIR=$MATLAB_ROOT -DSWIG_PYTHON=$SWIG_PYTHON -DBUILD_SHARED_LIBS=ON -DCOVERAGE=$COVERAGE -DCMAKE_INSTALL_PREFIX=$HOME/local ..
   - export CXX="$CXX -std=c++11" # needed for octave
   - cmake --build build
   - cmake -E chdir build ctest --output-on-failure
   - cmake --build build --target install
-  - python -c "import acados" # check installation
+  # - python -c "import acados" # check installation
```
dkouzoup: @roversch any idea why this fails when BUILD_SHARED_LIBS is ON? The python tests pass without problems.

roversch: It might be a problem with the acados-SWIG shared lib looking for the other shared libs (acados, blasfeo etc). Before, it was all in one big shared lib, so it didn't have to look for anything (apart from casadi)

roversch:
we can ditch it if you want, nobody is using it I guess.

.travis.yml
dkouzoup: ok, but if ctest succeeds (also works for me locally), doesn't that mean that the libs are found properly anyway?

dkouzoup: I guess not all the paths are set properly?

dkouzoup:
It's nice to have this percentage actually, but atm I am obsessed with seeing a green light! Is it possible to write the percentage but not giving a red light when the PR reduces the coverage? (btw, I don't even see how my three lines cause 2% reduction)

.travis.yml
dkouzoup: is this check really necessary or shall we merge regardless?

PR #325, opened 2018-08-24
GNSF 2.0

FreyJo:
Hey guys,

this PR additionally to #319 contains the following changes:

- I generalized the GNSF structure further, allowing more linear dependencies in the linear output type system and also introducing a split of z, such that it can be partly modelled by the linear output system part. Now the GNSF structure reads as
[this: gnsf_structure.pdf](https://github.com/acados/acados/files/2318322/gnsf_structure.pdf)
- Updated sim_gnsf and Matlab structure detection accordingly
- Regenerated GNSF models accordingly
- Cleaned the automatic model generation in Matlab and documented it in the wiki:
[How to: Dynamic System Models and Integrators in acados](https://github.com/acados/acados/wiki/How-to:-Dynamic-System-Models-and-Integrators-in-acados)

My intention is that this will be merged instead of #319 .


dkouzoup:
@FreyJo Can you pull the latest master here to only see the latest differences? (At least I think it works that way)

test/sim/sim_test_dae.cpp
```diff
@@ -72,8 +72,8 @@ double sim_solver_tolerance_dae(std::string const& inString)
 
 double sim_solver_tolerance_algebraic_dae(std::string const& inString)
 {
-    if (inString == "IRK")  return 1e-4;
-    if (inString == "GNSF") return 1e-4;
+    if (inString == "IRK")  return 3e-4;
+    if (inString == "GNSF") return 3e-4;
```
dkouzoup: why?

FreyJo: I changed the example used in the test, adding more nonlinearities to the algebraic equations. Thus, I had to adjust the accuracy required to pass the test.

FreyJo:
Yes for me this is ready to be merged! Also fixed the memory thing for which you pinged me yesterday.

test/sim/sim_test_dae.cpp
dkouzoup: Ok! Is it not possible to change the integrator options to give higher accuracy solutions? Even 1e-4 seems a bit high for a test.

dkouzoup:
Yeah I noticed! Which will force me to resolve a conflict but I forgive you :P

I will leave the merge to @giaf then, maybe you can meet in person tomorrow.

test/sim/sim_test_dae.cpp
FreyJo: Not really.
The output for algebraic variables gets computed without performing al Newton-iteration, but by evaluating the polynomial through the values of the first stage.
Other things could be implemented, but I think are not priority...

giaf:
@dkouzoup today we discussed a little about initialization of the integrator variables. Currently the initial values for xdot and z are passed in sim_in. But this does not scale to GNSF, since this requires the initialization of other quantities which are only present in GNSF (and therefore not in sin_in, whose definition is common to all integrators).

What would be a good solution here? Maybe move all initializations to memory, which is different for each integrator?

dkouzoup:
Hmmm good question. In my opinion, I would keep the initialization of variables where the variables belong. If we have xdot and z in sim_in, I would initialize them there, and variables that are located in the GNSF memory would be init. upon the creation of the GNSF memory, no?

FreyJo:
@dkouzoup The integration variables can not be initialized when the memory is created (with anything else but zeros), as no values enter the memory_create function, which has the following input:
`void *config, void *dims_, void *opts_, void *raw_memory`
Also, xdot and z are now part of sim_in to implement initialization for sim_irk, but I am not sure if they naturally belong there. Or if initialization should always happen through the memory. For multiple runs it should happen through the memory internally I guess, but the in the first call I am not sure.
Anyway I am fine with how it is now -  and could soon add this internal initialization, copying the last result for the integration variables in the memory and take it as new initial guess.

FreyJo:
Yes just wanted to clarify this.
Lets talk about this next week.
Anyway, I think the state now is quite good, especially I dont expect the GNSF stuff that this is PR is about to change. Thus I suggest to merge it.

giaf:
ok let's do it then

PR #326, opened 2018-08-27
Travis CI from scratch | Now with osx support

rqelibari:
# Travis CI from scratch
The current version of `.travis.yml` has gotten too messy and lacks any structure. Attempts to extend it with multi-os testing turned out very time consuming. Even after having brought some structuring in.
I hence decided to start from scratch and give it a solid structure.

# v.1.0-osx
## macOS build
This version now supports builds for macOS. I already tagged the current commit such that Travis created a new release for me (you will need to set GITHUB_OAUTH_TOKEN in Travis CI repository settings).

## Cache
I make use of Travis CI caching ability such that Travis does not need to rebuild all the code all the time. With this setup I was able to reduce the CI runtime as down as 2 minutes and ~15~ 08 seconds (sic!) 👍
travis-build

Linux and Windows (cross-compile) version will follow. I hope you guys like it as much as I do!


dkouzoup:
Hey @rqelibari, that's a great start. Thanks a lot for the initiative! I have the following remarks/suggestions:
- since we have a mac build now, why not running the unit tests too? You can check how we do it for linux, should be a one-liner.
- since we currently have version 0.1 in acados, I would suggest something like 0.2 instead of 1.0
- let us know when the linux functionality is restored in order to merge this PR, we do not need to wait for the windows cross-compilation.

PR #338, opened 2018-11-12
removed unused jacobian functions in chain_model and dependencies

FreyJo:
forgot to remove those in previous PR due to different naming convention in chain_model


FreyJo:
is also part of the new pull request, swig integrator interface

PR #350, opened 2018-12-12
experimental code to test SWIG wrapper for acados_c interface

schoelst:
I created this branch to evaluate the effort of creating a swig interface for acados_c interface. The result was very complex code in the high-level language. Because the concurrently developed MEX approach seemed easier at the time of writing this, this approach was abandoned.


giaf:
Thanks!

PR #356, opened 2018-12-17
[WIP] OSQP interface

bnovoselnik:
Hi guys! Sorry for the delay, but here is the first working version of the OSQP interface...
Continuing on the work done by @dkouzoup, the checklist has been updated as follows:

- [x] setup framework for new solver

- [x] convert data to OSQP format

- [x] solve QP

- [x] add tests

- [x] extend interfaces

- [x] fix travis

- [x] replace internal memory allocation (to a certain extent)


bnovoselnik:
In the latest commits I replaced most of internal dynamic memory allocations when creating the OSQP workspace without going deep into the codebase of the OSQP itself, i.e. I basically rewrote osqp_setup in acados style :)

However, a lot of mallocs still remain even deeper in OSQP, but to get rid of those one would have to really refactor the entire OSQP library which is not as trivial :/

test/ocp_qp/test_qpsolvers.cpp
```diff
@@ -145,7 +159,7 @@ TEST_CASE("mass spring example", "[QP solvers]")
 
     int ngN = 0;  // number of general constraints at last stage
 
-    double N2_values[] = {15, 5, 3};  // horizon of partially condensed QP for sparse solvers
```
giaf: Why? Any issue with N2=3?

giaf:
At a first glance it looks good, I will dig into it when I have a bit more time.

In the mean while, could you update the make build system too? Thanks!

test/ocp_qp/test_qpsolvers.cpp
bnovoselnik: Oh right, forgot about that. Not exactly an issue, more of an inconvenience. OSQP is not accurate enough, for N2=3 the tolerances for primal and dual residuals should be quite smaller to get an accurate solution. I'll add these tolerances to OSQP opts so that they can be changed from outside...

bnovoselnik:
Will do.

test/ocp_qp/test_qpsolvers.cpp
bnovoselnik: Nvm, I just realized polishing (for high accuracy) has been turned of in OSQP. With polishing turned on, the accuracy is on par with other solvers :)

giaf: nice :)

bnovoselnik:
> In the mean while, could you update the make build system too? Thanks!

Make build system updated!

PR #360, opened 2019-01-11
Support algebraic variables in linear least-squares costs

zanellia:
With this pull request we add support for algebraic variables in linear least-squares costs: 

|| Vx * x + Vu * u + Vz * z ||^2_W.

Additional changes: 

1. Updated templating interface to support DAEs and parameters in the dynamics.
2. Some more work on generation of S-Functions from the templating interface.


zanellia:
closing in favor of https://github.com/acados/acados/pull/367

PR #362, opened 2019-01-28
Minor examples fixes

tmmsartor:


giaf:
what about calling `end.sh` again as `setenv.sh` ?

tmmsartor:
Because I thought was more intuitive in that way, thus easier to assume its behavior and remember it.
This bacause it is not setting the enviroment, it is the enviroment to be setted by your current bash process.

PR #365, opened 2019-02-08
GNSF detection & documentation

FreyJo:
Worked a bit on GNSF, to prepare it for Matlab interface.
Also, found the `roadmap.md`, maybe we can try to update it at some point.


giaf:
Thanks!
Did you try if a transcribed linear model successfully goes throurght the GNSF sim in the core?

PR #371, opened 2019-03-05
update README

zanellia:
Hey guys, since the main README.md was quite outdated, me and @tmmsartor took the liberty to update it. Please feel free to add any change!


README.md
```diff
@@ -5,87 +5,57 @@
...
     mkdir -p build
     cd build
-    cmake -D SWIG_MATLAB=1 -D SWIG_PYTHON=1 ..   # set SWIG_MATLAB=0 if you don't have MATLAB installed
+    cmake .. (with optional arguments e.g. -DACADOS_WITH_OSQP=OFF/ON -DACADOS_INSTALL_DIR=<path_to_acados_installation_folder>)
     make install
     ```
+
+    When using the Makefile-based build sytem:
+    ```
+    make acados_shared
```
giaf: If you make the shared libraries, you also need to export
`LD_LIBRARY_PATH=$LD_LIBRARY_PATH:acados_folder/lib`
otherwise the examples will not run.

If you _only_ make the static libraries you don't need this. (You still need to do if you make both static and dynamic libraries)

README.md
```diff
@@ -5,87 +5,57 @@
...
-export PYTHONPATH=<path_to_acados_installation_folder>:$PYTHONPATH
-```
-To run a Python example from the acados root folder:
-```
-python examples/python/ocp_nlp.py
-```
-
-### Manual Installation
+### Optional requirements
+Some functionalities in acados require CasADi (version 3.4.0) to be installed on your system.
```
giaf: it should work for any version later than 3.4.0 (to be checked)

giaf: I just checked from octave, and it works with 3.4.5 (but the scripts generating the casadi functions need to be updated, because now they exit if the version is not 3.4.0)

tmmsartor: Yes but the python interface from @zanellia may be stricter about the required version, any is very easy to get most tested/supported version using the proper link.

README.md
tmmsartor: I made a pull request with this small fix

README.md
zanellia: let's just merge this? We can update it later on ;)

PR #373, opened 2019-04-01
Installation instructions casadi download path fix

jkoendev:
fixed the download paths, and added --show-progress to wget that you are able to see if the download succeeded


jkoendev:
also added ```cd ..``` to make the instructions consistent

PR #396, opened 2019-06-07
adding (stage-wise) setter for y_ref both in Simulink and Python

zanellia:
@giaf, in order to implement this, I have added a couple of things to the core and C interface, have a look ;)

### DIFF
--- a/acados/ocp_nlp/ocp_nlp_constraints_bgh.c
+++ b/acados/ocp_nlp/ocp_nlp_constraints_bgh.c
@@ -252,6 +252,22 @@ static void ocp_nlp_constraints_bgh_get_nb(void *config_, void *dims_, int* valu
 
 
 
+static void ocp_nlp_constraints_bgh_get_nbx(void *config_, void *dims_, int* value)
+{
+    ocp_nlp_constraints_bgh_dims *dims = (ocp_nlp_constraints_bgh_dims *) dims_;
+    *value = dims->nbx;
+}
+
+
+
+static void ocp_nlp_constraints_bgh_get_nbu(void *config_, void *dims_, int* value)
+{
+    ocp_nlp_constraints_bgh_dims *dims = (ocp_nlp_constraints_bgh_dims *) dims_;
+    *value = dims->nbu;
+}
+
+
+
 static void ocp_nlp_constraints_bgh_get_ng(void *config_, void *dims_, int* value)
 {
     ocp_nlp_constraints_bgh_dims *dims = (ocp_nlp_constraints_bgh_dims *) dims_;
@@ -285,6 +301,14 @@ void ocp_nlp_constraints_bgh_dims_get(void *config_, void *dims_, const char *fi
     else if (!strcmp(field, "nb"))
     {
         ocp_nlp_constraints_bgh_get_nb(config_, dims_, value);
+    }
+    else if (!strcmp(field, "nbx"))
+    {
+        ocp_nlp_constraints_bgh_get_nbx(config_, dims_, value);
+    }
+    else if (!strcmp(field, "nbu"))
+    {
+        ocp_nlp_constraints_bgh_get_nbu(config_, dims_, value);
     }
     else if (!strcmp(field, "ng"))
     {
@@ -1006,7 +1030,7 @@ void ocp_nlp_constraints_bgh_config_initialize_default(void *config_)
     config->dims_assign = &ocp_nlp_constraints_bgh_dims_assign;
     config->dims_initialize = &ocp_nlp_constraints_bgh_dims_initialize;
     config->dims_set = &ocp_nlp_constraints_bgh_dims_set;
-    config->get_dims = &ocp_nlp_constraints_bgh_dims_get;
+    config->dims_get = &ocp_nlp_constraints_bgh_dims_get;
     config->model_calculate_size = &ocp_nlp_constraints_bgh_model_calculate_size;
     config->model_assign = &ocp_nlp_constraints_bgh_model_assign;
     config->model_set = &ocp_nlp_constraints_bgh_model_set;
--- a/acados/ocp_nlp/ocp_nlp_cost_external.c
+++ b/acados/ocp_nlp/ocp_nlp_cost_external.c
@@ -576,6 +576,7 @@ void ocp_nlp_cost_external_config_initialize_default(void *config_)
     config->dims_assign = &ocp_nlp_cost_external_dims_assign;
     config->dims_initialize = &ocp_nlp_cost_external_dims_initialize;
     config->dims_set = &ocp_nlp_cost_external_dims_set;
+    config->dims_get = &ocp_nlp_cost_external_dims_get;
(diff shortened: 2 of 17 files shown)


acados/ocp_nlp/ocp_nlp_constraints_bgh.c
```diff
@@ -1006,7 +1030,7 @@ void ocp_nlp_constraints_bgh_config_initialize_default(void *config_)
     config->dims_assign = &ocp_nlp_constraints_bgh_dims_assign;
     config->dims_initialize = &ocp_nlp_constraints_bgh_dims_initialize;
     config->dims_set = &ocp_nlp_constraints_bgh_dims_set;
-    config->get_dims = &ocp_nlp_constraints_bgh_dims_get;
+    config->dims_get = &ocp_nlp_constraints_bgh_dims_get;
```
giaf: this kind of cleaning is always good :)

acados/ocp_nlp/ocp_nlp_cost_ls.c
```diff
@@ -179,6 +179,27 @@ void ocp_nlp_cost_ls_dims_set(void *config_, void *dims_, const char *field, int
...
+
+void ocp_nlp_cost_ls_dims_get(void *config_, void *dims_, const char *field, int* value)
+{
+    if (!strcmp(field, "ny"))
+    {
+        ocp_nlp_cost_ls_get_ny(config_, dims_, value);
+    }
+    else
+    {
+        printf("error: attempt to get dimension from constraint model, that is not there");
```
giaf: I would go for an error message like: 
`printf("error: ocp_nlp_cost_ls_dims_get: wrong field %s\n", field);`
which is much better for debugging, because it tells you exactly where and what is the issue

acados/ocp_nlp/ocp_nlp_cost_nls.c
```diff
@@ -139,6 +139,30 @@ void ocp_nlp_cost_nls_dims_set(void *config_, void *dims_, const char *field, in
...
+
+void ocp_nlp_cost_nls_dims_get(void *config_, void *dims_, const char *field, int* value)
+{
+    if (!strcmp(field, "ny"))
+    {
+        ocp_nlp_cost_nls_get_ny(config_, dims_, value);
+    }
+    else
+    {
+        printf("error: attempt to get dimension from constraint model, that is not there");
```
giaf: see above

acados/ocp_nlp/ocp_nlp_cost_nls.h
```diff
@@ -63,6 +63,8 @@ void *ocp_nlp_cost_nls_dims_assign(void *config, void *raw_memory);
 void ocp_nlp_cost_nls_dims_initialize(void *config, void *dims, int nx, int nu, int ny, int ns, int nz);
 //
 void ocp_nlp_cost_nls_dims_set(void *config_, void *dims_, const char *field, int* value);
+//
+void ocp_nlp_cost_nls_dims_get(void *config_, void *dims_, const char *field, int* value);
```
giaf: you need to do the same for `cost_external` too !

interfaces/acados_c/ocp_nlp_interface.c
```diff
@@ -501,6 +501,9 @@ void ocp_nlp_out_get(ocp_nlp_config *config, ocp_nlp_dims *dims, ocp_nlp_out *ou
 int ocp_nlp_dims_get(ocp_nlp_config *config, ocp_nlp_dims *dims, ocp_nlp_out *out,
```
giaf: I don't get the meaning of this function.
A function called `ocp_nlp_dims_get` should take as fields `nx`, `nu`, `ny`, `nbx`, ... and not the name of model members...

interfaces/acados_c/ocp_nlp_interface.c
```diff
@@ -509,9 +512,29 @@ int ocp_nlp_dims_get(ocp_nlp_config *config, ocp_nlp_dims *dims, ocp_nlp_out *ou
     {
         return dims->nx[stage];
```
giaf: This is wrong in any case...

acados/ocp_nlp/ocp_nlp_cost_ls.c
zanellia: good point, I must have copy pasted this from somewhere else. For now I have fixed these ones (see a90f6f8)

acados/ocp_nlp/ocp_nlp_cost_nls.c
zanellia: (see a90f6f8)

interfaces/acados_c/ocp_nlp_interface.c
zanellia: good catch, fixed here 1691f2c

interfaces/acados_c/ocp_nlp_interface.c
zanellia: The function name should probably be changed given the confusion it creates. The point is that when you need to set/get something you can ask the core for its dimension in order to enforce consistency and avoid memory corruptions, It should really implement a map from attributes to their dimensions ("x" -> nx, "lbx" -> nbx and so on)

acados/ocp_nlp/ocp_nlp_cost_nls.h
zanellia: The external cost doesn't contain data that can be set online, does it? Maybe just the parameters?

interfaces/acados_c/ocp_nlp_interface.c
zanellia: see 90f0d9c

acados/ocp_nlp/ocp_nlp_cost_nls.h
giaf: Surely the parameter.

But in any case, generally speaking all instances of the same module must implement the same functions. This means that, in any case, there should be at least a function `ocp_nlp_cost_external_dims_get` which does nothing (or complains and exit).

As it is, if called though `config`, config->dims_get will give segfault for external cost...

acados/ocp_nlp/ocp_nlp_cost_external.c
```diff
@@ -107,6 +107,14 @@ void ocp_nlp_cost_external_dims_set(void *config_, void *dims_, const char *fiel
 
 
 
+void ocp_nlp_cost_external_dims_get(void *config_, void *dims_, const char *field, int* value)
+{
+        printf("error: ocp_nlp_cost_external_dims_get: attempt to get dimensions of non-existing field %s\n", field);
+        exit(1);
+}
+
+
```
giaf: almost ;)

just set also the `config->dims_get = &ocp_nlp_cost_external_dims_get` and we are good :)

zanellia: oh god, sorry... 1ccdc

zanellia:
let's merge this before Thu? :)

PR #400, opened 2019-07-04
GNSF for fully linear systems

FreyJo:
In this PR, I added a linear term in the controls to the Linear Output part of GNSF.
This makes it possible to use GNSF with fully linear systems more efficiently without calling an external function.
In the case that f_LO is trivial, I moved more into the precomputation phase.
Also, the automatic structure detection is updated wrt to the new structure.

[New GNSF structure with BLO.pdf](https://github.com/acados/acados/files/3359595/gnsf_structure_blo.pdf)

[Old GNSF structure without BLO](https://github.com/acados/acados/files/2318322/gnsf_structure.pdf)


FreyJo:
I suggest to merge this, if there are no remarks/complaints.

giaf:
@FreyJo I have something else for you here, we can discuss when you come to the office ;)

giaf:
I have one fix in case of adjoint sensitivities. At line 1798 of `sim_gnsf.c` https://github.com/acados/acados/blob/6e3229f204adf6b9ae3d81bc2ef4dd5155e29682/acados/sim/sim_gnsf.c#L1798 you state that `forward and adjoint sensitivities are constant and already computed`.

This is true for the forward sensitivities (since the seeds are constantly `I` for the states and `0` for the inputs), but this is not true for the adjoint sensitivities (since the adjoint seed changes every time the routine is called).

To fix that, you just need to compute the adjoint sensitivity as the product of forward sensitivity and anjoint seed, `lambda = S_forw' * in->S_adj` using `blasfeo_dgemv_t`, at line1798 too. Then this https://github.com/acados/acados/blob/6e3229f204adf6b9ae3d81bc2ef4dd5155e29682/acados/sim/sim_gnsf.c#L2579 will correctly extract the adjoint sensitivity from `lambda`.

FreyJo:
Thanks @giaf ! I fixed this.
Also, I take the forward seed into account now.
However, I think we dont really need them.
In the whole project, they are only used with identity matrix.
So we could think about removing `in->S_forw` imo, since one can always multiply with something else outside.

PR #401, opened 2019-07-05
license: Update headers

tmmsartor:
This changes a lot of lines.


giaf:
I guess we should not add a copyright notice and license to all the CasADi-generated model files in the examples...

giaf:
ok good

if there is anything left to do, we can do in another PR

merging this now

PR #403, opened 2019-07-11
add GNSF to matlab mex interface

giaf:
- add GNSF to matlab mex interface
- use matlab eval & C pre-processing for templating of matlab mex interface
    * user-defined model name
    * multiple solvers in the same workspace
    * allows for future introduction of multiple phases


acados/sim/sim_gnsf.c
```diff
@@ -169,12 +169,10 @@ static void sim_gnsf_import_matrices(void *dims_, gnsf_model *model)
         model->c_LO[ii] = 0.0;
 
     // ipiv_x, z
-    double ipiv_x[nx];
```
giaf: oh yeah, good fix, we don't like variable size arrays ;)

Issue #330, opened 2018-09-04
Undefined behavior due to alignment


roversch:
When running e.g. the `nonlinear_chain_ocp_nlp_example` with `-fsanitize=undefined` compiler flags for Clang, I get many errors concerning undefined behavior, mainly of the type

- member access within misaligned address
- load of misaligned address

For example, in line 

https://github.com/acados/acados/blob/70a9a83fb4fdacd0d7429fe6acc75bce40589c64/acados/ocp_nlp/ocp_nlp_dynamics_cont.c#L64 

`sim` is not aligned to 8 bytes, but to 4. This is what invokes the undefined behavior. The error messages are quite helpful, these errors could be resolved by putting extra alignment points in `calculate_size` and `assign` functions.

dkouzoup:
Nice! This is very helpful, thanks for reporting

dkouzoup:
@roversch, is there anything special I need to do to enable this? By adding it in the CMAKE_C_FLAGS I get many errors like this one:

Undefined symbols for architecture x86_64:
  "___ubsan_handle_add_overflow", referenced from:
      _main in dense_qp.c.o
      _blasfeo_allocate_dmat in libblasfeo.a(d_aux_ext_dep_lib4.c.o)
      _blasfeo_allocate_dvec in libblasfeo.a(d_aux_ext_dep_lib4.c.o)
      _blasfeo_print_dmat in libblasfeo.a(d_aux_ext_dep_lib4.c.o)
      _blasfeo_print_to_file_dmat in libblasfeo.a(d_aux_ext_dep_lib4.c.o)
      _blasfeo_print_to_file_exp_dmat in libblasfeo.a(d_aux_ext_dep_lib4.c.o)
      _blasfeo_print_to_string_dmat in libblasfeo.a(d_aux_ext_dep_lib4.c.o)
      ...
  "___ubsan_handle_divrem_overflow", referenced from:
      _print_ocp_qp_info in libacados.a(print.c.o)
      _make_int_multiple_of in libacados.a(mem.c.o)
      _align_char_to in libacados.a(mem.c.o)
      _QProblemB_hotstart in libqpOASES_e.a(QProblemB.c.o)
      _QProblemB_updateFarBounds in libqpOASES_e.a(QProblemB.c.o)
      _QProblemB_backsolveRrem in libqpOASES_e.a(QProblemB.c.o)
      _QProblemB_performRatioTestB in libqpOASES_e.a(QProblemB.c.o)
      ...
[...]

PR #408, opened 2019-07-18
Portability: wrong alignments on 32-bit systems

roversch:
When porting acados to a 32-bit system, I encountered wrong alignment of nested structs. More general, we should think more carefully of alignment of memory, especially in the case of structs. 

This PR fixes some of the issues.


roversch:
See also #330

giaf:
This seems to break the unit tests. Could you investigate what the issue is?

roversch:
I had a look, but the log is cut off too early. Locally, the unit test work as they should.

giaf:
It can not be merged like this.

Since the changes are very little, you can try to open another PR to "debug on travis". You can push there the changes one by one with separate commits and/or disable single unit tests. You wrote them so you know much better than me how they work.

FreyJo:
@roversch the changes you made, helped me find a few issues here:
https://github.com/acados/acados/pull/597
Thanks for that!

Issue #412, opened 2019-07-24
Mex examples on Octave 4.2.2


jkoendev:
I needed to replace the `import casadi.*` because it is not supported on Octave 4.2.2

Then i get this error when running `linear_mass_spring_model/example_ocp.m`

```
compiling /home/jonas/repos/acados_octave/examples/matlab_mex/linear_mass_spring_model/../../../interfaces/acados_matlab/ocp_create.c
warning: mkoctfile: building exited with failure status
In file included from /home/jonas/repos/acados_octave/examples/matlab_mex/linear_mass_spring_model/../../../acados/ocp_qp/ocp_qp_common.h:25:0,
                 from /home/jonas/repos/acados_octave/examples/matlab_mex/linear_mass_spring_model/../../../acados/ocp_nlp/ocp_nlp_constraints_common.h:31,
                 from /home/jonas/repos/acados_octave/examples/matlab_mex/linear_mass_spring_model/../../../acados/ocp_nlp/ocp_nlp_common.h:42,
                 from /home/jonas/repos/acados_octave/examples/matlab_mex/linear_mass_spring_model/../../../interfaces/acados_c/ocp_nlp_interface.h:28,
                 from /home/jonas/repos/acados_octave/examples/matlab_mex/linear_mass_spring_model/../../../interfaces/acados_matlab/ocp_destroy.c:6:
/home/jonas/repos/acados_octave/examples/matlab_mex/linear_mass_spring_model/../../../external/hpipm/include/hpipm_d_ocp_qp.h:35:10: fatal error: blasfeo_target.h: No such file or directory
 #include <blasfeo_target.h>
          ^~~~~~~~~~~~~~~~~~
compilation terminated.
```

Ubuntu 18.04

zanellia:
I guess you are missing casadi-octave? `import` should be implemented there.

giaf:
yes if `casadi-octave` is in the matlab path, the implementation of `include` in there is used

giaf:
I also use octave 4.2.2

jkoendev:
ok right the `import` works now, not sure what was wrong before, prob. messed up the path with a second casadi installation. And the blasfeo error also goes on me, made a fresh clone of acados and forgot to compile ;)

PR #413, opened 2019-07-24
Windows support for mex interface

jkoendev:
Changes:

- Paths are generated using `fullfile`
- I had to remove the `\` in `CFLAGS=\$CFLAGS` (not sure if that has consequences), compile command fails on windows when the dash is there
- Replaced `system(mv)` commands with `movefile`
- Parameter `build_dir` to `acados_ocp` and `acados_sim` to specify build directory, defaults to 'pwd/build'
- Checks in acados_ocp, acados_sim destructors. Has led to Matlab crash if compilation was aborted early.

I tested the examples on 

- Windows 10, Matlab2018b
- Ubuntu 18.04, Matlab 2018b
- Ubuntu 18.04, Octave 4.2.2

~I could not test with Octave on Ubuntu, see #412~


examples/matlab_mex/linear_mass_spring_model/linear_mass_spring_model.m
```diff
@@ -1,7 +1,5 @@
 function model = linear_mass_spring_model()
 
-import casadi.*
```
giaf: why removing this?

interfaces/acados_matlab/acados_ocp.m
```diff
@@ -12,42 +12,42 @@
 	methods
 
 
-		function obj = acados_ocp(model, opts)
+		function obj = acados_ocp(model, opts, build_dir)
```
giaf: I would rather add `build_dir` in the `acados_ocp_opts` (where e.g. there are already the flags to enable/disable code generation and mex compilation)

interfaces/acados_matlab/acados_sim.m
```diff
@@ -12,15 +12,15 @@
 	methods
 
 
-		function obj = acados_sim(model, opts)
+		function obj = acados_sim(model, opts, build_dir)
```
giaf: also here, better have in in the `opts`

interfaces/acados_matlab/acados_ocp.m
```diff
@@ -12,42 +12,42 @@
...
 			obj.model_struct = model.model_struct;
 			obj.opts_struct = opts.opts_struct;
+      
+      if nargin <= 2
+        build_dir = fullfile(pwd, 'build');
```
giaf: default values for the options should be set in the `acados_ocp_opts` class

interfaces/acados_matlab/acados_sim.m
```diff
@@ -12,15 +12,15 @@
...
 			obj.model_struct = model.model_struct;
 			obj.opts_struct = opts.opts_struct;
+      
+      if nargin <= 2
+        build_dir = fullfile(pwd, 'build');
```
giaf: same here

interfaces/acados_matlab/ocp_compile_mex.m
```diff
@@ -58,16 +62,22 @@ function ocp_compile_mex(opts)
 		end
 	else
 		if (strcmp(opts.qp_solver, 'full_condensing_qpoases'))
-			mex(mex_flags, 'CFLAGS=\$CFLAGS -std=c99 -fopenmp -DACADOS_WITH_QPOASES', acados_include, acados_interfaces_include, external_include, blasfeo_include, acados_lib_path, '-lacados', '-lhpipm', '-lblasfeo', '-lqpOASES_e', mex_files{ii})
```
giaf: why removing `\`?

interfaces/acados_matlab/sim_set_ext_fun.m
```diff
@@ -138,29 +138,37 @@
...
-		system(['mv -f *.mex build/']);
-	else
-		system(['mv -f *.mexa64 build/']);
-	end
-
+			mex(mex_flags, 'CFLAGS=$CFLAGS -std=c99 -fopenmp', ['-DSET_FIELD=', set_fields{ii}], ['-DMEX_FIELD=', mex_fields{ii}], ['-DFUN_NAME=', fun_names{ii}], acados_include, acados_interfaces_include, acados_lib_path, acados_matlab_lib_path, model_lib_path, '-lacados', '-lhpipm', '-lblasfeo', ['-l', model_name], fullfile(acados_mex_folder, 'sim_set_ext_fun_gen.c'));
+    end
+    
+    if is_octave()
+      movefile('sim_set_ext_fun_gen.mex', [mex_names{ii}, '.mex'])
```
giaf: in `ocp_set_ext_fun.m` this is done with `mexext`: which one is the way to go?

giaf:
Any new on testing on Octave?

jkoendev:
works with Octave/Ubuntu now as well

examples/matlab_mex/linear_mass_spring_model/linear_mass_spring_model.m
jkoendev: i reverted the change, this should not have gone into this pull request (was related to my octave issue). But generally i would prefer not using * imports.

interfaces/acados_matlab/acados_ocp.m
jkoendev: moved to acados_ocp_opts

interfaces/acados_matlab/acados_sim.m
jkoendev: resolved

interfaces/acados_matlab/acados_ocp.m
jkoendev: resolved

interfaces/acados_matlab/acados_sim.m
jkoendev: resolved

interfaces/acados_matlab/ocp_compile_mex.m
jkoendev: Fails on windows with the message:
```
gcc: error: \-fexceptions: No such
file or directory
```
so the `\` seems to end up in the gcc command.

I do not understand the meaning of the `\` (escape?) here. Why is it there? The mex command works without on Win/Linux/Matlab/Octave, but i am not sure if potentially CFLAGS are getting lost.

interfaces/acados_matlab/sim_set_ext_fun.m
jkoendev: i think `mexext` is the way to go, forgot to change it here.

jkoendev: resolved

interfaces/acados_matlab/acados_ocp.m
jkoendev: i named the option `output_dir`

interfaces/acados_matlab/ocp_compile_mex.m
giaf: Ok, we need to check on older matlab versions I guess. I found that on online forums and it works on my machines.

interfaces/acados_matlab/acados_ocp.m
giaf: great!

interfaces/acados_matlab/ocp_compile_mex.m
jkoendev: so in the docs it is described and they are not using the `\`:
```
[mex docs - Override Default Compiler Switch Option](https://www.mathworks.com/help/matlab/ref/mex.html#btw193g-1)
```
but they only show examples for the command line form, and the functional form of `mex`.
On google i found both usages, but no explanation, so i guess we need to test it ourself.

FreyJo:
looks good to me!
Looking forward to the acados from Matlab on Windows demo next week :tada:

giaf:
Can you change the destination of the PR from `acados:measter` to `acados:matlab_interface`?
So I can play around a bit with that before merging into master

jkoendev:
done

interfaces/acados_matlab/acados_ocp.m
```diff
@@ -102,8 +99,13 @@ function set(varargin)
 
 
 		function delete(obj)
-			ocp_destroy_ext_fun(obj.model_struct, obj.C_ocp, obj.C_ocp_ext_fun);
-			ocp_destroy(obj.C_ocp);
+      if ~isempty(obj.C_ocp_ext_fun)
```
giaf: How can this be empty? Both are always created in the constructor, right?

interfaces/acados_matlab/ocp_compile_mex.m
```diff
@@ -58,16 +62,22 @@ function ocp_compile_mex(opts)
...
 if is_octave()
-	system(['mv -f *.o build/']);
-	system(['mv -f *.mex build/']);
-else
-	system(['mv -f *.mexa64 build/']);
+  movefile('*.o', opts.output_dir)
 end
+
+for k=1:length(mex_names)
+  clear(mex_names{k})
```
giaf: What is the `clear` for? in `sim_compile_mex` you did not add it.

interfaces/acados_matlab/acados_ocp.m
jkoendev: The compilation is done in the constructor. When you git `ctrl+c` during compilation, or if there is any error in the constructor, still the destructor is being called (at least on windows) which leads to a mex crash because the functions e.g. `ocp_destroy_ext_fun` have invalid arguments.

interfaces/acados_matlab/ocp_compile_mex.m
jkoendev: The problem is: when the mex files are loaded by Matlab you can not delete them, so the `movefile` command fails (e.g. on the second run). 
I will check this again later as i still had to do a `clear all` sometimes.

PR #417, opened 2019-07-29
Fixed make command for creating shared libraries in the documentation.

EnricaSo:
"acados_c_shared" is not existing. The correct should be "make acados_shared".


giaf:
Shanks for spotting it, we forgot to update the file after the change!

PR #418, opened 2019-07-31
Comments in cmake.in files in proper way

pariterre:


giaf:
Thanks for the fixes!

PR #420, opened 2019-08-02
Use mbuild to build ocp model ext fun

jkoendev:
Notes:
 - ~Need to create a temporary empty file `ocp_model.c` to set the output name of the shared library to `ocp_model.so`. Renaming is not possible somehow. `mbuild` uses the first c-file to determine the output name. Any idea how to set the output name with `mbuild` or `mex`?~
 - There are two possible commands using `mbuild` or `mex`. The `mbuild` command is a bit simpler thus preferred i guess, but we need to test if it works on all platforms.

### DIFF
--- a/interfaces/acados_matlab/ocp_generate_casadi_ext_fun.m
+++ b/interfaces/acados_matlab/ocp_generate_casadi_ext_fun.m
@@ -99,15 +99,15 @@ end
 lib_name = 'ocp_model';
 
 if ispc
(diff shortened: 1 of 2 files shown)


jkoendev:
If this gets approved i would also change the creation of `c_files` and create a cell array from the beginning and not append string to `c_sources`

jkoendev:
ready to be reviewed @giaf

giaf:
It looks good, it also sounds good to create the cell array from the beginning!

The same should be done for `sim_generate_casadi_ext_fun.m` too.

jkoendev:
ok i will do the same for sim. I'll write here again when its ready to review.

jkoendev:
changed `sim_generate_casadi_ext_fun.m` and creation of cell arrays, ready for review @giaf

PR #421, opened 2019-08-02
Matlab interface

FreyJo:
fix gnsf_structure detection for parametric models


interfaces/acados_matlab/check_reformulation.m
```diff
@@ -73,16 +73,17 @@
 y = gnsf.y;
 uhat = gnsf.uhat;
 
-if (strcmp(gnsf.param_f, 'true'))
-    p = model.p;
+isparametric = isfield(model, 'sym_p');
+if isparametric
```
giaf: now all casadi functions are assumed to be parametric, so ...

interfaces/acados_matlab/check_reformulation.m
```diff
@@ -73,16 +73,17 @@
...
+    p = model.sym_p;
     np = length(p);
     % create functions
     impl_ode_fun = Function('impl_ode_fun', {x, xdot, u, z, p}, {model.dyn_expr_f});
     phi_fun = Function('phi_fun',{y,uhat, p}, {gnsf.phi_expr});
     f_lo_fun = Function('f_lo_fun',{x(1:nx1), xdot(1:nx1), z, u, p}, {gnsf.f_lo_expr});
 else
     % create functions
-    impl_ode_fun = Function(['impl_ode_fun'], {x, xdot, u, z}, {model.dyn_expr_f});
+    impl_ode_fun = Function('impl_ode_fun', {x, xdot, u, z}, {model.dyn_expr_f});
```
giaf: ... `p` should always be there in the input list. Just if the function is not parametric, `p` does not appear in the function expression

FreyJo: okay, i changed that now

PR #423, opened 2019-08-04
templates: Rename library to follow linux conventions

tmmsartor:


giaf:
That's not just linux convention... ;)

zanellia:
Thanks @tmmsartor! Although, I guess this has broken the other examples using the template engine :P

zanellia:
actually not, it all works fine!

tmmsartor:
The name of the library file was used only in the generate_solver.py script inside the interface module, at least from my quick check, I didn't check the matlab version.

Issue #407, opened 2019-07-17
Setting up LD_LIBRARY_PATH for the Matlab interface


jkoendev:
I think it is undesirable to have to write and run/source an `env.sh` for each problem that you implement in Matlab. This disturbs the workflow a lot (also if used from OpenOCL) because you have to restart Matlab, and run Matlab from the command line after you implemented your model.

I investigated and tried writing a Matlab `.m` script that sets the environment variables in the same way but unfortunately setting LD_LIBRARY_PATH does not work as intended. I think Matlab only uses LD_LIBRARY_PATH at the startup to look for runtime libraries (.so), and it does not have an effect if you change the path later. I have not found any workaround for that.

~An option would be to use the matlab `system` command to set the environment variables, and restart matlab but this might have too many side effects (not sure if it works for all configurations).~

Another option would be to setup a runtime directory during compilation of acados, e.g. `acados/runtime` or using `acados/lib`, add it to LD_LIBRARY_PATH forever and copy all the .so/.mex files there. So the `libacados.so`, `libblasfeo.so` etc. but also the generated model files like `libocp_model.so`, `ocp_create.mexa64` etc. These need to have a unique name for each model then (which is probably a good idea anyway because there could be conflicts if multiple models are added to the path). This would be my preferred way atm.

jkoendev:
setting ld path more permanently on Linux (outside of bash) might involve `ldconfig`.

Another option would be to use linker arguments `-Wl,-rpath,/path/to/runtime/dir`.

giaf:
some inspiration maybe
https://tomopt.com/docs/instguide/tomlab_install006.php

PR #424, opened 2019-08-05
Mex windows doc and env setup script

jkoendev:
* Setting up the environment variable on Windows with a Matlab script.
* Some documentation in the `README.md` on how to run examples on Windows.

### DIFF
--- a/examples/matlab_mex/README.md
+++ b/examples/matlab_mex/README.md
@@ -36,3 +36,4 @@ Run the script `acados_examples_env.m` in this folder to setup environment
 variables and the path for acados mex interface.
 
 Navigate to one of the examples and run the example script.
+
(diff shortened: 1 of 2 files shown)


jkoendev:
Can be reviewed @giaf

FreyJo:
I think we should put a higher priority on documentation within the `docs` folder than local readme files, since the [docs.acados.org](docs.acados.org) webpage is generated through that, can be googled etc.
Thus, I suggest to include the information there.

jkoendev:
you are right, i was just not very familiar with the doxygen stuff yet. I just found this Markdown file which might be the appropriate place:
https://github.com/acados/acados/blob/master/docs/interfaces/index.md

This PR is more like a shortterm solution to set-up the path on windows, i didn't want to touch the Linux part for now. I think we should seek a more permanent solution in the long run for both operation systems, see also #407

giaf:
Agree. But it is also good to have this information in local readme files, that is the right information in the right place.

Issue #428, opened 2019-08-07
Mex setting maximum iterations


jkoendev:
Can i set the maximum iterations after the solver object was created somehow? 

e.g. 
```
ocp = acados_ocp(ocp_model, ocp_opts);
ocp.set('nlp_solver_max_iter', 1)
```

@giaf

FreyJo:
Have you tried this?
`ocp_opts.set('nlp_solver_max_iter', nlp_solver_max_iter);`
As far as I understand, this setter will set the option in the C object, which is used by the ocp_solver internally, created by
`ocp = acados_ocp(ocp_model, ocp_opts);`

jkoendev:
i just tried and it didn't work! In the pendulum on a cart example i did
```
ocp_opts.set('nlp_solver_max_iter', 4);
ocp.solve();
```
but i t still does 9 iterations

giaf:
That won't work, because it just changes the value locally in the `ocp_opts`.

The way to go is to implement `ocp.set('nlp_solver_max_iter', 1)`

FreyJo:
will be merged soon https://github.com/acados/acados/pull/431

giaf:
here it is https://github.com/acados/acados/commit/44c215e246b1dc92ace600c37a05d97265ae658d

giaf:
@FreyJo ops I also did it now... sorry I didn't see you PR...
BTW the matter was slightly more complicated, see commit above

PR #431, opened 2019-08-07
Matlab interface

FreyJo:
implemented this one
https://github.com/acados/acados/issues/428


FreyJo:
Most options can not be changed after the `ocp_nlp` solver was created

giaf:
Sorry @FreyJo I was also working independently on this and just pushed to the master before noticing your PR

BTW your PR would fail for a new value of `max_iter` that would be reached by the solver, giving segfault (I guess you didn't test this case). There was some more stuff to fix around to make this work, see the changes here https://github.com/acados/acados/commit/44c215e246b1dc92ace600c37a05d97265ae658d

FreyJo:
I see, thanks!

PR #441, opened 2019-08-16
ZynqMP bare-metal

besticka:
bare-metal compilation for Xilinx ZynqMP systems


CMakeLists.txt
```diff
@@ -38,7 +38,7 @@ set(BUILD_SHARED_LIBS ON CACHE STRING "Build shared libraries")
 
 # Additional targets
 option(ACADOS_UNIT_TESTS "Compile Unit tests" OFF)
-option(ACADOS_EXAMPLES "Compile Examples" OFF)
+option(ACADOS_EXAMPLES "Compile Examples" ON)
```
tmmsartor: Why do you need to change the default value?

tmmsartor: Anyway I will be in favor to keep the examples ON by default.
This was off because qpOASES was off by default,
so if you removed all qpOASES usage from the examples,
I would be up for this change.

examples/c/chain_model/chain_model.h
```diff
@@ -22,23 +22,23 @@
...
-
-#define XN_NM2_FILE "/home/gianluca/acados/examples/c/chain_model/xN_nm2.txt"
-#define XN_NM3_FILE "/home/gianluca/acados/examples/c/chain_model/xN_nm3.txt"
-#define XN_NM4_FILE "/home/gianluca/acados/examples/c/chain_model/xN_nm4.txt"
-#define XN_NM5_FILE "/home/gianluca/acados/examples/c/chain_model/xN_nm5.txt"
-#define XN_NM6_FILE "/home/gianluca/acados/examples/c/chain_model/xN_nm6.txt"
-#define XN_NM7_FILE "/home/gianluca/acados/examples/c/chain_model/xN_nm7.txt"
-#define XN_NM8_FILE "/home/gianluca/acados/examples/c/chain_model/xN_nm8.txt"
-#define XN_NM9_FILE "/home/gianluca/acados/examples/c/chain_model/xN_nm9.txt"
+#define X0_NM2_FILE "/home/bstickan/temp/acados/examples/c/chain_model/x0_nm2.txt"
```
tmmsartor: This file is templated by CMake,
this file shold not be tracked by git.

Not sure if it is templated by Makefile,
maybe it should be added there.

giaf: It is only templated by cmake. And actually now it is not even used at all, since I hand coded the initial values in a `.c` file instead. But I would keep this file as a reference for where the data in the `.c` file comes from (i.e. the `.m` script generated the `.txt` data files). If you feel like directly generating the `.c` files from the `.m` script, then afterwards you can remove this header completely.

cmake/Toolchain-aarch64.cmake
```diff
@@ -0,0 +1,36 @@
+set(CMAKE_SYSTEM_NAME "Generic" CACHE STRING "")
```
giaf: Should this file be renamed to include the fact that it is xilinx and bare metal? If I want to compile acados natively on linux on my ARM 64-bit board I should not use this file.

CMakeLists.txt
besticka: I thought it would be good to compile examples by default. Users would like to start with examples, but if these break the build process they have to figure out why...
Would "Compile Examples" implies qpOASES being enabled be an option?

cmake/Toolchain-aarch64.cmake
besticka: How about Toolchain-xilinx-a53-bare.cmake?
-->you can use this toolchain for every xilinx SoC-based board that has the a53 core if you want bare-metal compilation.

FreyJo:
Closing due to inactivity. Might be helpful for users on similar systems anyway.

PR #443, opened 2019-08-16
Running tests with appveyor, compile on appveyor with mingw

jkoendev:
- Use MinGW to compile acados on appveyor
- Fix for sh.exe and PATH to MinGW in appveyor script
- Build examples on appveyor
- Run examples on appveyor
- Floating point comparison fix in sim_gnsf (by @FreyJo)

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -29,7 +29,8 @@ build_script:
   # MinGW files need to be at the beginning of the PATH
   - set PATH=C:\msys64\usr\bin;%PATH%
   # Workaround for CMake not wanting sh.exe on PATH for MinGW
-  - rm "C:/Program Files/Git/usr/bin/sh.exe"
+  - rm "C:\msys64\usr\bin\sh.exe"
+  - rm "C:\Program Files\Git\usr\bin\sh.exe"
   - cmake -E make_directory build
   - cmake -E chdir build cmake -DCMAKE_PREFIX_PATH=C:\mingw64 -D ACADOS_WITH_QPOASES=ON -D BLASFEO_TARGET=GENERIC -D HPIPM_TARGET=GENERIC -D BUILD_SHARED_LIBS=OFF -D ACADOS_EXAMPLES=ON -D ACADOS_UNIT_TESTS=ON -DEIGEN3_INCLUDE_DIR="C:\dev\Eigen3\include\eigen3" -G "MinGW Makefiles" ..
   - cmake --build build --config Release
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -15,7 +15,7 @@ build_script:
   - rm "C:/Program Files/Git/usr/bin/sh.exe"
   - cmake -E make_directory build
   # TODO: make this work
-  # - cmake -E chdir build cmake -D ACADOS_WITH_QPOASES=ON -D BLASFEO_TARGET=GENERIC -D HPIPM_TARGET=GENERIC -D BUILD_SHARED_LIBS=OFF  -D ACADOS_EXAMPLES=ON ..
-  - cmake -E chdir build cmake -D ACADOS_WITH_QPOASES=ON -D BLASFEO_TARGET=GENERIC -D HPIPM_TARGET=GENERIC -D BUILD_SHARED_LIBS=OFF -G "MinGW Makefiles" ..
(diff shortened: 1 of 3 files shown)


jkoendev:
ready for review

FreyJo:
nice, thanks for fixing this!
Now something is actually run on appveyor.
Can you try to add `-D ACADOS_UNIT_TESTS=ON`, then it should not only run the examples, but also the tests in the `test` folder, which include checks for accuracy.

.appveyor.yml
```diff
@@ -2,18 +2,33 @@ branches:
...
-  # - set PATH=%PATH:C:/Program_Files/Git/usr/bin/;=%
-  # - rm "C:/Program Files/Git/usr/bin/sh.exe"
+  - rm "C:/Program Files/Git/usr/bin/sh.exe"
   - cmake -E make_directory build
-  # TODO: make this work
-  # - cmake -E chdir build cmake -D ACADOS_WITH_QPOASES=ON -D BLASFEO_TARGET=GENERIC -D HPIPM_TARGET=GENERIC -D BUILD_SHARED_LIBS=OFF -D ACADOS_EXAMPLES=ON -G "MinGW Makefiles" ..
-  - cmake -E chdir build cmake -D ACADOS_WITH_QPOASES=ON -D BLASFEO_TARGET=GENERIC -D HPIPM_TARGET=GENERIC -D BUILD_SHARED_LIBS=OFF ..
+  - cmake -E chdir build cmake -D ACADOS_WITH_QPOASES=ON -D BLASFEO_TARGET=GENERIC -D HPIPM_TARGET=GENERIC -D BUILD_SHARED_LIBS=OFF -D ACADOS_EXAMPLES=ON -D ACADOS_UNIT_TESTS=ON -G "MinGW Makefiles" ..
   - cmake --build build --config Release
   - cmake -E chdir build ctest -V
```
FreyJo: When it works, I would replace
`-V` with `--output-on-failure`.
This should make it easier to identify errors in the log.

jkoendev: done in https://github.com/acados/acados/pull/443/commits/f39b40200522ad2dde6b6b2dfda1952a7bc7b516

jkoendev:
ready for review

acados/sim/sim_gnsf.c
```diff
@@ -1715,9 +1716,10 @@ int sim_gnsf(void *config, sim_in *in, sim_out *out, void *args, void *mem_, voi
     int nxz2 = nx2 + nz2;
 
     // assert - only use supported features
-    if (mem->dt != in->T / opts->num_steps)
+    double epsilon = 1e-10 * fmax( fabs(mem->dt), fabs(in->T) );
```
giaf: Did you get an issue with that?

FreyJo: yes, the `sim_inverted_pendulum` example was failing on appveyor.
There, precompute is called after all options (and T) are set.

jkoendev: https://randomascii.wordpress.com/2012/03/21/intermediate-floating-point-precision/

jkoendev: oh i thought that the `DemoOfDanger` example applies here, but as dt and T are both `double` i am not so sure

giaf: The use of the math library should be made with care. The function `fabs` is in the standard math library, but `fmax` has only been introduced with the c99 extension to the math library. So it should not be used, as generally not available in the math library for embedded devices.

On top of that, I would keep it simple, and just check that `abs(mem->dt - in->T/opts->num_steps) < 1e-12` or something like that. Leave a note that the chosen threshold `1e-12` is valid since all quantities are doubles, but should be changed if we (ever) make a single precision version.

giaf: But getting back to @FreyJo 's comment, the quantity here should really be exactly equal, since `mem->dt` was computed in the first place using the exact same operation. Right?

giaf: @FreyJo actually here the check only wants to catch the case of non-inizialisation. What about replacing this check on `dt` and instead introducing somewhere in the memory a bool/int flag which is set when everything is properly inizialised?

FreyJo: @giaf thanks for pointing this out.
The check is there to verify that `mem->dt` (which is computed in precompute) is consistent with `T` and `num_steps`.
Since the latter ones can be always changed, i.e. also after the precompute step, we have to check for consistency.

FreyJo: The problem was however caused by something else.
I reverted the change on gnsf and appveyor still passes.
I suggest to merge https://github.com/acados/acados/pull/446 instead.

PR #446, opened 2019-08-17
Appveyor gnsf test

FreyJo:
Reverted the change on the gnsf check.
I suggest to merge this one instead of https://github.com/acados/acados/pull/443


jkoendev:
i think it now passes on appveyor because i changed the compiler to mingw64 7.3. Before it was MinGW 32-bit 5.3.0

giaf:
ok merging!

PR #448, opened 2019-08-19
Initialize integrators

FreyJo:
- [x]  direct enable direct initialization for Newton iterates in implicit nonlifted integrators from C
- [x] for matlab
- [x] implicit integrator, save last iterates as initial guess for subsequent calls (-> this is default for integrators as SQP submodules)


acados/sim/sim_erk_integrator.c
```diff
@@ -434,12 +434,42 @@ void sim_erk_opts_update(void *config_, void *dims, void *opts_)
...
+{
+    int status = ACADOS_SUCCESS;
+
+    if (!strcmp(field, "guesses"))
+    {
+        // no guesses/initialization in ERK
+    }
+    else
+    {
+        status = ACADOS_FAILURE;
```
giaf: Error handling is not handled systematically in acados, and just returning a chain of `ACADOS_FAILURE` would make very difficult to debug something like a wrong field.

Therefore, the way to go is rather to `printf("\nerror: sim_erk_memory_set_to_zero: wrong field %s\n", field);`, which tells exactly what and where is the issue.
Eventually all setters/getters should do so, but for now let's at least stick to this way for the new ones.

FreyJo: done, also for other setters & checks

giaf:
Right now the CI tests here fail (in a file that now does not exist any longer). Try to merge with the latest master (after my PR from yesterday) to see if it gets fixed.

acados/ocp_qp/ocp_qp_partial_condensing_solver.c
```diff
@@ -0,0 +1,446 @@
+/*
```
giaf: I just removed this file in my merged PR. Why is it getting back here?

giaf:
merging

PR #449, opened 2019-08-20
examples/c: removed chain_model.h

FreyJo:
As @giaf wrote here: https://github.com/acados/acados/pull/441#discussion_r314939037

> It is only templated by cmake. And actually now it is not even used at all, since I hand coded the initial values in a `.c` file instead. But I would keep this file as a reference for where the data in the `.c` file comes from (i.e. the `.m` script generated the `.txt` data files). If someone feels like directly generating the `.c` files from the `.m` script, then afterwards we can remove this header completely.

I think leaving the reference in the `.m` file as I did is sufficient.
And we should not longer let this file disturb peoples workflow.


giaf:
You can not just remove the entire `chain_model.h`, since the casadi functions defined there are used here https://github.com/acados/acados/blob/master/examples/c/nonlinear_chain_ocp_nlp.c#L59

If you really want to quick fix this, turn off the code generation here https://github.com/acados/acados/blob/master/examples/c/chain_model/chain_model.h.in#L25 and leave these 16 lines commented out

FreyJo:
Sorry that was quite sloppy..

Anyway I removed the lines completely from the .h file, since otherwise there will be always a git diff in this file after building with cmake, which is what I wanted to get rid off.

PR #450, opened 2019-08-20
acados_matlab: introduced acados_sim_detect_dims - detect dimensions …

FreyJo:
…from CasADi symbolics

I thought it would be nice to remove the necessity for the user to specify the model dimensions when using our Matlab/Octave interface.

The same should be possible for the OCP module.
But I would like to get some feedback first.

### DIFF
--- a/examples/matlab_mex/pendulum_on_cart_model/example_sim.m
+++ b/examples/matlab_mex/pendulum_on_cart_model/example_sim.m
@@ -46,9 +46,9 @@ end
 compile_mex = 'true';
 codgen_model = 'true';
 gnsf_detect_struct = 'true';
-%method = 'erk';
-%method = 'irk';
-method = 'irk_gnsf';
+method = 'erk';
+% method = 'irk';
+% method = 'irk_gnsf';
 sens_forw = 'true';
 num_stages = 4;
 num_steps = 4;
@@ -76,9 +76,6 @@ end
 if isfield(model, 'sym_p')
     sim_model.set('sym_p', model.sym_p);
 end
-sim_model.set('dim_nx', model.nx);
-sim_model.set('dim_nu', model.nu);
-
 
 if (strcmp(method, 'erk'))
 	sim_model.set('dyn_type', 'explicit');
@@ -90,7 +87,6 @@ else % irk irk_gnsf
 %	if isfield(model, 'sym_z')
 %		sim_model.set('sym_z', model.sym_z);
 %	end
-%	sim_model.set('nz', model.nz);
 end
 
 
--- a/examples/matlab_mex/test/test_sens_adj.m
+++ b/examples/matlab_mex/test/test_sens_adj.m
@@ -70,8 +70,6 @@ for integrator = {'irk_gnsf', 'irk', 'erk'}
     if isfield(model, 'sym_p')
         sim_model.set('sym_p', model.sym_p);
     end
-    sim_model.set('dim_nx', model.nx);
-    sim_model.set('dim_nu', model.nu);
 
 
     if (strcmp(method, 'erk'))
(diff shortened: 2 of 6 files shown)


giaf:
This is possible in the `sim` module, but in general not possible in the `ocp` module.

This is for the simple reason that the matlab/octave mex interface is partially numerical and partially symbolic (aka casadi based). Not all parts of the OCP are defined using casadi expressions or symbolics. As an example, `nb`, `ng` or `ns` do not enter in any casadi expression, so they can not be detected from there. Or, maybe even more significative, `ny` can be detected from casadi symbolics in the nonlinear least squares objective, but not in the linear least squares one.

Therefore, in my opinion, it does not make sense to have half dimensions detected and half not, and even without clear distinction of which ones can be detected. Detection form numerical matrices or vectors is also possible, but as in the case of linear vs nonlinear least squares, it would require the handling of numerous special cases.

My suggestion would be to 'force' the user to specify the dimensions explicitly as now (since in any case (s)he needs to explicitly create them to correctly set up the problem numerical matrices and vectors), and to use these dimensions as the 'true' ones when comparing with automatically detected dimensions from casadi symbolics or numerical matrices and vectors to find mismatches.

So I would keep the machinery you are developing, but also keep the dimensions and use it to check them.

Automatic detection of all dims would belong to a higher level interface, where the entire OCP is defined though casadi expressions and the problem structure is detected automatically (similar to the GNSF). So more for something like OpenOCL.

FreyJo:
> This is possible in the `sim` module, but in general not possible in the `ocp` module.
> 
I agree that the interface would change quite a bit, to do this for the `ocp` module.

> This is for the simple reason that the matlab/octave mex interface is partially numerical and partially symbolic (aka casadi based). Not all parts of the OCP are defined using casadi expressions or symbolics.

I am thinking about automatic structure detection for cost and constraint functions similar to the one for GNSF.
Probably, I would even connect them in a next step, such that cost and constraints can be handled efficiently by the linear output system of GNSF.

> As an example, `nb`, `ng` or `ns` do not enter in any casadi expression, so they can not be detected from there. Or, maybe even more significative, `ny` can be detected from casadi symbolics in the nonlinear least squares objective, but not in the linear least squares one.

Within an interface that does structure detection for all parts of the OCP, cost and constraints would be defined through CasADi expressions, just like the dynamics.
This would also result in a unified, generic interface for all types of cost functions that are treated by different modules in the core, which could look like the one for the external cost now.
From there, it can be detected into which category it fits, preferably it can be reformulated into the linear least squares form, but for sure it can be used with the external cost module.

In this case, I don't see a problem to detect dimensions

> Therefore, in my opinion, it does not make sense to have half dimensions detected and half not, and even without clear distinction of which ones can be detected.

I understand your dislike of inconsistency.
One intermediate solution would be to remove them only for the standalone integrator, which I did.

> So I would keep the machinery you are developing, but also keep the dimensions and use it to check them.

As I described above, in my current vision, removing the dimensions from the interface to the ocp, goes with introducing structure detection for the other modules and making the interface more purely CasADi/symbolics based. 

> Automatic detection of all dims would belong to a higher level interface, where the entire OCP is defined though casadi expressions and the problem structure is detected automatically (similar to the GNSF). So more for something like OpenOCL.

This is what I am thinking about.
For GNSF, it seemed like the structure detection fitted quite well into the acados matlab interface.
Since I was planning to concern myself further with structure detection within such a kind of interface, I am wondering if the acados Matlab interface is a good framework, which I think it is.
But if the feedback from the other acados developers is that it should not be done there, I will probably regard other options, like OpenOCL.

giaf:
This is a long discussion which we can have once back from holidays.

Just to add a few words now, my suggestion was not to do it in OpenOCL, but that it make sense to have it in that kind of high-level interface.

My feeling is that the best option would be to have a higher level matlab interface with structure detection on top of the current mex API. This higher level interface would also take care of detecting dimensions and passing them to the current API, where the checks would happen. In this way the current functioning/trusted interface would not been completely touched again, and more importantly structure detection would be decoupled and 'stand-alone', increasing modularity and maintainability.

PR #452, opened 2019-08-21
Mex interface: add checks, more setters

FreyJo:
Some work on the mex interface:
- add checks on dimensions to setters
- return error messages
- add more setters
- work towards multi phase formulation


FreyJo:
The next step imo would be to separate the `ocp_create`, into more functions.
The separation could be as follows:
1) Set up the dimensions, create all the C objects
2) Set all the numerical data. This function should somehow call the `ocp_set` function, such that checks are done and error messages are returned.

examples/matlab_mex/pendulum_on_cart_model/example_closed_loop.m
```diff
@@ -45,56 +43,48 @@
...
-sim_method = 'irk';
-sim_sens_forw = 'false';
+sim_method = 'irk'; % erk, irk, irk_gnsf
+sim_sens_forw = 'false'; % true, false
 sim_num_stages = 4;
 sim_num_steps = 4;
 % ocp
 param_scheme = 'multiple_shooting_unif_grid';
 ocp_N = 100;
-nlp_solver = 'sqp';
```
giaf: All these options commented out were good to quickly switch between them while developing, without having to type them all time time... I would keep them there...

examples/matlab_mex/pendulum_on_cart_model/example_closed_loop.m
```diff
@@ -45,56 +43,48 @@
...
 qp_solver = 'partial_condensing_hpipm';
-%qp_solver = 'full_condensing_hpipm';
+        % full_condensing_hpipm, partial_condensing_hpipm
 qp_solver_cond_N = 5;
-qp_solver_cond_ric_alg = 0;
-qp_solver_ric_alg = 0;
 qp_solver_warm_start = 0;
-ocp_sim_method = 'erk';
-%ocp_sim_method = 'irk';
+qp_solver_cond_ric_alg = 0; % HPIPM specific? what does it stand for?
```
giaf: this chooses if the hessian is factorized or not in the condensing

examples/matlab_mex/pendulum_on_cart_model/example_closed_loop.m
```diff
@@ -45,56 +43,48 @@
...
+qp_solver_ric_alg = 0; % HPIPM specific? what does it stand for?
```
giaf: yes HPIPM specific

interfaces/acados_matlab/acados_ocp.m
```diff
@@ -110,8 +110,8 @@ function set(varargin)
 			elseif nargin==4
 				obj = varargin{1};
 				field = varargin{2};
-				stage = varargin{3};
-				value = varargin{4};
```
giaf: Why? I would keep the value last, and put everything needed to route it before...

interfaces/acados_matlab/ocp_create.c
```diff
@@ -844,9 +845,10 @@ void mexFunction(int nlhs, mxArray *plhs[], int nrhs, const mxArray *prhs[])
 	memcpy(fieldnames[4],"out",sizeof("out"));
 	memcpy(fieldnames[5],"solver",sizeof("solver"));
 	memcpy(fieldnames[6],"sens_out",sizeof("sens_out"));
+	memcpy(fieldnames[7],"plan",sizeof("plan"));
```
giaf: Why do you need to keep the `plan`?

giaf:
I would suggest to have a look together once we are both back, it's a bit hard to fully understand all changes and ideas behind from here.

examples/matlab_mex/pendulum_on_cart_model/example_closed_loop.m
FreyJo: These examples are what we point the end users to. I think they should be as clear and compact as possible.
I was documenting each option with all possible values, e.g.
` sim_method = 'irk'; % erk, irk, irk_gnsf`
It is one line -> compact.
It lists all possible options -> clear.
Switching between options is still fairly easy, just copy a string in the comment behind and replace the one in code.

examples/matlab_mex/pendulum_on_cart_model/example_closed_loop.m
FreyJo: Interesting, but I dont have a clear picture of how it works.
I would add a comment:
`HPIPM specific: 0: dont factorize hessian in the condensing; 1: factorize`
okay?

interfaces/acados_matlab/acados_ocp.m
FreyJo: In Matlab, it is common to have optional arguments.
The convention is to put the optional argument in the end.
This is also, how the `ocp_set` function is implemented.
Imagine, inside within `ocp_set`, `value` to be `prhs[5]` or `prhs[6]` depending if `stage` is passed.
Thus, it is more intuitive for a Matlab user to call the function like this, i.e. with stage as last and optional argument.

interfaces/acados_matlab/ocp_create.c
FreyJo: I added it in order to be able to do checks depending on it.
It is good to remember what `plan` was chosen, when the object was created, since the user can change the other variables on the Matlab side.

examples/matlab_mex/pendulum_on_cart_model/example_closed_loop.m
giaf: this is not HPIPM specific: all solvers use some sort of condensing

giaf: If the Hessian is positive definite, then factorize the Hessian is usually faster. If the Hessian in not positive definite, then the factorization can not be used.

examples/matlab_mex/pendulum_on_cart_model/example_closed_loop.m
giaf: 10 lines more in the file are not going to make any difference about compactness, but anything that takes longer than commenting in or out a line is annoying for development.

That is the reason why that stuff exists at the beginning of the examples. It is not part of the interface, and actually the correct place to list all possible options is not there, but in the actual setter, e.g. https://github.com/acados/acados/blob/master/examples/matlab_mex/pendulum_on_cart_model/example_closed_loop.m#L222

interfaces/acados_matlab/ocp_create.c
giaf: In the mex interface it is possible to do the same form the matlab structs, and it is done like that in some other mex file in the interface.

Either way works here.

FreyJo: Checking against a matlab structs is just less save, because the user can modify it after `acados_sim` is created.

PR #473, opened 2019-09-17
template: fix casadi parametric fun

tmmsartor:
Fix template for parametric function integrator.


zanellia:
Thanks @tmmsartor!

FreyJo:
I guess to keep the symmetry with the Matlab interface, it would be better to always add `p` as an (sometimes empty) argument in the function generation.
So, we dont need the if in the template and not in the generating file
https://github.com/acados/acados/blob/master/interfaces/acados_template/acados_template/generate_c_code_explicit_ode.py

PR #474, opened 2019-09-17
Changed alpha for convergence and added swarming example

EnricaSo:
This pull request adds an example to the matlab folder. The cost is a Nonlinear LS function. Convergence only happens if alpha is changed from 1.0 to 0.2. 
Alpha can be found here: https://github.com/acados/acados/blob/master/acados/ocp_nlp/ocp_nlp_sqp.c#L206

### DIFF
--- a/examples/acados_matlab_octave/swarming/env.sh
+++ b/examples/acados_matlab_octave/swarming/env.sh
@@ -1,0 +1,72 @@
+#! /usr/bin/bash
+#
+# Copyright 2019 Gianluca Frison, Dimitris Kouzoupis, Robin Verschueren,
+# Andrea Zanelli, Niels van Duijkeren, Jonathan Frey, Tommaso Sartor,
+# Branimir Novoselnik, Rien Quirynen, Rezart Qelibari, Dang Doan,
+# Jonas Koenemann, Yutao Chen, Tobias Schöls, Jonas Schlagenhauf, Moritz Diehl
+#
+# This file is part of acados.
+#
+# The 2-Clause BSD License
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+# 1. Redistributions of source code must retain the above copyright notice,
+# this list of conditions and the following disclaimer.
+#
+# 2. Redistributions in binary form must reproduce the above copyright notice,
+# this list of conditions and the following disclaimer in the documentation
+# and/or other materials provided with the distribution.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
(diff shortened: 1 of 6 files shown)


giaf:
Hi Enrica!

Thanks for committing the swarm example. Now it is possible to set the SQP `step_length` directly from the matlab interface, using this option https://github.com/acados/acados/blob/master/interfaces/acados_matlab_octave/acados_ocp_opts.m#L108
As of now there are no examples of its use in the matalb examples, but basically you just need to set it in the opts like any other option.

Can you add this to your examples, so there is no need to change the SQP C source code to make them converge?

After this change I'll merge your PR :)

examples/matlab_mex/swarming/example_ocp.m
```diff
@@ -0,0 +1,354 @@
...
+time_tot = ocp.get('time_tot');
+time_lin = ocp.get('time_lin');
+time_reg = ocp.get('time_reg');
+time_qp_sol = ocp.get('time_qp_sol');
+
+fprintf('\nstatus = %d, sqp_iter = %d,  time_int = %f [ms] (time_lin = %f [ms], time_qp_sol = %f [ms], time_reg = %f [ms])\n', status, sqp_iter, time_tot*1e3, time_lin*1e3, time_qp_sol*1e3, time_reg*1e3);
+% time_ext = %f [ms], time_ext*1e3,
+
+stat = ocp.get('stat');
+if (strcmp(nlp_solver, 'sqp'))
```
FreyJo: for this stuff you can do `ocp.print('stat')` now :wink:

giaf:
> Hi Enrica,
> thanks for contributing!
> Could you remove the .png file and move the example to `/examples/acados_matlab_octave/swarming/` (the directory name changed lately).

Good catch, I didn't notice the old folder name!

EnricaSo:
Which branch should I work in? I am now on the master branch and if I ran the examples ```example_sim.m``` of the ```acados_matlab_octave``` folder, such as pendulum_on_cart_model/example_sim.m, I get 

```
Undefined function or variable 'acados_sim_model'.

Error in example_sim (line 74)
sim_model = acados_sim_model();```

FreyJo:
> Which branch should I work in? I am now on the master branch and if I ran the examples `example_sim.m` of the `acados_matlab_octave` folder, such as pendulum_on_cart_model/example_sim.m, I get
> 
> ```
> Undefined function or variable 'acados_sim_model'.
> 
> Error in example_sim (line 74)
> sim_model = acados_sim_model();
> ```

yes, master should work fine.
It seems like something went wrong with the environment variables.
The file `acados_sim_model` is in `interfaces/acados_matlab_octave`.

If you have a env.sh file that you source it should also include the updated path `interfaces/acados_matlab_octave` which was `interfaces/matlab_mex` before.

The env.sh file should look like this
https://github.com/acados/acados/blob/master/examples/acados_matlab_octave/pendulum_on_cart_model/env.sh

PR #476, opened 2019-09-18
Implement exact output of algebraic variables and corresponding sensitivities in IRK

FreyJo:
Outdated:
<del>There was a mistake when updating the `z` variables in the `ocp_nlp_sqp_[rti]` solver.
The fix is in this commit: https://github.com/acados/acados/commit/b394ad5e53067aa1d64aeb0cc5c2be7565350d95

fixed in https://github.com/acados/acados/pull/478 instead..

### DIFF
--- a/acados/ocp_nlp/ocp_nlp_sqp.c
+++ b/acados/ocp_nlp/ocp_nlp_sqp.c
@@ -199,21 +199,21 @@ void ocp_nlp_sqp_opts_initialize_default(void *config_, void *dims_, void *opts_
     opts->num_threads = ACADOS_NUM_THREADS;
 #endif
 
-	opts->ext_qp_res = 0;
+    opts->ext_qp_res = 0;
 
-	opts->qp_warm_start = 0;
+    opts->qp_warm_start = 0;
 
-	opts->step_length = 1.0;
+    opts->step_length = 1.0;
 
     // submodules opts
 
     // qp solver
     qp_solver->opts_initialize_default(qp_solver, dims->qp_solver, opts->qp_solver_opts);
-	// overwrite default
-	qp_solver->opts_set(qp_solver, opts->qp_solver_opts, "tol_stat", &opts->tol_stat);
-	qp_solver->opts_set(qp_solver, opts->qp_solver_opts, "tol_eq", &opts->tol_eq);
-	qp_solver->opts_set(qp_solver, opts->qp_solver_opts, "tol_ineq", &opts->tol_ineq);
-	qp_solver->opts_set(qp_solver, opts->qp_solver_opts, "tol_comp", &opts->tol_comp);
+    // overwrite default
+    qp_solver->opts_set(qp_solver, opts->qp_solver_opts, "tol_stat", &opts->tol_stat);
+    qp_solver->opts_set(qp_solver, opts->qp_solver_opts, "tol_eq", &opts->tol_eq);
+    qp_solver->opts_set(qp_solver, opts->qp_solver_opts, "tol_ineq", &opts->tol_ineq);
+    qp_solver->opts_set(qp_solver, opts->qp_solver_opts, "tol_comp", &opts->tol_comp);
 
     // regularization
     regularize->opts_initialize_default(regularize, dims->regularize, opts->regularize);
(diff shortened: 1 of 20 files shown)


acados/ocp_nlp/ocp_nlp_sqp.c
```diff
@@ -1229,10 +1229,9 @@ static void sqp_update_variables(void *config_, ocp_nlp_dims *dims, ocp_nlp_out
         blasfeo_daxpy(2*ni[i], alpha, mem->qp_out->t+i, 0, nlp_out->t+i, 0, nlp_out->t+i, 0);
 
         // linear update of algebraic variables using state and input sensitivity
-
         if (i < N)
         {
-            blasfeo_dgemv_t(nu[i]+nx[i], nz[i], 1.0, mem->dzduxt+i, 0, 0, nlp_out->ux+i, 0, 1.0, mem->z_alg+i, 0, nlp_out->z+i, 0); 
+            blasfeo_dgemv_t(nu[i]+nx[i], nz[i], 1.0, mem->dzduxt+i, 0, 0, nlp_out->ux+i, 0, 0.0, mem->z_alg+i, 0, nlp_out->z+i, 0);
```
zanellia: Is this correct? it should be z^+ <- z + dzdx * delta_x + dzdu * delta_u

giaf: where does the minus sign come from?

zanellia: oh sorry, it's an arrow, not a minus

giaf:
> There was a mistake when updating the `z` variables in the `ocp_nlp_sqp_[rti]` solver.
> The fix is in this commit:
> [b394ad5](https://github.com/acados/acados/commit/b394ad5e53067aa1d64aeb0cc5c2be7565350d95)

Why should it be wrong with the 1.0?
This is what we discussed some day ago: before it was simply copied `mem->z_alg` into `nlp_out->z`, i.e. the algebraic variables would be 1 iteration behind.
The exact ones would require to call the integrators once more, which we agree to be too expensive.
So the agreed solution was to make a linear approximation, where the 1-iteration-behind value `mem->z_alg` is updated (i.e. the 1.0 in the `dgemv`) using the sensitivities of `z` wrt `x` and `u`.

In general, what is the purpose of the PR? What is `exact_z_output` doing?

acados/ocp_nlp/ocp_nlp_sqp.c
zanellia: btw, these updates are both wrong, aren't they?

giaf: Me and @zanellia had a chat about the correct fix, and I have an idea for a test to check if it is indeed correct. I will implement it in another PR soon.

FreyJo: In my understanding `nlp_out->ux` does not contain `delta_x, delta_u`, but `x, u` looking at this line:
https://github.com/acados/acados/blob/2f3f19232ab14a69eb55179d5ed666db224b6b11/acados/ocp_nlp/ocp_nlp_sqp.c#L1215
Now thinking about this again, the correct update should be:
``blasfeo_dgemv_t(nu[i]+nx[i], nz[i], alpha, mem->dzduxt+i, 0, 0, mem->qp_out->ux+i, 0, 1.0, mem->z_alg+i, 0, nlp_out->z+i, 0); 
``
Does this make sense to you?

FreyJo:
In contrast to extrapolating the `z` and `dz_dxu` in the shooting nodes to the beginning of the simulation. There is the possibility to solve f(x_0, u_0, xdot_0, z_0) for `xdot_0, z_0` with few Newton iterations and then apply implicit function theorem at this point.
I was implementing this option in `sim_irk`, in order to get more accurate `z, S_alg` and called it `exact_z_output`.

acados/ocp_nlp/ocp_nlp_sqp.c
giaf: Yes that's the correct one. Anyway, I quickly want to implement this fix and a check in another PR, and then you can merge it in your.

giaf:
So, me and @zanellia had a chat about the different approaches.

The "extrapolate approach" has the advantage of giving values of `z` and `S_alg` which are consistent with the collocation interpolating polynomial (which eventually would be the ones used to evaluate the continuous cost, for example, and which define the discrete problem). Another advantage is that they don't require additional function evaluations or Newton steps. The drawback is that these polynomial are constructed to give high accuracy at the end point, but they are less accurate when used for interpolation at other points.

The approach you implement here gives values of `z` and `S_alg` which are not consistent with the interpolating polynomial, but which have better consistency w.r.t. the continuous time problem (i.e. w.r.t. the implicit DAE equation). And they may give better consistency in the evaluation of constraints, since these are in any case only enforced at the shooting nodes.

In that regard, before merging the PR, we would like to make some additional investigation. The idea would be to take one of the nonlinear examples (e.g. the pendulum on cart), replace some of the states in the ODE with some algebraic equations, and then add the constraint that `z=x` for the replaced states. In this way, it is possible to build a nonlinear example which can also be solved to high accuracy as an ODE. Using this example, it can be tested (besides the correctness of both implementations in a nonlinear case) which approach is better suited for which case (e.g. if indeed your implementation for this PR gives better consistency in the evaluation of the constraints `z=x`).

How does it sound?

interfaces/acados_matlab_octave/ocp_set.c
```diff
@@ -422,6 +422,13 @@ void mexFunction(int nlhs, mxArray *plhs[], int nrhs, const mxArray *prhs[])
         int nlp_solver_max_iter = (int) value[0];
         ocp_nlp_opts_set(config, opts, "max_iter", &nlp_solver_max_iter);
     }
+    else if (!strcmp(field, "qp_warm_start"))
```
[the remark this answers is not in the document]
FreyJo: yes. Just in the `ocp_create` though. I added it to `ocp_set`

FreyJo:
I did what @giaf and @zanellia suggested and added the file
`examples/acados_matlab_octave/pendulum_on_cart_model/experiment_dae_formulation.m`,
which shows that the exact_z_output option works for IRK and indeed gives more accurate results compared to the extrapolation.
The file produces the following output:

```
Constraint values

ODE 		DAE-exact_z	DAE-extrapolation
-8.9974e-65	-8.9974e-65	-3.3241e-55
-7.9639e-02	-7.9639e-02	-7.9639e-02
-1.2577e+00	-1.2577e+00	-1.2577e+00
-6.1291e+00	-6.1291e+00	-6.1291e+00
-1.7759e+01	-1.7759e+01	-1.7759e+01
-3.6683e+01	-3.6683e+01	-3.6683e+01
-4.0000e+01	-4.0000e+01	-3.5929e+01
-4.0000e+01	-4.0000e+01	-4.2980e+01
-3.3693e+01	-3.3693e+01	-3.3932e+01
-2.0004e+01	-2.0004e+01	-2.0177e+01
-1.0459e+01	-1.0459e+01	-1.0577e+01
-3.9678e+00	-3.9678e+00	-4.0504e+00
7.2693e-01	7.2693e-01	6.6237e-01
4.6870e+00	4.6870e+00	4.6257e+00
8.7970e+00	8.7970e+00	8.7262e+00
1.3865e+01	1.3865e+01	1.3772e+01
2.0650e+01	2.0650e+01	2.0525e+01
2.9644e+01	2.9644e+01	2.9483e+01
4.0000e+01	4.0000e+01	4.0055e+01
4.0000e+01	4.0000e+01	3.8523e+01
2.8029e+01	2.8029e+01	2.8130e+01
1.1652e+01	1.1652e+01	1.2125e+01
2.3482e+00	2.3482e+00	2.6055e+00
-3.3184e-01	-3.3184e-01	-2.6533e-01
-3.9900e-01	-3.9900e-01	-3.9964e-01
-1.0687e-01	-1.0687e-01	-1.1457e-01
-1.7618e-03	-1.7618e-03	-3.0164e-03
-2.0067e-02	-2.0067e-02	-1.7092e-02
-5.1244e-02	-5.1244e-02	-4.7702e-02
-5.8255e-02	-5.8255e-02	-5.5764e-02
-4.7494e-02	-4.7494e-02	-4.6137e-02
-3.2277e-02	-3.2277e-02	-3.1645e-02
-1.9774e-02	-1.9774e-02	-1.9500e-02
-1.1508e-02	-1.1508e-02	-1.1386e-02
-6.6024e-03	-6.6024e-03	-6.5411e-03
-3.8082e-03	-3.8082e-03	-3.7740e-03
-2.1873e-03	-2.1873e-03	-2.1703e-03
-1.1694e-03	-1.1694e-03	-1.1684e-03
-4.3914e-04	-4.3914e-04	-4.5546e-04
1.7053e-04	1.7053e-04	1.3495e-04


Constraint violations
ODE 		DAE-exact_z	DAE-extrapolation
7.1054e-15	7.1054e-15	2.9805e+00
```

giaf:
looks good, merging!

PR #479, opened 2019-09-25
Mex conditional compilation

FreyJo:
mex: compile interface if not available even if `compile_mex = false`
set `compile_mex` default to `false`
avoid generating `_hess` function, when not needed (irk & irk_gnsf)


giaf:
It looks like the previous PR is in the diff. Can you try to merge master into this PR?

FreyJo:
Done

PR #480, opened 2019-09-26
fix SQP related TODOs

FreyJo:
This includes:

- warm start setter for qpOASES,
- option to warm start first qp in SQP (default false)
- improved error messages
- fixed some  sqp related TODOs
- cleaning/formatting

Feedback Welcome!

On top of this one https://github.com/acados/acados/pull/479


acados/ocp_nlp/ocp_nlp_common.c
```diff
@@ -146,8 +146,7 @@ ocp_nlp_config *ocp_nlp_config_assign(int N, void *raw_memory)
  * dims
  ************************************************/
 
-// TODO(oj): should be static, but used by current ocp_nlp c++ interface
-int ocp_nlp_dims_calculate_size_self(int N)
+static int ocp_nlp_dims_calculate_size_self(int N)
```
giaf: The reason it was there, was that Robin needed in the C++ interface. If not it is not used any longer in there, then we can just merge it again with `dims_calculate_size`.

acados/ocp_nlp/ocp_nlp_common.c
```diff
@@ -209,8 +208,7 @@ int ocp_nlp_dims_calculate_size(void *config_)
 
 
 
-// TODO(oj): should be static, but used by current ocp_nlp c++ interface
-ocp_nlp_dims *ocp_nlp_dims_assign_self(int N, void *raw_memory)
+static ocp_nlp_dims *ocp_nlp_dims_assign_self(int N, void *raw_memory)
```
giaf: same for this (and all other `_self`s)

acados/ocp_nlp/ocp_nlp_sqp.c
```diff
@@ -1348,10 +1314,6 @@ int ocp_nlp_sqp(void *config_, void *dims_, void *nlp_in_, void *nlp_out_,
 #if defined(ACADOS_WITH_OPENMP)
     #pragma omp for
 #endif
-    for (ii = 0; ii < N; ii++)
-    {
-        config->dynamics[ii]->model_set(config->dynamics[ii], dims->dynamics[ii], nlp_in->dynamics[ii], "T", nlp_in->Ts+ii);
```
giaf: Why did you remove this?

acados/ocp_nlp/ocp_nlp_sqp.c
```diff
@@ -1477,10 +1431,18 @@ int ocp_nlp_sqp(void *config_, void *dims_, void *nlp_in_, void *nlp_out_,
         qp_info *qp_info_;
         ocp_qp_out_get(mem->qp_out, "qp_info", &qp_info_);
         nlp_out->qp_iter = qp_info_->num_iter;
+        // printf("\nqp_iter = %d, sqp_iter = %d, max_sqp_iter = %d\n", nlp_out->qp_iter, sqp_iter, opts->max_iter);
         qp_iter = qp_info_->num_iter;
 
+        // save statistics of last qp solver call
```
giaf: Better to always save the statistics for the qp here (i.e. remove the ones above about `qp_status` and `qp_iter`).

But you also need to add a safeguard similar to the one at (new) line 1448: this is because the memory for stat is fixed when the memory is created, and it is for `stat_m` lines. If the users afterwards in creases the max sqp iter, this still shouldn't result in overwriting memory around.

acados/ocp_nlp/ocp_nlp_sqp.c
```diff
@@ -1477,10 +1431,18 @@ int ocp_nlp_sqp(void *config_, void *dims_, void *nlp_in_, void *nlp_out_,
...
+            mem->stat[mem->stat_n*(sqp_iter+1)+4] = qp_status;
+            mem->stat[mem->stat_n*(sqp_iter+1)+5] = qp_iter;
+        }
+
         // compute external QP residuals (for debugging)
-        if(opts->ext_qp_res)
+        if (opts->ext_qp_res)
         {
             ocp_qp_res_compute(mem->qp_in, mem->qp_out, work->qp_res, work->qp_res_ws);
             if (sqp_iter+1 < mem->stat_m)
```
giaf: like this safeguard

acados/ocp_nlp/ocp_nlp_sqp.c
```diff
@@ -1581,17 +1543,22 @@ int ocp_nlp_sqp_precompute(void *config_, void *dims_, void *nlp_in_, void *nlp_
...
 
-    // extract dims
     int N = dims->N;
     int status = ACADOS_SUCCESS;
 
     int ii;
 
-    // TODO(fuck_lint) checks
-    // TODO(fuck_lint) flag to enable/disable checks
+    // TODO(giaf) flag to enable/disable checks
```
giaf: this is not for `giaf`...

acados/ocp_nlp/ocp_nlp_sqp_rti.c
```diff
@@ -1301,10 +1269,6 @@ int ocp_nlp_sqp_rti(void *config_, void *dims_, void *nlp_in_, void *nlp_out_,
 #if defined(ACADOS_WITH_OPENMP)
     #pragma omp for nowait
 #endif
-    for (int ii = 0; ii < N; ii++)
```
giaf: why are you removing this here too?

acados/utils/math.c
```diff
@@ -1092,41 +1093,6 @@ void acados_eigen_decomposition(int dim, double *A, double *V, double *d, double
 
 
 
-void eigen_decomposition(int dim, double *A, double *V, double *d)
```
giaf: are you sure this is not used in the regularization???

FreyJo: Yes. I found that all calls were replaced with calls to `acados_eigen_decomposition` which doesnt use `calloc`.

acados/ocp_nlp/ocp_nlp_common.c
FreyJo: The C++ interface cannot be used with recent acados versions anyway.
So, I just wanted to make it clear it should just be used internally.
Merging with non `self` ones can still be done later.

acados/ocp_nlp/ocp_nlp_sqp.c
FreyJo: because of incompatibility with `irk_gnsf`.
I put it back and added a note since it might brake the templating interface

giaf:
ok good

PR #487, opened 2019-10-04
Present the mex interface

FreyJo:
- Fix compiler warnings
- Mex: improve error messages, fix some possibilities to crash Matlab

### DIFF
--- a/interfaces/acados_matlab_octave/sim_create.c
+++ b/interfaces/acados_matlab_octave/sim_create.c
@@ -60,15 +60,14 @@ void mexFunction(int nlhs, mxArray *plhs[], int nrhs, const mxArray *prhs[])
 
     /* LHS */
     // field names of output struct
-    char *fieldnames[6];
+    char *fieldnames[7];
     fieldnames[0] = (char*)mxMalloc(50);
     fieldnames[1] = (char*)mxMalloc(50);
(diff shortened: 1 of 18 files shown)


interfaces/acados_matlab_octave/sim_create.c
```diff
@@ -60,34 +60,32 @@ void mexFunction(int nlhs, mxArray *plhs[], int nrhs, const mxArray *prhs[])
...
     memcpy(fieldnames[0],"config",sizeof("config"));
     memcpy(fieldnames[1],"dims",sizeof("dims"));
     memcpy(fieldnames[2],"opts",sizeof("opts"));
     memcpy(fieldnames[3],"in",sizeof("in"));
     memcpy(fieldnames[4],"out",sizeof("out"));
     memcpy(fieldnames[5],"solver",sizeof("solver"));
-
-    // memcpy(fieldnames[6],"method",sizeof("method"));
-
+    memcpy(fieldnames[6],"method",sizeof("method"));
```
giaf: What is method? Is it the same as 'plan' for the ocp?

FreyJo: For `sim` the `plan` is destroyed after this function.
So, I dont store a pointer to a C object but instead the integer corresponding to the `enum` of the `sim_solver_t`.

PR #492, opened 2019-10-13
Adding nonlinear least-squares to templates and Python interface

zanellia:


zanellia:
@giaf @FreyJo would you have a quick look? Just to keep you updated with what is going on with Python ;)

examples/acados_template/python/pendulum_example/generate_c_code.py
```diff
@@ -98,6 +116,20 @@
 Vx_e[3,3] = 1.0
 
 nlp_cost.Vx_e = Vx_e
+if FORMULATION == 'NLS':
+    x = SX.sym('x', 4, 1)
+    u = SX.sym('u', 1, 1)
+    ocp.cost_r.expr = vertcat(u, x)
```
FreyJo: Note that this is called `cost_expr_y`, which might be more consistent with the LS cost type, since the residual is `y - y_ref `.

interfaces/acados_template/acados_template/generate_solver.py
```diff
@@ -35,6 +35,8 @@
 from .generate_c_code_explicit_ode import *
 from .generate_c_code_implicit_ode import *
 from .generate_c_code_constraint import *
+from .generate_c_code_nls_cost import *
+from .generate_c_code_nls_cost_e import *
```
FreyJo: I think this function is not there (yet).
It might be easier to gather the functionality in one function.

examples/acados_template/python/pendulum_example/generate_c_code.py
zanellia: I agree, I am planning to move this auxiliary class into acados.cost anyway, stay tuned :P

interfaces/acados_template/acados_template/generate_solver.py
zanellia: Thanks! I have pushed the missing file.

examples/acados_template/python/pendulum_example/generate_c_code.py
giaf: Agree with Jonathan. The advantage of having `expr_y - y_ref` instead of `r` is that `y_ref` is numerical and therefore it can be changed easily when there are changes in the reference (which is a common case). The use of `r` requires the use of parameters instead.

If you agree with that, I guess it's easier to code `expr_y` from the beginning, instead of coding first as `r` and having to change it later on ;)

giaf:
I don't have further comments. But note that there are conflicts to be resolved.

examples/acados_template/python/pendulum_example/generate_c_code.py
zanellia: The formulation is already r - y_ref. I guess I can just change r to expr_y ;)

PR #501, opened 2019-10-21
Mex getting started

FreyJo:
- getting started folder
- set up CasADi automatically for Matlab users
- update docs, README
- add TEX file with problem formulation used in the MEX interface

@giaf @zanellia 
Could you please have a look at  [this pdf](https://github.com/FreyJo/acados/blob/mex_getting_started/docs/problem_formulation/problem_formulation_ocp_mex.pdf), which should give a complete description of the OCP solved with the `MEX` interface.

I was omiting the upper/lower slack separation, but I am convinced that with the absolute value in the slack L1 penalty and the fact that there are no bounds on the slacks, it should be equivalent.


FreyJo:
I just updated the docs again, using the lower/upper slack formulation that is actually used.
Any further comments? @giaf  @zanellia

zanellia:
That's great! Now that there is some more documentation on the formulation we should also make the docs page available on the main README.md. The formulation is by the way 90% compatible with the one used in the Python interface. Just two remarks: 

1) The constraints are not imposed at t=0 at the moment, right? Then you should use (0,T) and not [0,T) (I personally think they should be, maybe we can talk about it again with @giaf as well? :P).   
2) You are missing a square on the norms for the cost terms.

FreyJo:
Thanks for checking @zanellia 

> 1.The constraints are not imposed at t=0 at the moment, right?
I just checked, and this is only the case for the bounds on x constraints.
The others are also enforced at stage 0, see [e.g here](https://github.com/acados/acados/blob/8ede01e341e79786eba317e2a5bbff27f0faa1e7/interfaces/acados_matlab_octave/ocp_create.c#L1665)

I changed it to t\in (0, T) for these constaints.

> You are missing a square on the norms for the cost terms.
thanks for spotting!

Also, I added a link to the main Readme.
I agree it is nearly the same for the Python interface.
The main difference is that it uses `idx*` instead of `J*`, which would probably be easier to change compared to adding another problem formulation like this one.
Are there other differences that come to your mind?

zanellia:
> Thanks for checking @zanellia
> 
> > 1.The constraints are not imposed at t=0 at the moment, right?
> 
> I just checked, and this is only the case for the bounds on x constraints.
> The others are also enforced at stage 0, see [e.g here](https://github.com/acados/acados/blob/8ede01e341e79786eba317e2a5bbff27f0faa1e7/interfaces/acados_matlab_octave/ocp_create.c#L1665)
> 
> I changed it to t\in (0, T) for these constaints.
> 
> > You are missing a square on the norms for the cost terms.
> 
> thanks for spotting!
> 
> Also, I added a link to the main Readme.
> 
> I agree the formulation is nearly the same for the Python interface.
> The main difference is that it uses `idx*` instead of `J*`, which would probably be easier to change compared to adding another problem formulation like this one.
> Are there other differences that come to your mind?

I guess it would make most sense either to enforce all of the constraints or none at t=0, or I am missing something? Regarding J* instead of idx*, I totally agree. Actually, it should already be possible to use the J* setters, simply there are no attributes associated with them in the Python class nor in the JSON file, since I decided to keep only idx* there. I don't recall any other difference in the formulation at the moment, but, if there is any, I should just change the interface ;)

giaf:
> I guess it would make most sense either to enforce all of the constraints or none at t=0, or I am missing something?

There is an equality constraints on x0 and therefore its value can not be decided by the optimizer: what is the point in having box constraints on it? The optimizer can not enforce them.

Additionally, from a practical point of view, these equality constraints are enforced as equal upper and lower bounds, so no other bounds can be enforced at the same stage.

giaf:
@FreyJo from the table, `constr_x0` is not required: if it is not there, the problem is a MHE instead of MPC

docs/interfaces/index.md
```diff
@@ -37,16 +37,22 @@ Afterwards, launch Matlab or Octave from the same shell.
 
 If you want to run the examples in a different folder, please close the current shell and open a new one to repeat the procedure: this ensures the correct setting of the environment variables.
 
+### acados MATLAB (going embedded)
+There is the option to generate embeddable `C` code from Matlab.
+The workflow uses the same templates as the Python interface (see below) and the `Tera` renderer.
+After creating an acados solver `ocp`, you can use the routine `ocp.generate_c_code` to generate `C` code which can be used for embedded applications.
 
+Note: This part of the MATLAB/Octave interface does not support all features of the one mentioned before.
```
giaf: `... does not support (yet) all features ...`

docs/interfaces/index.md
```diff
@@ -37,16 +37,22 @@ Afterwards, launch Matlab or Octave from the same shell.
...
 
 
+## acados embedded - Python
+
 `acados_template` is a Python package that can be used to specify optimal control problems from Python and to generate self-contained C code that uses the acados solvers to solve them.
 In comparison with the MATLAB interface for rapid prototyping (see above), it supports less features, but it allows the user to generate a self-contained C library  
 that can be easily deployed on an embedded system.
 
-The framework is based on templated C files which are rendered from Python using the templating engine `Jinja2`.
+The framework is based on templated C files which are rendered from Python using the templating engine `Jinja2` or `Tera`.
```
giaf: @zanellia did you switch to only `Tera` now?

examples/acados_matlab_octave/getting_started/check_acados_requirements.m
```diff
@@ -0,0 +1,99 @@
+function check_acados_requirements()
```
giaf: cool :)

giaf:
It looks good for me!

FreyJo:
@giaf thanks for checking!
I addressed your feedback.
Merging now

PR #502, opened 2019-10-22
Tera template edits

FreyJo:
@zanellia I would appreciate a review from the template master :wink:


zanellia:
It looks great! It's mostly polishing, right? Does it need any testing on the Python examples? (I guess not)

FreyJo:
Thanks!
No, I am quite sure it cant break python stuff.

PR #503, opened 2019-10-23
Quickfix templates: set params for all functions

zanellia:
1) set parameters for all parametric functions on a per-stage basis
2) add Python setter (`ocp.set(<stage>, "p", <value>)`)


zanellia:
@FreyJo @giaf, could you have a quick look and merge? :)

examples/acados_template/python/soft_constraints/generate_c_code.py
```diff
@@ -235,8 +235,8 @@ def export_terminal_nonlinear_constraint():
 ocp.solver_config.nlp_solver_type = 'SQP'
 
 # set header path
-ocp.acados_include_path  = '/usr/local/include'
-ocp.acados_lib_path      = '/usr/local/lib'
+ocp.acados_include_path  = '~/acados/include'
```
FreyJo: I guess it would be great to initialize the paths in the constructor, i.e. here:
https://github.com/acados/acados/blob/e5865e45d2d4b2c51fc735ab2b55b6efe4b8a19d/interfaces/acados_template/acados_template/acados_ocp_nlp.py#L1281
Is it possible to get the `ACADOS_ROOT_DIR` there?

interfaces/acados_template/acados_template/c_templates/main.in.c
```diff
@@ -94,7 +76,6 @@ external_function_param_casadi * sim_impl_dae_fun_jac_x_xdot_z;
 external_function_param_casadi * sim_impl_dae_jac_x_xdot_u_z;
```
FreyJo: In tera the `acados_sim_solver` is not used in the `main.in.c` file.
I think it would be great to track the differences in the tera, jinja2 template.
I guess the plan is still to use only use `tera` eventually.

Beside the fact that acados_sim should be used in the same way in the main templates, do you remember something else?
It would be good to track these todos in the `ROADMAP.md`.

zanellia: Yeah, I want to get rid of Jinja as soon as IFAC deadline is over :P

examples/acados_template/python/soft_constraints/generate_c_code.py
zanellia: not so easy, I think. The Python package can get the path to its install location, but not the path to  the source code, which would then provide the location to default install location. An alternative would be to have a local script that sets an environment variable or to have the same variable set on a system-wide basis, but I still prefer the current approach to both of these alternatives. @tmmsartor maybe has some (additional) suggestions here? :)

PR #499, opened 2019-10-16
Fix SCQP

zanellia:
This PR fixes and extends the implementation of convex-over-nonlinear constraints of the form h(r(x,u)) (where h is the convex outer part and r the nonlinear inner one) in the BGHP module: 

1) support general outer convex function (not just Euclidean norm)
2) support for h being a general convex map  to $R^{n_h}$ with $n_h \geq 1$
3) fix multipliers computation

PR #506, opened 2019-10-30
Fix conflicts in SCQP PR (replaces #499)

zanellia:
This PR fixes and extends the implementation of convex-over-nonlinear constraints of the form \phi(r(x,u)) (where \phi is the convex outer part and r the nonlinear inner one) in the BGP (previously BGHP) module: 

1) support general outer convex function (not just Euclidean norm)
2) support for h being a general convex map  to $R^{n_\phi}$ with $n_\phi \geq 1$
3) fix multipliers computation 

Additional things added on the way:

1) added Python examples to unit tests
2) updated Tera templates which will be the default rendering engine 
3) config -> options in Python interface
4) fixed modularity of constraints


zanellia:
I think I am done with this :) @giaf I have implemented the changes you required in #499. As usual, if @FreyJo wants too have a look too he can add his comments ;)

examples/c/CMakeLists.txt
```diff
@@ -245,13 +245,13 @@ add_executable(simple_dae_example simple_dae_example.c
...
-# -------------------- pendulum_scqp
-add_executable(pendulum_scqp pendulum_scqp.cpp
-    pendulum_model/pendulum_ode_expl_vde_forw.c
-    pendulum_model/constraint.c
-    pendulum_model/position.c)
-target_link_libraries(pendulum_scqp acados)
-add_test(pendulum_scqp pendulum_scqp)
+# # -------------------- pendulum_scqp
+# add_executable(pendulum_scqp pendulum_scqp.cpp
+#     pendulum_model/pendulum_ode_expl_vde_forw.c
```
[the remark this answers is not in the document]
zanellia: I am afraid so, we could try to revive it, although the code generation script for that example is versioned.... Anyway, SCQP is used in two of the Python examples that run on Travis :)

interfaces/acados_template/setup.py
```diff
@@ -59,17 +59,6 @@
        'c_templates/Makefile.in',
        'c_templates/model.in.h',
        'c_templates/main.in.h',
-       'c_templates/acados_solver.in.c',
-       'c_templates/acados_solver.in.h',
-       'c_templates/acados_sim_solver.in.c',
```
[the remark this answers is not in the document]
zanellia: Moving to Tera ;)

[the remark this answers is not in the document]
zanellia: I have removed the leftover Jinja templates here aec87609.

examples/acados_python/test/test_data/generate_c_code_out_LS_FULL_CONDENSING_HPIPM_ERK_SQP_RTI.json
```diff
@@ -0,0 +1,906 @@
+{
```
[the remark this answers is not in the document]
zanellia: It's a different formatting in test/main_test.py. I thought it would be more readable like this.

acados/ocp_nlp/ocp_nlp_constraints_bgh.c
```diff
@@ -374,6 +374,10 @@ void ocp_nlp_constraints_bgh_dims_get(void *config_, void *dims_, const char *fi
     {
         ocp_nlp_constraints_bgh_get_nh(config_, dims_, value);
     }
+    else if (!strcmp(field, "nphi"))
```
giaf: this is not needed any more and can be removed => modularity

acados/ocp_nlp/ocp_nlp_constraints_bgp.c
```diff
@@ -0,0 +1,1253 @@
...
+        printf("\nerror: compute_hess!=0 not supported (yet) in ocp_nlp_constraints_bgp\n");
+        exit(1);
+    }
+
+    return;
+}
+
+
+
+void ocp_nlp_constraints_bgp_compute_fun(void *config_, void *dims_, void *model_, void *opts_, void *memory_, void *work_)
```
giaf: Could you implement this, since how this constraint module is fixed?

This is just computing the value of the constraint function. It may be good to add a new external function type to return just the function (no jac or hess), one for `phi` and one for `r`.

acados/ocp_nlp/ocp_nlp_constraints_bgp.c
```diff
@@ -0,0 +1,1253 @@
...
+    config->memory_set_idxs_ptr = &ocp_nlp_constraints_bgp_memory_set_idxs_ptr;
+    config->workspace_calculate_size = &ocp_nlp_constraints_bgp_workspace_calculate_size;
+    config->initialize = &ocp_nlp_constraints_bgp_initialize;
+    config->update_qp_matrices = &ocp_nlp_constraints_bgp_update_qp_matrices;
+    config->compute_fun = &ocp_nlp_constraints_bgp_compute_fun;
+    config->config_initialize_default = &ocp_nlp_constraints_bgp_config_initialize_default;
+
+
+    return;
+}
```
giaf: It is nearly impossible to check the changes to the file, so ok on trust ;)

examples/acados_python/pendulum_example/generate_c_code.py
```diff
@@ -144,21 +144,19 @@
 nlp_con.idxbu = np.array([0])
 
 # set QP solver
-# ocp.solver_config.qp_solver = 'PARTIAL_CONDENSING_HPIPM'
```
giaf: great :)

giaf:
The C part, besides the few comments, it looks good. Anyway, we discussed enough in person about ;) The changes on the bgp files are difficult to judge without diff working properly, though.

The python/template part I can not say much.

acados/ocp_nlp/ocp_nlp_constraints_bgp.c
zanellia: I don't know what happens with with git when we have to fix PRs locally. It would be great to understand what is going on :) Anyway, if you checkout the PR locally and rename `ocp_nlp_constraints_bgp.h -> ocp_nlp_constraints_bghp.c` and do `git diff 00071c4013af0d -- ocp_nlp_constraints_bghp.c` you can see all the changes (same for `ocp_nlp_constraints_bgp.h`)

acados/ocp_nlp/ocp_nlp_constraints_bgp.c
zanellia: I gave it a shot in 95338d58be, but I don't know how to test it :P Why do we need a function for r too?

acados/ocp_nlp/ocp_nlp_constraints_bgh.c
zanellia: good catch, fixed.

acados/ocp_nlp/ocp_nlp_constraints_bgp.c
giaf: Since the nonlinear function is $\phi(r(x,u))$, I (apparently wrongly) assumed that $\phi$ was defined accordingly also in the C code, i.e. $\phi: R^{nr} \Rightarrow R^{nphi}$. But in your code it looks like it is defined as the composite function $\phi \circ r: R^{nx+nu} \Rightarrow R^{nphi}$.

Can you clarify on this point?

acados/ocp_nlp/ocp_nlp_constraints_bgp.c
```diff
@@ -0,0 +1,1334 @@
...
+        // (dphidu + dphidz*dzdu)*(u - \bar{u})  
+        
+        // update DCt
+        blasfeo_dgemm_nn(nu+nx, nphi, nz, 1.0, memory->dzduxt, 0, 0, &work->tmp_nz_nphi, 0, 0, 0.0, &work->tmp_nv_nphi, 0, 0, &work->tmp_nv_nphi, 0, 0);
+        blasfeo_dgead(nu+nx, nphi, 1.0, &work->tmp_nv_nphi, 0, 0, memory->DCt, ng, 0);	
+        // update memory->fun	
+        blasfeo_dgemv_t(nu+nx, nphi, -1.0, &work->tmp_nv_nphi, 0, 0, memory->ux, 0, 1.0, &memory->fun, 0, &memory->fun, 0);
+    }
+
+    if (nr > 0)
```
giaf: How can it be nphi>0 and nr=0?

acados/ocp_nlp/ocp_nlp_constraints_bgp.c
zanellia: `nl_constr_phi_jac` takes as input $x$, $u$, $z$ and $p$ and outputs $\phi(r(x,u,z))$, $\frac{d \phi(r(x,u,z))}{d x,u}$, $\frac{d \phi(r(x,u,z))}{d z}$ and the Hessians $\frac{d^2 \phi_i}{d r}$, which together with $\frac{d r(x,u,z)}{d x,u}$ are the quantity necessary to build the constraint linearization and the SCQP Hessian contribution.

acados/ocp_nlp/ocp_nlp_constraints_bgp.c
giaf: Just wait a moment before rushing with changes, I think it should be discussed a bit more the overall approach :p

acados/ocp_nlp/ocp_nlp_constraints_bgp.c
giaf: Thanks for the clarification. I think it is a bit confusing, as there is in my opinion a clash between $\phi$ and the composite function $\phi \circ r$. Also the way this is implemented made me believe something wrong at fist sight (shame on me for not digging deeper in the code before). I will explain in a new thread below.

acados/ocp_nlp/ocp_nlp_constraints_bgp.c
zanellia: OK, but simply $nr$ cannot be equal to zero.

acados/ocp_nlp/ocp_nlp_constraints_bgp.c
zanellia: I thought mathematically speaking this is consistent since $\phi$ only denotes the image of the function (which is the same for $\phi(r(x,u,z))$ and $\phi(r)$), but we can make it more explicit/clear if you like.

Regarding the whole approach, this should be following Robin's notation in the most general form I could imagine :P

giaf:
The aim here is to implement nonlinear constraints in the form $\phi ( r (x,u) )$, where for the time being we can disregard convexity of $\phi$. There are two ways to properly implement this
- implementation of the two functions (and relative derivatives) separately. This means we have two separate casadi functions implementing $r : R^{nu+nx} -> R^{nr}$ and $\phi : R^{nr} -> R^{nphi}$. The composition between the two functions happens in acados at numerical level, by means of blasfeo matrix-matrix and matrix-vector multiplications.
- implementation of the composite function $\phi \circ r : R^{nu+nx} -> R^{nphi}$ directly. The implementation of the casadi function would return the values of both $\phi \circ r$ and $r$ (and similarly for the jacobians), as exploiting forward AD mode the value and jacobian of $r$ is computed completely for free, as an intermediate step in the computation of the $\phi \circ r$ value and jacobian.

Both approaches are valid in my opinion, only benchmarking can tell which one is better and in which case. So any of them you want to go for, it is good for me.

However, what you implemented is something "in between", which is surely not the most efficient implementation. This is because the value of the function $r$ and jacobian is computed twice, first as part of the composite function $\phi \circ r$ and once more as a stand-alone function.

If you decide to go for the second approach (which is likely closer to your current implementation), I would suggest to rename the single casadi function as `nl_constr_phi_r_fun` or `nl_constr_phi_o_r_fun` or something like that, to clearly show that this implements the composite function.

zanellia:
> The aim here is to implement nonlinear constraints in the form $\phi ( r (x,u) )$, where for the time being we can disregard convexity of $\phi$. There are two ways to properly implement this
> 
> * implementation of the two functions (and relative derivatives) separately. This means we have two separate casadi functions implementing $r : R^{nu+nx} -> R^{nr}$ and $\phi : R^{nr} -> R^{nphi}$. The composition between the two functions happens in acados at numerical level, by means of blasfeo matrix-matrix and matrix-vector multiplications.
> * implementation of the composite function $\phi \circ r : R^{nu+nx} -> R^{nphi}$ directly. The implementation of the casadi function would return the values of both $\phi \circ r$ and $r$ (and similarly for the jacobians), as exploiting forward AD mode the value and jacobian of $r$ is computed completely for free, as an intermediate step in the computation of the $\phi \circ r$ value and jacobian.
> 
> Both approaches are valid in my opinion, only benchmarking can tell which one is better and in which case. So any of them you want to go for, it is good for me.
> 
> However, what you implemented is something "in between", which is surely not the most efficient implementation. This is because the value of the function $r$ and jacobian is computed twice, first as part of the composite function $\phi \circ r$ and once more as a stand-alone function.
> 
> If you decide to go for the second approach (which is likely closer to your current implementation), I would suggest to rename the single casadi function as `nl_constr_phi_r_fun` or `nl_constr_phi_o_r_fun` or something like that, to clearly show that this implements the composite function.

OK, I would go for the second one (btw, notice that the composition in approach 1, would not be numerical: you would need to evaluate $r(x,u)$ and pass it to the CasADi function $\phi(r)$).

giaf:
> > The aim here is to implement nonlinear constraints in the form $\phi ( r (x,u) )$, where for the time being we can disregard convexity of $\phi$. There are two ways to properly implement this
> > 
> > * implementation of the two functions (and relative derivatives) separately. This means we have two separate casadi functions implementing $r : R^{nu+nx} -> R^{nr}$ and $\phi : R^{nr} -> R^{nphi}$. The composition between the two functions happens in acados at numerical level, by means of blasfeo matrix-matrix and matrix-vector multiplications.
> > * implementation of the composite function $\phi \circ r : R^{nu+nx} -> R^{nphi}$ directly. The implementation of the casadi function would return the values of both $\phi \circ r$ and $r$ (and similarly for the jacobians), as exploiting forward AD mode the value and jacobian of $r$ is computed completely for free, as an intermediate step in the computation of the $\phi \circ r$ value and jacobian.
> > 
> > Both approaches are valid in my opinion, only benchmarking can tell which one is better and in which case. So any of them you want to go for, it is good for me.
> > However, what you implemented is something "in between", which is surely not the most efficient implementation. This is because the value of the function $r$ and jacobian is computed twice, first as part of the composite function $\phi \circ r$ and once more as a stand-alone function.
> > If you decide to go for the second approach (which is likely closer to your current implementation), I would suggest to rename the single casadi function as `nl_constr_phi_r_fun` or `nl_constr_phi_o_r_fun` or something like that, to clearly show that this implements the composite function.
> 
> OK, I would go for the second one (btw, notice that the composition in approach 1, would not be numerical: you would need to evaluate $r(x,u)$ and pass it to the CasADi function $\phi(r)$).

True for the function value. For the jacobians, it can be numerical (that's what I had in mind) or using forwards seeds in forward AD mode.

zanellia:
> > > The aim here is to implement nonlinear constraints in the form $\phi ( r (x,u) )$, where for the time being we can disregard convexity of $\phi$. There are two ways to properly implement this
> > > 
> > > * implementation of the two functions (and relative derivatives) separately. This means we have two separate casadi functions implementing $r : R^{nu+nx} -> R^{nr}$ and $\phi : R^{nr} -> R^{nphi}$. The composition between the two functions happens in acados at numerical level, by means of blasfeo matrix-matrix and matrix-vector multiplications.
> > > * implementation of the composite function $\phi \circ r : R^{nu+nx} -> R^{nphi}$ directly. The implementation of the casadi function would return the values of both $\phi \circ r$ and $r$ (and similarly for the jacobians), as exploiting forward AD mode the value and jacobian of $r$ is computed completely for free, as an intermediate step in the computation of the $\phi \circ r$ value and jacobian.
> > > 
> > > Both approaches are valid in my opinion, only benchmarking can tell which one is better and in which case. So any of them you want to go for, it is good for me.
> > > However, what you implemented is something "in between", which is surely not the most efficient implementation. This is because the value of the function $r$ and jacobian is computed twice, first as part of the composite function $\phi \circ r$ and once more as a stand-alone function.
> > > If you decide to go for the second approach (which is likely closer to your current implementation), I would suggest to rename the single casadi function as `nl_constr_phi_r_fun` or `nl_constr_phi_o_r_fun` or something like that, to clearly show that this implements the composite function.
> > 
> > 
> > OK, I would go for the second one (btw, notice that the composition in approach 1, would not be numerical: you would need to evaluate $r(x,u)$ and pass it to the CasADi function $\phi(r)$).
> 
> True for the function value. For the jacobians, it can be numerical (that's what I had in mind) or using forwards seeds in forward AD mode.

Right, although, as you probably meant before, we are talking about the usual question about exploiting sparsity (CasADi) vs high-performance dense LA (BLASFEO). The answer is not trivial and might depend a lot on the application. 

I am implementing the version with a single function, I hope this is fine? :)

giaf:
> I am implementing the version with a single function, I hope this is fine? :)

Sure perfect.

zanellia:
@giaf, thanks for the review, I am done with the last changes ;)

examples/acados_python/test/test_data/generate_c_code_out_LS_FULL_CONDENSING_HPIPM_ERK_SQP_RTI.json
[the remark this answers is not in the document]
FreyJo: Hm, I dont see a change in this file though, it says.
`File renamed without changes.`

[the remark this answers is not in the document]
zanellia: it's this line here, not directly in main_test.py, sorry: https://github.com/acados/acados/pull/506/files#diff-e230ecf1f8731fce690ee3a8eaf0cf5cR258

zanellia:
merge? :)

giaf:
fine with me

PR #516, opened 2019-11-21
small edit to qp module

giaf:
- implement some OSQP options
- add mem_get to pass some e.g. qp timings (TODO remove qp info)
- add HPMPC and OSQP to matlab interface

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -139,6 +139,7 @@ CLEAN_DEPS += qore_clean
 endif
 ifeq ($(ACADOS_WITH_OSQP), 1)
 STATIC_DEPS += osqp_static
+SHARED_DEPS += osqp_shared
 CLEAN_DEPS += osqp_clean
 endif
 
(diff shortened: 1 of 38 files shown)


giaf:
@zanellia it looks like there is the soft constraint test in the python interface failing... Any idea what is going on?

zanellia:
> @zanellia it looks like there is the soft constraint test in the python interface failing... Any idea what is going on?

maybe this? https://github.com/acados/acados/pull/516/files#diff-cc904b69d3f19b60bba0bee83e4c035cR103

giaf:
> > @zanellia it looks like there is the soft constraint test in the python interface failing... Any idea what is going on?
> 
> maybe this? https://github.com/acados/acados/pull/516/files#diff-cc904b69d3f19b60bba0bee83e4c035cR103

Yes it looks like this made the trick :) Thanks for the suggestion!

PR #524, opened 2019-12-08
Template: functional MEX OCP solver, set scaling

FreyJo:
This PR enables the possibility to use the acados templates from Matlab.
There is a Matlab object defined that uses the solver generated by the templates, which is also used when using acados with Simulink.
This makes it easy to interact with the templated Code in the same way, as you can interact with the purely MEX based interface. Especially, since it borrows the `ocp_set` and `ocp_get` functions used therein.
This also enables one to compare the behaviour of templated solver with the purely MEX based one.
In this way, @zanellia  and me found that the scaling of the cost module was not yet done in the templates, which is why some python tests/examples still have to be fixed, see comment below.

I added `examples/acados_matlab_octave/getting_started/test_templated_mex.m` to showcase this new feature.
This file should be transformed into an actual CI test eventually, once this workflow is also supported from Octave.

Note: When switching to this version:
- Matlab/Octave users have to rebuilt the MEX interface, remove the `build` folder in your example folder to do so.
- Python users might get different results compared to previous versions, since the scaling of the cost modules is now set correctly, you might want to adjust the weighting in your OCP formulation.


FreyJo:
The Python test on b30f05b1 will fail, because the results are compared against results that are obtained without setting the scaling in the cost module.
@zanellia, could you tell me
what is the workflow to (re)generate the files in `examples/acados_python/test/test_data`?

zanellia:
> The Python test on [b30f05b](https://github.com/acados/acados/commit/b30f05b1f7a79bba815ab5556e7751786a57c396) will fail, because the results are compared against results that are obtained without setting the scaling in the cost module.
> @zanellia, could you tell me
> what is the workflow to (re)generate the files in `examples/acados_python/test/test_data`?

I guess we should change the code in the examples in order to match the current formulation of the OCP, rather than changing the OCP (in some cases - pmsm and rsm examples, I think - you can just remove the explicit scaling in front of the cost weights).

FreyJo:
Unfortunately, I did not manage to obtain the same results by "unscaling" the weighting matrices, see 
b3a6d9d. Same for the soft constraint example/test.

Instead, I regenerated the test data with the new scaled version of the template.
Thinking again about the files in `test_data`, maybe we should host them in a submodule.
I can also undo 2afd66f, to not have changed 10k+ lines.

As we saw yesterday, this scaling can make a big difference in the behaviour of the controller, so we should really let users know, when merging this into master.
I suggest to merge this PR into `acados/develop_interfaces` for now.
Then, it would be great if you can have a look into the python examples/tests and edit them.

What do you think?

interfaces/acados_template/acados_template/c_templates_tera/acados_mex_create.in.c
```diff
@@ -71,37 +70,30 @@ void mexFunction(int nlhs, mxArray *plhs[], int nrhs, const mxArray *prhs[])
...
-    mxFree( fieldnames[1] );
-    mxFree( fieldnames[2] );
-    mxFree( fieldnames[3] );
-    mxFree( fieldnames[4] );
-    mxFree( fieldnames[5] );
-    mxFree( fieldnames[6] );
+    int fields_ocp = 8;
+    int fields_ext_fun = 11;
+    int nfields = MAX(fields_ext_fun, fields_ocp);
+    char *fieldnames[nfields];
```
giaf: This is a variable size array, which we avoided to use everywhere in acados (core), since this gives issues with e.g. visual studio compiler. Not sure if it is safe for portability to use it here...

interfaces/acados_template/acados_template/c_templates_tera/acados_solver.in.c
```diff
@@ -873,7 +869,17 @@ int acados_create()
 
     external_function_param_casadi_create(&r_e_cost, {{ dims.np }});
     {%- endif %}
-	for (int i = 0; i < N; ++i)
+
+    /* nlp_in */
+    nlp_in = ocp_nlp_in_create(nlp_config, nlp_dims);
+
+    double Ts = Tf/N;
```
giaf: The mex interface also supports non-uniform multiple-shooting grids. It may be worthwhile to add this feature to the template interface already now, instead of leaving a "hidden" mismatch.

giaf: Or if too complicated, at least add a note that this is only valid for uniform multiple-shooting grids, TODO the other, and so on...

FreyJo: sure, done

FreyJo:
alright, I managed to get the same results for the python examples/tests as before.
I undid the changes to the test data.
This can be merged if there are no further remarks

giaf:
Good that the python tests work again now :)
No further remarks from me.

interfaces/acados_template/acados_template/c_templates_tera/acados_mex_create.in.c
```diff
@@ -146,11 +137,105 @@ void mexFunction(int nlhs, mxArray *plhs[], int nrhs, const mxArray *prhs[])
...
-    // mxArray *sens_out_mat  = mxCreateNumericMatrix(1, 1, mxINT64_CLASS, mxREAL);
-    // l_ptr = mxGetData(sens_out_mat);
-    // l_ptr[0] = (long long) sens_out;
-    // mxSetField(plhs[0], 0, "sens_out", sens_out_mat);
+    mxArray *sens_out_mat  = mxCreateNumericMatrix(1, 1, mxINT64_CLASS, mxREAL);
+    l_ptr = mxGetData(sens_out_mat);
+    l_ptr[0] = (long long) 1;
+    mxSetField(plhs[0], 0, "sens_out", sens_out_mat);
+
+    /* store external funciton pointers */
```
zanellia: function :)

zanellia:
Looks good to me too, good job! Merging.

PR #526, opened 2019-12-13
Mex template for Octave, related test on travis, disable codecoverage

FreyJo:
Summary:
- add  [jsonlab](https://github.com/fangq/jsonlab) as a submodule to dumb json files from Octave
- switch to this json en/decoding functionality also in Matlab
- run test_templated_mex work on travis with Octave (more tests to be added)
- disable code coverage: the code coverage suggest a declining coverage because it only takes into account what is covered by the C tests and all tests added in the last year are Python/Octave based.
I think it is better to remove this number than have it suggest a low coverage. Also I think this number is not meaningful anymore.
- add license headers


FreyJo:
ready to be reviewed

FreyJo:
checked with @zanellia

PR #527, opened 2019-12-16
add zero Vx and Vu when using NLS

zanellia:


interfaces/acados_template/acados_template/generate_solver.py
```diff
@@ -69,9 +69,13 @@ def generate_solver(acados_ocp, json_file='acados_ocp_nlp.json'):
         generate_c_code_constraint_e(acados_ocp.con_h_e)
 
     if acados_ocp.cost.cost_type == 'NONLINEAR_LS':
+
+        acados_ocp.cost.Vx = np.zeros((acados_ocp.dims.ny, acados_ocp.dims.nx))
```
FreyJo: why not leave it empty? as initialized in the class constructor?

zanellia: because the Python interface does some consistency checks on the JSON file which (for the moment) do not take into account the type of cost (https://github.com/acados/acados/blob/master/interfaces/acados_template/acados_template/acados_ocp_nlp.py#L1965). They are based on the acados_layout.json file. If you find a nice way to do this systematically, we should implement it ;)

FreyJo: I see, the problem is basically in the constraint type dependent meaning of `ny`.
I think there is no clean way to do this while keeping the overloaded meaning of `ny`..

PR #532, opened 2020-01-07
Develop Sfunction

FreyJo:
Summary of Changes:

- add Simulink example for Matlab

S function for ocp solver:
- add more inputs, bounds to all constraints
- use template logic to only create nonempty input ports

S function for simulation:
- most options hardcoded!
- would be cleaner to export this from `sim` Matlab/Python class

Bug fixes:
- integer division in templated integrator
- Compilation without qpOASES


zanellia:
@FreyJo it would be great to be able to generate an S-Function with a bus with all the necessary signals such that the number of ports does not change if the OCP changes (that would mess up the Simulink model quite a bit). I have looked into this some time ago, but could not come to a solution since buses are usually generated by the Legacy Tool. Maybe we can look into this again?

FreyJo:
@zanellia I totally agree, it would be great to just have one bus as input and output.
Looking into that, I also only only found the legacy tool solution.
And this is also how `ACADO` implements it.
I think, that would be the next step of polishing it.
However, I will probably not do it in this PR.

Another problem is that the Sfunction, can not be tested easily on travis..

zanellia:
Hey @FreyJo and @giaf I think I have figured out how to handle buses in a nice way without having to deal with the Legacy Tool. In this way, the user can add to a Bus Assignment block all the signals that he/she needs to set without blowing up the number of ports unnecessarily (this is by the way how the Simulink block used to be in ACADO). Once this is merged let's have a look ;)

FreyJo:
From my side, this PR is ready to be merged.
@giaf Do you want to add more stuff here?
If not, it would be great to get a review by @zanellia

giaf:
In this PR I implemented time-varying parameters and y_ref in the S-function, but everything else (e.g. bounds) are still time-invariant, so it is a bit of a mixed bag. I would like to have a chat about e.g. on Wednesday, before merging the PR.

But if everyone else is in a hurry to merge it, I guess it can also be merged now and fixed later, up to you.

zanellia:
> In this PR I implemented time-varying parameters and y_ref in the S-function, but everything else (e.g. bounds) are still time-invariant, so it is a bit of a mixed bag. I would like to have a chat about e.g. on Wednesday, before merging the PR.
> 
> But if everyone else is in a hurry to merge it, I guess it can also be merged now and fixed later, up to you.

No hurry on my side. I just wanted to make everybody aware that there might be a valid alternative to adding explicitly a new port for every signal ;)

FreyJo:
I suggest to merge the PR as it is now, even though the wind turbine simulink example in 
190bdd6 is broken.
There are still a few differences in template/MEX interface, which I am trying to resolve one after another.

PR #536, opened 2020-01-16
Free x0

FreyJo:
- Update MEX & Python interface to allow for different state bounds on initial stage
- Make options consistent between Matlab & Python
- Restructured acados_ocp_solver template
- add minimal example python, which is equivalent to the matlab one
- clean out rsm and pmsm examples in Python (thanks to @zanellia )


.travis.yml
```diff
@@ -41,12 +41,12 @@ matrix:
       env:
         - DEPLOY_NAME="osx-python"
 
-    - name: Linux Gcc6 - Octave template
+    - name: Linux Gcc6 - Octave NO_template
```
zanellia: Why no template here?

examples/acados_python/pendulum_example/export_pendulum_ode_model.py
```diff
@@ -48,7 +48,7 @@ def export_ode_model():
     v1      = SX.sym('v1')
     dtheta  = SX.sym('dtheta')
     
-    x = vertcat(x1, v1, theta, dtheta)
```
zanellia: the model here was just wrong, oder?

examples/acados_python/pendulum_example/minimal_example.py
```diff
@@ -0,0 +1,205 @@
+#
+# Copyright 2019 Gianluca Frison, Dimitris Kouzoupis, Robin Verschueren,
```
zanellia: I guess the pendulum example could be removed since we have this, right?

.travis.yml
FreyJo: There is one build with "Octave with templates" and one "Octave wihtout template".
It is just to get an overview over what is broken without looking into the logs.

FreyJo:
> Thanks @FreyJo, I looks pretty good to me! Just one thing: why is testing of templates switched off on macOS?

The macOS travis settings didnt change in this PR.
So, testing templates on MacOS was never done. I agree that we could add it though.

examples/acados_python/pendulum_example/export_pendulum_ode_model.py
FreyJo: yes.

examples/acados_python/pendulum_example/minimal_example.py
FreyJo: There is a bit less happening in the minimal example.
But yes, I think it makes sense to remove the other pendulum example, and rather add some `example_closed_loop` to show the interaction with the solver.
Some example of this kind will be added soon, so I will remove the old example now.

PR #537, opened 2020-01-17
Quickfix BGP

zanellia:
1) fix typo in acados_solver.in.c regarding terminal constraint type
2) fix rsm example
3) clean up pmsm and rsm examples
4) exclude swp files from Tera glob


FreyJo:
Thanks @zanellia 
merging into #536 fixed the issue with the rsm example there.

PR #540, opened 2020-01-23
Polish Python Interface

FreyJo:
- [x] Python: remove constraint class
- [x] Python: remove cost class
- [x] Refactor test with correct pendulum on cart model
- [x] Python naming convention: Classes are CamelCase
- [x] Fix race car example
- [x] Make race car example travis test

Note: This PR contains some breaking changes regarding the python interface, the following should be a draft for corresponding Release Notes:

Dynamics:
-  `acados_dae() ` -> `AcadosSimModel()`

Constraints:
- removed the Python class `acados_constraint`, instead the nonlinear constraints that are defined in CasADi symbolics are defined within `AcadosOcpModel`, this simplifies the formulation, as x, u, z, p just have to be provided once.
- The nonlinear constraint `h` within BGH is specified as the field `con_h_expr` within `AcadosOcpModel`.
- The BGP constraint `phi(r)` aka convex over nonlinear constraint is specified using the fields `con_phi_expr`, `con_r_expr`, `con_r_in_phi` within `AcadosOcpModel`, which used to be `con_phi_expr`, `con_r_expr`, `r` within the `acados_constraint` class.

Cost:
- The nonlinear least squares cost term is now specified by setting the CasADi expressions:
`cost_y_expr`, respectively `cost_y_expr_e` within `AcadosOcpModel`, which used to be `expr` within an `acados_cost` instance called `cost_r`, respectively `cost_r_e`.


FreyJo:
Closing, since I think it is better/easier to directly merge #542

PR #545, opened 2020-02-02
Split RTI into preparation and feedback steps

zanellia:
1. split RTI into preparation and feedback steps: it is not possible to set a `phase` option, which affects what happens when we call `ocp_nlp_sqp_rti`: 
 - `phase = 0` means preparation *and* feedback will be carried out   
- `phase = 1` means that only preparation is performed
- `phase = 2` means that only feedback is performed

2. format ocp_nlp_sqp_rti.c/h
3. restructure Python examples folder. I know that @FreyJo is not a big fan of this, but to me there is way too much stuff going on in the getting_started folder. I think a getting started example needs to be as clean and simple as possible.
4. module names should be lower case: https://www.python.org/dev/peps/pep-0008/#package-and-module-names


interfaces/acados_template/acados_template/AcadosOcpSolver.py
```diff
@@ -414,11 +414,20 @@ def __init__(self, acados_ocp, json_file='acados_ocp_nlp.json'):
...
-    def solve(self):
+    def solve(self, phase=0):
         """
         solve the ocp with current input
+        :param phase: 0 = preparation + feedback, 1 = preparation only,
+         2 = feedback only (if SQP_RTI is used, otherwise only 0 (default) is allowed)
         """
-        status = self.shared_lib.acados_solve()
+        if isinstance(phase, int) == False or phase < 0 or phase > 2: 
+            raise Exception('AcadosOcpSolver.solve():')
```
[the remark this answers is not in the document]
zanellia: indeed :) fixed here: f5eb094

examples/acados_python/getting_started/mhe/minimal_example_mhe.py
```diff
@@ -31,6 +31,10 @@
 # POSSIBILITY OF SUCH DAMAGE.;
 #
 
+import sys
+sys.path.insert(0, '../common')
+sys.path.insert(0, '../ocp')
```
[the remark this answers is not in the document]
zanellia: good idea. Addressed here (5a1ffca, 865103e)

examples/acados_python/getting_started/ocp/minimal_example_closed_loop.py
```diff
@@ -31,6 +31,9 @@
 # POSSIBILITY OF SUCH DAMAGE.;
 #
 
+import sys
```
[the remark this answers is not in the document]
zanellia: yes, sounds good! (8526d31)

acados/ocp_nlp/ocp_nlp_sqp.h
```diff
@@ -64,11 +64,12 @@ typedef struct
     double tol_stat;     // exit tolerance on stationarity condition
     double tol_eq;       // exit tolerance on equality constraints
     double tol_ineq;     // exit tolerance on inequality constraints
-    double tol_comp;     // exit tolerance on complemetarity condition
+    double tol_comp;     // exit tolerance on complementarity condition
     int max_iter;
     int ext_qp_res;      // compute external QP residuals (i.e. at SQP level) at each SQP iteration (for debugging)
     int qp_warm_start;   // qp_warm_start in all but the first sqp iterations
     bool warm_start_first_qp; // to set qp_warm_start in first iteration
+    int phase;           // only phase 0 at the moment
```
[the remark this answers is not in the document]
zanellia: Yes, for example, I need to revive zero-order NMPC with lifted integrators pretty soon (next PR) and it will use something similar to this field (or even the same). Parallelization of mulit-level iterations is yet another topic, but I haven't thought about it because I don't need it at the moment. Do you have some ideas about it?

acados/ocp_nlp/ocp_nlp_constraints_bgh.c
```diff
@@ -1293,6 +1293,36 @@ void ocp_nlp_constraints_bgh_compute_fun(void *config_, void *dims_, void *model
...
+    int nu = dims->nu;
+    int nb = dims->nb;
+    int ng = dims->ng;
+    int nh = dims->nh;
+
+    // box
+    blasfeo_dvecex_sp(nb, 1.0, model->idxb, memory->ux, 0, &work->tmp_ni, 0);
+
+    blasfeo_daxpy(nb, -1.0, &work->tmp_ni, 0, &model->d, 0, &memory->fun, 0);
+    blasfeo_daxpy(nb, -1.0, &model->d, nb+ng+nh, &work->tmp_ni, 0, &memory->fun, nb+ng+nh);
```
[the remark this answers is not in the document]
zanellia: Indeed that's the point :) we could make it more general, but the most efficient way to do it, and the one that complies with Moritz's original RTI, is to exploit what he calls 'initial value embedding'. To my best knowledge, in most of his literature, this requires the assumption that $x_0$ enters the NLP linearly ($C * x_0 = \bar{x}_0$). A slightly more restrictive formulation, the one we use, assumes that $C= I$, hence, we only need to update the box constraints. 

On a side note, initial value embedding or, more generally, parameter embedding, can be done even for nonlinear constraints and/or costs, but we would need to compute derivatives of the underlying functions with respect to parameters too and compute the linearization accordingly.

acados/ocp_nlp/ocp_nlp_sqp_rti.c
```diff
@@ -441,22 +522,14 @@ int ocp_nlp_sqp_rti(void *config_, void *dims_, void *nlp_in_, void *nlp_out_,
     ocp_nlp_sqp_rti_cast_workspace(config, dims, opts, mem, work);
     ocp_nlp_workspace *nlp_work = work->nlp_work;
 
-    // zero timers
-    double total_time = 0.0;
 	double tmp_time;
-    mem->time_qp_sol = 0.0;
```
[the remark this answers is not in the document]
zanellia: ouch, thanks! 52ecbbe

acados/ocp_nlp/ocp_nlp_sqp_rti.c
```diff
@@ -546,51 +688,93 @@ int ocp_nlp_sqp_rti(void *config_, void *dims_, void *nlp_in_, void *nlp_out_,
...
+    ocp_nlp_sqp_rti_workspace *work = work_;
+    ocp_nlp_sqp_rti_cast_workspace(config, dims, opts, mem, work);
+    ocp_nlp_workspace *nlp_work = work->nlp_work;
+
+    int qp_iter = 0;
+    int qp_status = 0;
+	double tmp_time;
+
+    // linearizate NLP and update QP matrices
+    acados_tic(&timer1);
```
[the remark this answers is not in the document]
zanellia: Actually, there is no linearization at all in the feedback phase. I have removed the comment. Moreover, since initial value embedding + vectors update should be very fast, I would not add a separate timing entry for it. It will only count towards the total_time computation. (b01ff2a)

zanellia:
Thanks @FreyJo for the review! @giaf could you have a quick look?

acados/ocp_nlp/ocp_nlp_sqp.h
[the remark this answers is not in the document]
FreyJo: > I need to revive zero-order NMPC with lifted integrators pretty soon (next PR) and it will use something similar to this field (or even the same).

Sounds good!
I think as soon as you introduce another meaning for `phase` it would be really good to use `enum`

> Parallelization of mulit-level iterations is yet another topic, but I haven't thought about it because I don't need it at the moment. Do you have some ideas about it?

No, it remains interesting though :grinning:

acados/ocp_nlp/ocp_nlp_constraints_bgh.c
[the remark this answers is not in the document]
FreyJo: Thanks for that clarification!

acados/ocp_nlp/ocp_nlp_sqp.h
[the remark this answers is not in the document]
giaf: +1 for enum!

acados/ocp_nlp/ocp_nlp_sqp_rti.c
```diff
@@ -546,51 +688,92 @@ int ocp_nlp_sqp_rti(void *config_, void *dims_, void *nlp_in_, void *nlp_out_,
...
+    ocp_nlp_sqp_rti_cast_workspace(config, dims, opts, mem, work);
+    ocp_nlp_workspace *nlp_work = work->nlp_work;
+
+    int qp_iter = 0;
+    int qp_status = 0;
+    double tmp_time;
+    mem->time_qp_sol = 0.0;
+    mem->time_qp_solver_call = 0.0;
+
+    // embed initial value (this actually updates all bounds...)
```
giaf: it only updates the bounds at stage zero, so the comment is misleading

acados/ocp_nlp/ocp_nlp_sqp_rti.c
```diff
@@ -693,10 +874,16 @@ int ocp_nlp_sqp_rti_precompute(void *config_, void *dims_, void *nlp_in_, void *
...
+            dims->dynamics[ii], nlp_in->dynamics[ii], "T", nlp_in->Ts+ii);
+
         // dynamics precompute
-        status = config->dynamics[ii]->precompute(config->dynamics[ii], dims->dynamics[ii], nlp_in->dynamics[ii],
-                                                  opts->nlp_opts->dynamics[ii], nlp_mem->dynamics[ii], nlp_work->dynamics[ii]);
+        status = config->dynamics[ii]->precompute(config->dynamics[ii],
+            dims->dynamics[ii], nlp_in->dynamics[ii],
+            opts->nlp_opts->dynamics[ii],
+            nlp_mem->dynamics[ii],
+            nlp_work->dynamics[ii]);
```
giaf: ok trying to void too long lines, but this is a bit excessive

acados/ocp_nlp/ocp_nlp_sqp_rti.h
```diff
@@ -57,36 +57,41 @@ extern "C" {
...
 {
 	ocp_nlp_opts *nlp_opts;
     int compute_dual_sol;
-    int ext_qp_res;      // compute external QP residuals (i.e. at SQP level) at each SQP iteration (for debugging)
-    int qp_warm_start;   // NOTE: this is not actually setting the warm_start! Just for compatibility with sqp.
+    int ext_qp_res;           // compute external QP residuals (i.e. at SQP level) at each SQP iteration (for debugging)
+    int qp_warm_start;        // NOTE: this is not actually setting the warm_start! Just for compatibility with sqp.
     bool warm_start_first_qp; // to set qp_warm_start in first iteration
-    int print_level;     // possible values 0, 1 
+    int phase;                // phase of RTI. Possible values 0, 1 and 2.
```
giaf: I would specify a bit better in the comment what phase of RTI is (i.e. feedback/preparation).

In general the name `phase` is confusing as it may refer to multi-phase, which is something we want to implement anyway at some point.
What about renaming `phase` into `rti_phase`?

examples/acados_python/getting_started/minimal_example_closed_loop.py
```diff
@@ -112,11 +115,13 @@
 
 # closed loop
 for i in range(N):
+
     # solve ocp
     acados_ocp_solver.set(0, "lbx", xcurrent)
     acados_ocp_solver.set(0, "ubx", xcurrent)
 
-    status = acados_ocp_solver.solve()
+    status = acados_ocp_solver.solve(phase=0)
```
giaf: Honestly I don't like passing the phase to the solve method.
`phase` is just one of the many options affecting the solver, and as such it should be set in the options (before creating the solver), and in the solver itself after its creation
`acados_ocp_solver.set("rti_phase", 0)`
If not handled in this way, the argument list for `solve()` would indefinitely increase as we come up with more stuff.

zanellia: the options can be lumped into a `opts` object - that way the behavior of the call to `solve` would not depend on this additional state of the solver, but rather on the arguments explicitly passed by the user. What do you think?

acados/ocp_nlp/ocp_nlp_sqp_rti.c
zanellia: well, that's the best I could do to stay within 80 lines... which is exactly how many lines fit on a vim vertical split on a 13 inch display with a reasonable resolution :p

acados/ocp_nlp/ocp_nlp_sqp_rti.h
zanellia: I agree (addressed 064a5ff)

acados/ocp_nlp/ocp_nlp_sqp_rti.c
zanellia: addressed 064a5ff

acados/ocp_nlp/ocp_nlp_sqp_rti.c
FreyJo: Was also thinking this.
I remember the lint rule was 100 characters per line, which I still try to stick to.

examples/acados_python/getting_started/minimal_example_closed_loop.py
FreyJo: I think both points are valid:
Adding inputs to the `solve` call will add complexity in another place.

Options that can be changed after the creation of the solver are now be gathered here:
https://github.com/acados/acados/blob/35288465716ec7abc538909d77e98dd1e0b70374/interfaces/acados_template/acados_template/AcadosOcpSolver.py#L641

Right now, this also makes
> the behavior of the call to solve would not depend on this additional state of the solver


But having only one API to change these options is definitely desirable IMO.

acados/ocp_nlp/ocp_nlp_sqp_rti.c
zanellia: it was 80 recommended, 100 max admitted, I think. But again, it's not a style exercise. They just don't fit on a 13 inch display in horizontal split. Try yourself!

zanellia:
Thanks @giaf and @FreyJo! Can I merge this? I will implement enums in the next PR together with zero-order NMPC, if that's fine.

zanellia:
alright, I have removed the `rti_phase` argument from calls to `solve`-like functions (1ac4e3d). Merge? :)

PR #547, opened 2020-02-04
add print_level option

zanellia:
1. adding print_level option
2. add opts_set setter to Python interface
3. changing to shorter setter name cons_set in Python interface

### DIFF
--- a/interfaces/acados_template/acados_template/AcadosOcpSolver.py
+++ b/interfaces/acados_template/acados_template/AcadosOcpSolver.py
@@ -539,7 +539,7 @@ class AcadosOcpSolver:
         return
 
 
-    def constraints_set(self, stage_, field_, value_):
+    def cons_set(self, stage_, field_, value_):
         """
         set numerical data in the constraint module of the solver:
         Parameters:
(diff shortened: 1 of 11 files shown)


interfaces/acados_template/acados_template/AcadosOcpSolver.py
```diff
@@ -539,7 +539,7 @@ def cost_set(self, stage_, field_, value_):
         return
 
 
-    def constraints_set(self, stage_, field_, value_):
+    def cons_set(self, stage_, field_, value_):
```
FreyJo: I think `constraints_set` is more clear and not too long.
If you really want to abbreviate it, I would use `constr` as in the Mex interface.
`cons` could be misinterpreted as constant.

zanellia: Fair enough. I like it better too, now that I think about it. Then let's make the convention that all setters must be <full_module_name>_set.

zanellia: @giaf?

zanellia:
let's merge? :)

PR #548, opened 2020-02-04
fixing bugs in acados_solver.in.c - C_e not set, wrong D and C dims

zanellia:
The first one was introduced recently, how did it not break the tests?

### DIFF
--- a/interfaces/acados_template/acados_template/c_templates_tera/acados_solver.in.c
+++ b/interfaces/acados_template/acados_template/c_templates_tera/acados_solver.in.c
@@ -894,13 +894,13 @@ int acados_create()
     double ug[NG];
 
     {% for j in range(end=dims.ng) %}
-        {%- for k in range(end=dims.nx) %}
+        {%- for k in range(end=dims.nu) %}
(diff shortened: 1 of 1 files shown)


zanellia:
> The first one was introduced recently, how did it not break the tests?

I see, the solution of the PMSM example is not checked of course.

FreyJo:
> The first one was introduced recently, how did it not break the tests?

Right, the C_e probably got lost, while restructuring!
I just checked the last release, there the dimensions of D and C are also wrong.

FreyJo:
Yes, adding a test for the general linear constraint would be great!

> > The first one was introduced recently, how did it not break the tests?
> 
> I see, the solution of the PMSM example is not checked of course.

Adding a test for the general linear constraints would be great!

zanellia:
> > The first one was introduced recently, how did it not break the tests?
> 
> Right, the C_e probably got lost, while restructuring!
> I just checked the last release, there the dimensions of D and C are also wrong.

I know, that's why I said the first one :) The second one was never detected because the only example that makes use of the polytopic constraints at t=T had n_x = n_u.

FreyJo:
Ah good to know :)
I was not sure about "first", cause I looked mainly at the diff.

PR #550, opened 2020-02-04
fix bug in S-Function

zanellia:
Global data should only be defined ones either in the main and in the S-Function OR in the solver.


zanellia:
> Global data should only be defined ones either in the main and in the S-Function OR in the solver.

This somehow does not causes linking problems in MATLAB, but it does on dSPACE.

PR #555, opened 2020-02-19
Exact hess template

FreyJo:


giaf:
The test failing may be due to the `-std=c99` flag, which may be set by us or by default in cmake. I will try to replace `asm` with `__asm__`, and check if this fixes the issue for all tests. Otherwise, we should investigate the `std` flag.

giaf:
@FreyJo please upload blasfeo and try again
https://github.com/giaf/blasfeo/commit/2569492775bdbe2fe33822d6d690728b4146e90d

giaf:
pull blasfeo and try again :p
https://github.com/giaf/blasfeo/commit/e33e163f1391ae03fe936b9f7a90a9fc62e8c825

FreyJo:
@giaf great that worked :ok_hand:

docs/interfaces/index.md
```diff
@@ -49,10 +49,8 @@ Note: This part of the MATLAB/Octave interface does not yet support all features
 ## Python (templates)
 
 `acados_template` is a Python package that can be used to specify optimal control problems from Python and to generate self-contained C code that uses the acados solvers to solve them.
-In comparison with the MATLAB interface for rapid prototyping (see above), it supports less features, but it allows the user to generate a self-contained C library
-that can be easily deployed on an embedded system.
-
-The framework is based on templated C files which are rendered from Python using the templating engine `Tera`.
+The packaged is based on templated code (C files, Header files and Makefiles), which are rendered from Python using the templating engine `Tera`.
```
zanellia: `package` and which `is` rendered :)

docs/interfaces/index.md
```diff
@@ -69,17 +67,25 @@ make install -j4
 ```
 pip3 install <acados_root>/interfaces/acados_template
 ```
+Note: Use `pip` instead of `pip3` if you are installing in a virtual Python environment.
```
zanellia: using a virtual env is not a necessary not sufficient condition for switching between pip and pip3 as far as I know

examples/acados_python/getting_started/main_test.py
```diff
@@ -36,34 +36,78 @@
 from random import sample
 
 TEST_SAMPLE = True # only test random sample instaed of all possible combinations
```
zanellia: instead

examples/acados_python/getting_started/main_test.py
```diff
@@ -36,34 +36,78 @@
...
-SAMPLE_SIZE = 20
+SAMPLE_SIZE = 30
```
zanellia: Actually, I am not sure that random testing is a good idea. It would make debugging more difficult in my opinion.

FreyJo: CI tests are supposed to run relatively fast.
The combinations of settings are otherwise 486 right now.
It is some kind of trade-off.

Running the 60 test cases takes ~ 1 minute on my machine and I think it should not take much longer.

I agree there might be cases that are not detected right away, but a skipping a module/ setting entirely is very unlikely.

docs/interfaces/index.md
FreyJo: Using a virtual env should be sufficient to use `pip`.
If you run `which pip`, it points to the one in the virtual env.
Or am I missing something? :thinking:

examples/acados_python/getting_started/main_test.py
FreyJo: Do you have another idea of doing this tradeoff between testing most settings and reasonable runtime?

zanellia: random with fixed seed? then you would get always the same combination at least

docs/interfaces/index.md
zanellia: Depending on the installation of pip you could have different aliasing. Indeed on Ubuntu when you use a virtualenv pip -> pip3, but I would not be surprised if this was not the case with other environment manager/OS combinations (please prove me wrong :))

examples/acados_python/getting_started/main_test.py
FreyJo: Then the cases which are not in the "seed" are never tested.
I am not convinced that this is better.

zanellia: Well you said that `skipping a module/ setting entirely is very unlikely`, then we can as well pick an arbitrary set of tests and use it all the time. Otherwise/in addition we can add one test to the Travis matrix that runs all the tests which we turn on only before merging - dunno. I just wouldn't like to be the guy who needs to find a bug with a randomized unit test :p

docs/interfaces/index.md
FreyJo: > with other environment manager/OS combinations

Okay, I wont look into these combinations :sweat_smile: 
I made the note more generic:

Note: If you are working with a virtual Python environment, use the `pip` corresponding to this Python environment instead of `pip3`.

examples/acados_python/getting_started/main_test.py
FreyJo: > I just wouldn't like to be the guy who needs to find a bug with a randomized unit test :p

Actually, one can easily turn of the `TEST_SAMPLE` flag or set a fixed combination, which you want to debug when going through the history.

I understand the dislike of the unit test being random though and implemented a fixed seed test in b604241

giaf: I'm also against random testing, since it could give both "false positive" (i.e. I push something, the test fails, but the issue is not in my push) and "false negative" (i.e. I push something broken but this is not detected). Debugging seems a nightmare.

Many of the fields are orthogonal to each other, so not all combinations of values have to be tested. This greatly reduces the amount of combinations necessary to cover all features.

examples/acados_python/getting_started/test_ocp_setting.py
```diff
@@ -210,9 +238,11 @@
 ocp.solver_options.nlp_solver_tol_ineq = TEST_TOL
 ocp.solver_options.nlp_solver_tol_comp = TEST_TOL
 
-ocp.solver_options.qp_solver_cond_N = 10
-ocp.solver_options.nlp_solver_max_iter = 80
+ocp.solver_options.qp_solver_cond_N = int(N/2)
```
giaf: maybe add a safeguards that this should be at least 1

giaf: I mean, if N=1, `int(N/2)` is 0 or 1?

examples/acados_python/getting_started/test_ocp_setting.py
```diff
@@ -258,9 +315,10 @@
     simX_error = np.linalg.norm(test_data['simX'] - simX)
     simU_error = np.linalg.norm(test_data['simU'] - simU)
 
-    if simX_error > TEST_TOL or simU_error > TEST_TOL:
+    CHECK_TOL = 50 * TEST_TOL
```
giaf: what is this for?

examples/acados_python/getting_started/utils.py
```diff
@@ -36,7 +36,7 @@
 import matplotlib.pyplot as plt
 import numpy as np
 
-def plot_pendulum(h, u_max, U, X_true, X_est=None, Y_measured=None, latexify=True):
+def plot_pendulum(h, u_max, U, X_true, X_est=None, Y_measured=None, latexify=False):
```
giaf: what is `latexify`?

interfaces/acados_c/ocp_nlp_interface.c
```diff
@@ -533,6 +533,10 @@ int ocp_nlp_dims_get_from_attr(ocp_nlp_config *config, ocp_nlp_dims *dims, ocp_n
     {
         return dims->nx[stage];
     }
+    else if (!strcmp(field, "pi"))
+    {
+        return dims->nx[stage];
```
giaf: the size of `pi[stage]` is `nx[stage+1]`

interfaces/acados_template/acados_template/c_templates_tera/acados_solver.in.h
```diff
@@ -95,10 +95,10 @@ extern external_function_param_casadi h_e_constraint;
 {%- endif %}
 
 {% if cost.cost_type == "NONLINEAR_LS" %}
-extern external_function_param_casadi * r_cost;
+extern external_function_param_casadi * cost_y_fun;
```
giaf: +1 for homogenizing the notation :)

examples/acados_python/getting_started/utils.py
FreyJo: It makes the plot function use fancy latex fonts

interfaces/acados_c/ocp_nlp_interface.c
FreyJo: good catch, thanks!

examples/acados_python/getting_started/test_ocp_setting.py
FreyJo: TEST_TOL is the tolerance set for the ocp_solver.
Here, we compare x and u trajectories wrt reference solutions.
This error can be bit larger depending on the cost function scaling.
Anyway, I had to increase this to make Travis pass.

FreyJo:
I am glad writing additional tests already paid off.
Namely, the regularization has to be called directly before the QP solve, at least for `CONVEXIFY`.
@zanellia and I would have expected it to be independent of the right hand sides, which seems to be not true.
Maybe @giaf can give some insight..

examples/acados_python/getting_started/test_ocp_setting.py
FreyJo: this is just the test.
It is checked in the corresponding setter:
https://github.com/FreyJo/acados/blob/exact_hess_template/interfaces/acados_template/acados_template/acados_ocp.py#L1810

giaf:
The `convexify` regularization method aims at giving the exact solution in the case the reduced Hessian is positive definite. It is based on a recursion analogue to backward Riccati recursion, and dynamics equation enters in the computation of the cost transfer between stages.
In Robin's thesis, this is done only for the case of `b=0`, and consequently `r` and `q` are unchanged by the convexification algorithm. However, in the implementation I correctly address the most general case of `b!=0`, which results in corrections also for the terms `r` and `q`.
So, surely convexification can not happen before `r` and `q` are computed.
The current implementation does not consider currently active inequality constraints, so convexification should safely be placed also before the initial value embedding.

FreyJo:
@giaf thanks for the insight!
>  The current implementation does not consider currently active inequality constraints, so convexification should safely be placed also before the initial value embedding.

PR #559, opened 2020-03-14
Various fixes

FreyJo:
This PR contains:
- C: 
   - Went through Hessian propagation and cost modules
   - Fix bug in NLS for exact Hessian https://github.com/acados/acados/pull/559/commits/12da6f3fd71051b4cf118813f5f03a6640787602
- Python:
   - fix getters of sensitivities for sim solver
    - add GNSF support for sim template
    - add more timing getters
- Python test: equivalent scaling with external cost
- Attempted to run Python on MacOS, added some useful stuff, but didnt finalize yet.

### DIFF
--- a/acados/ocp_nlp/ocp_nlp_constraints_bgp.c
+++ b/acados/ocp_nlp/ocp_nlp_constraints_bgp.c
@@ -1049,7 +1049,6 @@ void ocp_nlp_constraints_bgp_update_qp_matrices(void *config_, void *dims_, void
 
     int nv = nx + nu;
 
-    // XXX large enough ?
     ext_fun_arg_t ext_fun_type_in[3];
     void *ext_fun_in[3];
     ext_fun_arg_t ext_fun_type_out[5];
--- a/acados/ocp_nlp/ocp_nlp_cost_external.c
+++ b/acados/ocp_nlp/ocp_nlp_cost_external.c
@@ -629,17 +629,16 @@ void ocp_nlp_cost_external_update_qp_matrices(void *config_, void *dims_, void *
     // blasfeo_print_tran_dvec(2*ns, memory->Z, 0);
     // blasfeo_print_tran_dvec(nu+nx+2*ns, &memory->grad, 0);
 
-	// TODO compute fun
-
     return;
 }
 
 
 
-void ocp_nlp_cost_external_compute_fun(void *config_, void *dims_, void *model_, void *opts_, void *memory_, void *work_)
+void ocp_nlp_cost_external_compute_fun(void *config_, void *dims_, void *model_,
+                                       void *opts_, void *memory_, void *work_)
 {
-//	printf("\nerror: ocp_external_cost_nls_compute_fun: not implemented yet\n");
-//	exit(1);
+//    printf("\nerror: ocp_external_cost_nls_compute_fun: not implemented yet\n");
+//    exit(1);
 
     ocp_nlp_cost_external_dims *dims = dims_;
     ocp_nlp_cost_external_model *model = model_;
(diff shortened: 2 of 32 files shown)


acados/ocp_nlp/ocp_nlp_constraints_bgp.c
```diff
@@ -1049,7 +1049,6 @@ void ocp_nlp_constraints_bgp_update_qp_matrices(void *config_, void *dims_, void
 
     int nv = nx + nu;
 
-    // XXX large enough ?
```
giaf: All these comments on `XXX large enough ?` are intended to remind that these quantities have to be checked and updated in case the definition of the external function changes, since this is something very easy to overlook (and it indeed happened in the past).

You can replace them with a nicer statement if you like, but I would leave some note about this.

acados/ocp_nlp/ocp_nlp_cost_nls.c
```diff
@@ -784,7 +784,7 @@ void ocp_nlp_cost_nls_update_qp_matrices(void *config_, void *dims_, void *model
 
         // RSQrq = RSQrq + tmp_nv_ny * tmp_nv_ny^T
         blasfeo_dsyrk_ln(nu+nx, ny, model->scaling, &work->tmp_nv_ny, 0, 0, &work->tmp_nv_ny, 0, 0,
-                         model->scaling, &work->tmp_nv_nv, 0, 0, memory->RSQrq, 0, 0);
+                         1.0, memory->RSQrq, 0, 0, memory->RSQrq, 0, 0);
```
giaf: It was correct as it was before, since the gauss-newton term was added to (the scaled version of) the term `hess(r)*r`, which was stored in `tmp_nv_nv`. After your "fix", this term is disregarded and the code computes the same as in the `opts->gauss_newton_hess` case.

giaf: If anything, the comment just above is to be fixed.

acados/ocp_nlp/ocp_nlp_constraints_bgp.c
FreyJo: I think it is quite obvious that these would have to be changed when the external function expected by the module changes.
I checked that all these sizes with respect to how it is used..

acados/ocp_nlp/ocp_nlp_cost_nls.c
FreyJo: Thanks a lot for this comment!
I will undo this change and fix the comment above.

Moreover, I will have a look at the other cost modules, I made this change to make the NLS cost equivalent to the other two on the test problem.
Which makes me think now, that the other two cost modules do have the bug that I introduced here..

acados/ocp_nlp/ocp_nlp_cost_external.c
```diff
@@ -612,34 +612,38 @@ void ocp_nlp_cost_external_update_qp_matrices(void *config_, void *dims_, void *
...
+    blasfeo_dveccpsc(2*ns, 2.0, &model->z, 0, &work->tmp_2ns, 0);
+    blasfeo_dvecmulacc(2*ns, &model->Z, 0, memory->ux, nu+nx, &work->tmp_2ns, 0);
+    memory->fun += 0.5 * blasfeo_ddot(2*ns, &work->tmp_2ns, 0, memory->ux, nu+nx);
+
     // scale
-    if(model->scaling!=1.0)
+    if (model->scaling!=1.0)
     {
         blasfeo_dvecsc(nu+nx+2*ns, model->scaling, &memory->grad, 0);
+        memory->fun *= model->scaling;
```
giaf: good point

acados/ocp_nlp/ocp_nlp_cost_external.c
```diff
@@ -680,16 +683,13 @@ void ocp_nlp_cost_external_compute_fun(void *config_, void *dims_, void *model_,
     ext_fun_out[0] = &memory->fun;  // function: scalar
 
     // evaluate external function
-    model->ext_cost_fun->evaluate(model->ext_cost_fun, ext_fun_type_in, ext_fun_in, ext_fun_type_out, ext_fun_out);
-
-    // slacks
-    blasfeo_dveccp(2*ns, &model->z, 0, &memory->grad, nu+nx);
-    blasfeo_dvecmulacc(2*ns, &model->Z, 0, memory->ux, nu+nx, &memory->grad, nu+nx);
```
giaf: I guess this was some copy-and-paste

acados/ocp_nlp/ocp_nlp_cost_ls.c
```diff
@@ -851,27 +863,42 @@ void ocp_nlp_cost_ls_compute_fun(void *config_, void *dims_, void *model_, void
...
+
+        blasfeo_dgemv_n(ny, nz, +1.0, &model->Vz,
+                0, 0, &work->tmp_nz, 0, 1.0, &work->y_ref_tilde, 0, &work->y_ref_tilde, 0);
+
+        // res = \tilde{V}_x * x + \tilde{V}_u * u - \tilde{y}_ref
+        blasfeo_dgemv_t(nu + nx, ny, 1.0, &work->Cyt_tilde, 0, 0, memory->ux,
+                0, -1.0, &work->y_ref_tilde, 0, &memory->res, 0);
+
+        // tmp_ny = W * res
+        blasfeo_dsymv_l(ny, ny, 1.0, &model->W, 0, 0, &memory->res,
```
giaf: you can use the (cached) cholesky factor of W and `dtrmv` as in the `else` code, it's a bit faster

acados/ocp_nlp/ocp_nlp_cost_ls.c
```diff
@@ -851,27 +863,42 @@ void ocp_nlp_cost_ls_compute_fun(void *config_, void *dims_, void *model_, void
...
+                0, 0.0, &work->tmp_ny, 0, &work->tmp_ny, 0);
+
+        memory->fun = 0.5 * blasfeo_ddot(ny, &work->tmp_ny, 0, &memory->res, 0);
+        printf("\nerror: ocp_nlp_cost_ls_compute_fun: not implemented yet for nz>0\n");
+        exit(1);
```
giaf: I guess this should go now

Issue #562, opened 2020-03-22
Duplicate custom target 'uninstall' when installing acados with OSQP


andschi:
Hi, 

I just wanted to install acados with CMake on my Ubuntu 18.04 machine. When enabling the installation of OSQP, cmake throws the error that there are duplicate targets named 'uninstall'. See the CMake-Ouput below. 
The add_custom_target(uninstall ...) appears in both CMakeLists. 

Best regards, 
Andi

cmake .. -DACADOS_WITH_OSQP=ON    
-- The C compiler identification is GNU 8.3.0
-- The CXX compiler identification is GNU 8.3.0
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Build type is Release
-- The ASM compiler identification is GNU
-- Found assembler: /usr/bin/cc
-- Using linear algebra: HIGH_PERFORMANCE
-- Using external BLAS: 0
-- Testing target X64_INTEL_HASWELL: compilation [success]
-- Testing target X64_INTEL_HASWELL: run [success]
-- Detected target X64_INTEL_HASWELL
-- Using BLASFEO path: 
-- We are on a Linux system
-- Embedded is OFF
-- Printing is ON
-- Profiling is OFF
-- User interrupt is OFF
-- Floats are OFF
-- Long integers (64bit) are ON
-- Code coverage is OFF
-- MKL Pardiso: ON
-- Floats are OFF
-- Long integers (64bit) are ON
CMake Error at CMakeLists.txt:202 (add_custom_target):
  add_custom_target cannot create target "uninstall" because another target
  with the same name already exists.  The existing target is a custom target
  created in source directory
  "/home/USER/Downloads/acados/external/osqp/lin_sys/direct/qdldl/qdldl_sources".
  See documentation for policy CMP0002 for more details.

-- Target: BLASFEO is X64_AUTOMATIC, HPIPM is X64_AUTOMATIC
-- Linear algebra: HIGH_PERFORMANCE
-- Matlab MEX (OFF)
[...]

PR #563, opened 2020-03-22
Fix the cmake installation with OSQP

FreyJo:
This should fix the cmake installation with OSQP
https://github.com/acados/acados/issues/562


FreyJo:
Just tested that `make uninstall_acados` does would it should :heavy_check_mark: 
Merging

PR #566, opened 2020-03-24
Use PBR to simplify package build of Python acados_template

MaciejKucia:
PBR will auto-generate version. This way, creation of acados_template can be automated.

The new package version/name vs the old:
```
acados_template-0.0.1.dev6853-py3-none-any.whl
acados_template-0.1-py3-none-any.whl
```

The version is generated based on annotated git tag. For example:
```
$ git tag -a 1.2.3 -m "test"
$ ls dist/
acados_template-1.2.3-py2.py3-none-any.whl
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -94,6 +94,7 @@ install:
   - source "${SCRIPT_DIR}/dispatch.sh"
 
 script:
+  - git fetch --unshallow || true
   - SECTION='script'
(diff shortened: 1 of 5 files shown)


MaciejKucia:
I have proposed another, simpler solution that works with `pip install .` in https://github.com/acados/acados/pull/567

PR #567, opened 2020-03-24
Use setuptools-scm to simplify package building of Python acados_template

MaciejKucia:
When using tools like `python setup.py bdist_wheel` a python package can be produced and shared with other developers.
Because of frequent changes, it is not convenient to update version in setup.py on each change. The scm module will update change accordingly to git version (using annotated tags).

### DIFF
--- a/interfaces/acados_template/setup.py
+++ b/interfaces/acados_template/setup.py
@@ -48,6 +48,12 @@ setup(name='acados_template',
    license='BSD',
    packages = find_packages(),
    include_package_data = True,
+   setup_requires=['setuptools_scm'],
+   use_scm_version={
+     "fallback_version": "0.1-local",
+     "root": "../..",
+     "relative_to": __file__
+   },
    install_requires=[
       'numpy',
       'scipy',


MaciejKucia:
@FreyJo could you consider this change for merge?

FreyJo:
Hey @MaciejKucia,
first of all thanks for the effort!
I looked a bit into python packaging https://packaging.python.org/tutorials/packaging-projects/
And also generated one with `bdist_wheel`.

With the current workflow and acados not being super stable yet, I think we will not distribute the Python interface as a package soon.

I really like to have it installed symbolically with `pip -e`, so I don't have to worry when switching versions.

Just for my understanding: What is the use case where the packaging simplifies the workflow significantly?

MaciejKucia:
As discussed in another topic, I do not expect this to be released on pypi.

Each project I use is under ci/cd pipeline. Each change is producing a new build which has artefacts. It's a part of larger, universal process.
Having automatic versioning will make this much easier and I believe it can help community as well.

The "main" part of acados can be packed as a [Conan](https://conan.io/) package which I can contribute too. I think it does not make sense at this stage of the project though. Since Conan does not package python packages, I need to have two, correlated packages.

Fell free to decline this change, all I hope for is some feedback :smiley:

FreyJo:
I see the possible advantages in CI/CD heavy workflows.
For me, it can be merged.
Maybe @zanellia has still some remark?

zanellia:
No remarks on my side. I have no experience with `setuptools-scm`, I guess we will figure out how it can be useful later on :)

PR #568, opened 2020-03-25
cmake: fix uninstall target properly with if guard, see 58826cae

FreyJo:


imciner2:
You can also move the uninstall target creation before the creation of any of the submodule targets so that acados' gets created first - then OSQP and QDLDL won't creat their own (since they are guarded by the `if( NOT TARGET uninstall)` check. That will then include your directory deletion in the uninstall process.

FreyJo:
Awesome, this include uninstall also works now! :tada: 
Thanks again for watching acados!

PR #570, opened 2020-03-26
Python detect dimensions - replacing #569

FreyJo:
Replacing https://github.com/acados/acados/pull/569 for more compact diff view.

On top of https://github.com/acados/acados/pull/559, this contains

- Python:
    - all dimensions are being detected (except N)
   - setting lower bounds on slack variables is optional, zeros will be used by default
  - add slacked general linear constraints
   - getters for slack variables
- Templates:
   - fix parameter initialization and update
   - add missing `free`s


examples/acados_python/getting_started/minimal_example_closed_loop.py
```diff
@@ -55,12 +55,14 @@
...
 
 # set dimensions
-ocp.dims.nx  = nx
-ocp.dims.ny  = ny
-ocp.dims.ny_e = ny_e
-ocp.dims.nbu = nu
-ocp.dims.nu  = nu
-ocp.dims.N   = N
+ocp.dims.N = N
+# NOTE: all dimensions but N are now detected automatically in the
```
giaf: What about soft constraints, are they detected? Today I realized soft constraints are not detected also in the matlab interface ATM.

FreyJo: In Python, I also did it for the slack dimensions, here:
https://github.com/FreyJo/acados/blob/b60dbfa922e48e4ab7bd849f2cca0a7b57f26076/interfaces/acados_template/acados_template/acados_ocp_solver.py#L224-L367

I think it would be great to also do this in Matlab.

examples/acados_python/tests/soft_constraint_test.py
```diff
@@ -61,23 +61,14 @@ def run_closed_loop_experiment(FORMULATION):
     N = 20
 
     # set dimensions
-    ocp.dims.nx  = nx 
-    ocp.dims.ny  = ny 
-    ocp.dims.ny_e = ny_e 
-    ocp.dims.nu  = model.u.size()[0]
     ocp.dims.ns = nu
```
zanellia: maybe add a comment on why these dims are specified and others not?

interfaces/acados_template/acados_template/acados_ocp.py
```diff
@@ -217,196 +229,210 @@ def nx(self, nx):
         if type(nx) == int and nx > 0:
             self.__nx = nx
         else:
-            raise Exception('Invalid nx value. Exiting.')
+            raise Exception('Invalid nx value, expected integer. Exiting.')
```
zanellia: very minor remark, but the exception message does not match exactly the reason why it was raised. It could say `expected positive integer` or `non negative` for some others below

interfaces/acados_template/acados_template/acados_ocp_solver.py
```diff
@@ -96,6 +81,291 @@ def make_ocp_dims_consistent(acados_ocp):
...
         dims.np = casadi_length(model.p)
+    if acados_ocp.parameter_values.shape[0] != dims.np:
+        raise Exception('inconsistent dimension np, regarding model.p and parameter_values.')
+
+    ## cost
+    # path
+    if cost.cost_type == 'LINEAR_LS':
+        ny = cost.W.shape[0]
+        if cost.Vx.shape[0] != ny or cost.Vu.shape[0] != ny:
+            raise Exception('inconsistent dimension ny, regarding W, Vx, Vu.')
```
zanellia: How about printing the actual dimensions that are causing the Exception? Given that the dimensions are now detected and not set by the user, I guess you would need to print something like 

`'[...] W[ny,ny], Vx[ny,nx], Vu[ny,nu], you have: W[{},{}], Vx[{},{}], Vu[{},{}]'.format( [...]` 

I think it would improve a lot user friendliness :)

examples/acados_python/tests/soft_constraint_test.py
FreyJo: actually also `ns*` are detected. addressed..

interfaces/acados_template/acados_template/acados_ocp.py
FreyJo: right!
addressed, even though the idea is that these setters dont really have to be called anymore :)

PR #573, opened 2020-03-30
Nonuniform templates

FreyJo:
- nonuniform discretization unified for Matlab & Python
- Python: add slacked general linear constraints

Compare these two examples in Python to see how the front end looks like:
examples/acados_python/getting_started/ocp/minimal_example_ocp.py
examples/acados_python/getting_started/ocp/nonuniform_discretization_example.py

Was requested in the forum:
https://discourse.acados.org/t/discrete-dynamics-or-variable-sampling-time/121

### DIFF
--- a/interfaces/acados_matlab_octave/acados_ocp.m
+++ b/interfaces/acados_matlab_octave/acados_ocp.m
@@ -86,8 +86,8 @@ classdef acados_ocp < handle
                 obj.model_struct = detect_constr(obj.model_struct, 1);
             end
 
-            % detect dimensions
-            obj.model_struct = detect_dims_ocp(obj.model_struct);
+            % detect dimensions & sanity checks
+            [obj.model_struct, obj.opts_struct] = detect_dims_ocp(obj.model_struct, obj.opts_struct);
 
             % compile mex interface (without model dependency)
             if ( strcmp(obj.opts_struct.compile_interface, 'true') )
--- a/interfaces/acados_matlab_octave/acados_ocp.m
+++ b/interfaces/acados_matlab_octave/acados_ocp.m
@@ -39,6 +39,7 @@ classdef acados_ocp < handle
         model_struct
         opts_struct
         acados_ocp_nlp_json
+        ext_fun_type
     end % properties
 
 
@@ -65,6 +66,9 @@ classdef acados_ocp < handle
                     obj.model_struct = get_gnsf_structure(obj.model_struct);
                 end
             end
+
+            % store ext_fun_type
+            obj.ext_fun_type = obj.model_struct.ext_fun_type;
 
             % detect cost type
             if (strcmp(obj.model_struct.cost_type, 'auto'))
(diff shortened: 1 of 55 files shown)


interfaces/acados_matlab_octave/acados_ocp.m
```diff
@@ -39,6 +39,7 @@
         model_struct
         opts_struct
         acados_ocp_nlp_json
+        ext_fun_type
```
giaf: I'm not 100% convinced about adding this alone here, but for now it's fine, we can always change later on

giaf:
It's fine with me.

interfaces/acados_template/acados_template/acados_layout.json
```diff
@@ -569,6 +569,12 @@
         ]
     },
     "solver_options": {
+        "time_steps": [
```
zanellia, line 572: "time_steps" could be a bit confusing (one could think of integrator steps or something else), how about using "shooting_nodes" and "delta_shooting_nodes" or something like that?

interfaces/acados_template/acados_template/acados_sim.py
```diff
@@ -73,28 +73,28 @@ def nx(self, nx):
...
-            raise Exception('Invalid nx value. Exiting.')
+            raise Exception('Invalid nx value, expected positive integer. Exiting.')
 
     @nz.setter
     def nz(self, nz):
         if type(nz) == int and nz > -1:
             self.__nz = nz
         else:
-            raise Exception('Invalid nz value. Exiting.')
+            raise Exception('Invalid nz value, expected positive integer. Exiting.')
```
zanellia, line 83: non negative :p

interfaces/acados_template/acados_template/acados_layout.json
FreyJo: I was not using `shooting_nodes` in the templates, because they have size N+1 -> you cannot do check against layout :sweat_smile: 
Moreover, there is the option to set `shooting_nodes` and it will be translated into time_steps, both in Matlab and Python, see e.g.:
https://github.com/acados/acados/pull/573/files#diff-b03620fce6e237f180ca3a6ba33659bf

I guess "thinking of integrator steps" is actually a good way to think about it, since they are the step sizes for the integrators..
Thus, I am not fully convinced about changing to `delta_shooting_nodes`.

interfaces/acados_template/acados_template/acados_sim.py
FreyJo: right! :P

interfaces/acados_template/acados_template/acados_layout.json
zanellia: I guess `time_steps` is fine then. It's just that I tend to think about "intermediate" integrator steps when I read `time_steps`, but I guess this is anyway a near-best choice.

PR #574, opened 2020-04-01
Levenberg Marquardt Regularization term

FreyJo:
Only the last commit is not included in #573 
https://github.com/acados/acados/pull/574/commits/2d6ecebd954161533f8560fd5cebdfd8a15cb1f2

### DIFF
--- a/acados/ocp_nlp/ocp_nlp_common.c
+++ b/acados/ocp_nlp/ocp_nlp_common.c
@@ -1052,6 +1052,8 @@ void ocp_nlp_opts_initialize_default(void *config_, void *dims_, void *opts_)
 #endif
 
     opts->step_length = 1.0;
+    opts->levenberg_marquardt = 0.0;
+
 
     // submodules opts
 
@@ -1169,6 +1171,11 @@ void ocp_nlp_opts_set(void *config_, void *opts_, const char *field, void* value
         {
             double* step_length = (double *) value;
             opts->step_length = *step_length;
+        }
+        else if (!strcmp(field, "levenberg_marquardt"))
+        {
+            double* levenberg_marquardt = (double *) value;
+            opts->levenberg_marquardt = *levenberg_marquardt;
         }
         else if (!strcmp(field, "exact_hess"))
         {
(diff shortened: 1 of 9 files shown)


acados/ocp_nlp/ocp_nlp_common.c
```diff
@@ -1834,6 +1841,10 @@ void ocp_nlp_approximate_qp_matrices(ocp_nlp_config *config, ocp_nlp_dims *dims,
         // init Hessian to 0 
         blasfeo_dgese(nu[i] + nx[i], nu[i] + nx[i], 0.0, mem->qp_in->RSQrq+i, 0, 0);
 
+        // Levenberg Marquadt term
+        if (opts->levenberg_marquadt > 0.0)
```
zanellia, line 1845: why not moving the check into the setter and throw and error if the value is negative?

zanellia: moreover, I guess it would make most sense to scale the contribution by the discretization step length.

FreyJo: > why not moving the check into the setter and throw and error if the value is negative?

There is such a check in the Python setter.
I put the if here, just to skip the operation if `levenberg_marquardt` is not used.
Not sure if it is a good idea, since it is such a cheap operation though.

giaf:
I would also scale by the length of the discretization step.

As a side note, are you aware that it's also possible to regularize the Hessian with a diagonal term in HPIPM? The effect is not the same, as in HPIPM the regularization is only applied to the KKT solution step and not to the residuals, so the solver should still converge (more slowly) to the original QP solution.

FreyJo:
I fixed the misspelling and scaled the regularization with the discretization time step, using 1.0 for the terminal Hessian block.

@giaf I was not aware of this HPIPM option.
How is the option called?

I guess we can still merge this, since it adds the possibility for all qp solvers (?)

zanellia:
I guess what @giaf means is a different thing. When using LVM you do want to change the solution of the QP... Or am I missing something?

giaf:
> I guess what @giaf means is a different thing. When using LVM you do want to change the solution of the QP... Or am I missing something?

Yes sure they are two different things.
I just wanted to make people aware that there are other options too.
https://github.com/giaf/hpipm/blob/master/include/hpipm_d_ocp_qp_ipm.h#L158

FreyJo:
Alright, so it can be merged as is?

FreyJo:
> you do want to change the solution of the QP

it is more like, you expect that behaviour when you use LVM, right?
There is no motivation to change the solution of the QP.

zanellia:
> > you do want to change the solution of the QP
> 
> it is more like, you expect that behaviour when you use LVM, right?
> There is no motivation to change the solution of the QP.

What do you mean? When you use LVM you do want to change the solution to the QPs wrt not-using-LVM. Otherwise you would be taking the same steps... What @giaf was pointing out is that you could change the Hessian inside the QP solver to fix some numerical issues, but still converging to the "exact" solution (i.e without messing with the rhs). But this is a different thing, as he confirmed.

FreyJo:
> (I have the feeling we are already all on the same page :P)

I think too, thanks!  :ok_hand:

PR #576, opened 2020-04-02
s-fun templates: fix pointer. avoids segfault and matlab crash

ardayigit:
Matlab crashes when using the generated s-function with a segfault (including when using the example getting_started - simulink_model_closed_loop).


giaf:
Thanks for spotting this!

PR #577, opened 2020-04-02
fix expected identifier error. happens when installing hpipm and osqp

ardayigit:
After cmake with ACADOS_WITH_OSQP=ON, during install, expected identifier error happens because osqp/include/constants.h defines MAX_ITER, and hpipm/include/hpipm_common.h has an enum with a MAX_ITER element. osqp includes need to be done after hpipm includes.


FreyJo:
Thanks for spotting!

PR #597, opened 2020-04-22
 Memory management - some fixes wrt 32 Bit architectures

FreyJo:
Replacing #593:

I added the file:
`docs/memory_management.md`, which is intended for developers and documents how the `*_calculate_size`, `*_assign` functions should look like.

I made the `minimal_ocp_example` work on the Raspberry Pi 2 instance in the office, by adapting some of the core functions wrt the guidline file.
https://discourse.acados.org/t/convergence-failure-on-embedded-platform/122/6

Note: that for now, this does not mean that there is full acados support on 32 Bit architectures.
The next steps to do on a 32 Bit machine would be
- [ ] make C tests / examples work
- [ ] make the Python tests work
- [ ] ideally have some CI to check some form of 32 Bit compatibility.

I suggest to copy these points (and further ones from discussion) to a new issue and close related outdated issues.

For now the official statement should still be that acados does not support 32 Bit architectures.

Feedback very welcome!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -103,7 +103,7 @@ endif()
 
 if(CMAKE_CXX_COMPILER_ID MATCHES "GNU" OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
     set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=c99 -Wall -fdiagnostics-show-option")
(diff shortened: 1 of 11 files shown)


giaf:
Thanks for implementing these fixes!

Just to understand, besides the minimal example, how are the other tests doing?
Do you know what else is still failing and should be fixed?
What about the QP solvers? Found any issue there?

FreyJo:
The pure simulation tests all work.
I think there are still problems in the QP/condensing part.
Most tests of the OCP/QP tests still fail.
I just managed to fix the wind_turbine_nmpc_example with qpOASES after 5f9ede2.

PR #602, opened 2020-04-29
OCP QP C interface

FreyJo:


examples/c/ocp_qp.c
```diff
@@ -56,65 +53,67 @@ int main() {
...
+    {
+        ocp_qp_xcond_solver_dims_set(config, dims, i, "nx", &nx);
+        ocp_qp_xcond_solver_dims_set(config, dims, i, "nu", &nu);
+    }
+    // last stage has no controls
+    ocp_qp_xcond_solver_dims_set(config, dims, N+1, "nu", &nu_e);
+
+    ocp_qp_xcond_solver_dims_set(config, dims, 0, "nbx", &nx);
+
+    ocp_qp_in *qp_in = ocp_qp_in_create_from_xcond_dims(dims);
```
giaf, line 80: I wouldn't go for this.

Here IMO the correct thing to do is to create a `qp_dims` of type `ocp_qp_dims` to set the QP up, and a different `solver_dims` of type `ocp_qp_xcond_solver_dims` to set the solver up.

So there is no need for this ` ocp_qp_in_create_from_xcond_dims`.
At most, there could be a routine to conveniently fill in the `solvers_dims` by using a previously-created `qp_dims`.
This routine could internally use this HPIPM routine
https://github.com/giaf/hpipm/blob/master/ocp_qp/x_ocp_qp_dim.c#L148
in order to assure maintainability in case of changes in HPIPM.

examples/c/ocp_qp.c
```diff
@@ -128,16 +127,16 @@ int main() {
      ************************************************/
 
     double res[4];
-    ocp_qp_inf_norm_residuals(&dims, qp_in, qp_out, res);
+    ocp_qp_inf_norm_residuals(dims->orig_dims, qp_in, qp_out, res);
```
giaf, line 130: The struct members should not be accessed directly, as this will break in case of changes.
The correct solution is the one from the other comment, with two dims structures, one for the QP (to use here) and another one for the solver.

FreyJo:
@giaf thanks for the review!
I think the way it is now, with the `qp_dims` and `solver_dims` it is both clean and flexible.

PR #604, opened 2020-05-07
Exact Hessian for MEX templating, hessian for nonlinear constraints with algebraic variables

FreyJo:
This contains:
- Exact Hessian for MEX templating
- Hessian for nonlinear constraints with algebraic variables
   - changed signature of `nl_constr_h_fun_jac_hess` and genrating files in Matlab and Python accordingly
- Matlab: getters for `sl`, `su`, `t`, requested https://discourse.acados.org/t/slack-variable-values/119/6
- C: fixes to print level

PR #605, opened 2020-05-13
 Exact Hessian for MEX templating, hessian for nonlinear constraints with algebraic variables, GNSF fixes

FreyJo:
Rebased https://github.com/FreyJo/acados/pull/46 on top of https://github.com/acados/acados/pull/604

This contains:
- Exact Hessian for MEX templating
- Hessian for nonlinear constraints with algebraic variables
   - changed signature of `nl_constr_h_fun_jac_hess` and genrating files in Matlab and Python accordingly
- Matlab: getters for `sl`, `su`, `t`, requested https://discourse.acados.org/t/slack-variable-values/119/6
- C: fixes to print level
- GNSF fix: leading to invalid write.
- GNSF Python: dump DAE from Python & call Octave from Python for GNSF detection
- Templates: added main_sim.in.c to test standalone integrator.


acados/ocp_nlp/ocp_nlp_constraints_bgh.c
```diff
@@ -1307,19 +1315,37 @@ void ocp_nlp_constraints_bgh_update_qp_matrices(void *config_, void *dims_, void
...
+            ext_fun_out[4] = &work->hess_z;  // hess_z*mult: nz * nz
 
             model->nl_constr_h_fun_jac_hess->evaluate(model->nl_constr_h_fun_jac_hess,
                     ext_fun_type_in, ext_fun_in, ext_fun_type_out, ext_fun_out);
 
+            // tmp_nv_nv += dzdxu^T * (hess_z * dzdxu)
+            blasfeo_dgemm_nt(nz, nu+nx, nz, 1.0, &work->hess_z, 0, 0, memory->dzduxt, 0, 0,
+                             0.0, &work->tmp_nz_nv, 0, 0, &work->tmp_nz_nv, 0, 0);
+            blasfeo_dgemm_nn(nu+nx, nu+nx, nz, 1.0, memory->dzduxt, 0, 0, &work->tmp_nz_nv, 0, 0,
+                             1.0, &work->tmp_nv_nv, 0, 0, &work->tmp_nv_nv, 0, 0);
```
giaf, line 1339: If you know the result matrix is symmetric (as it is the case here if `hess_z` is symmetric), you can compute as
`( dzduxt * hess_z' ) * dzduxt'`
where first product is done using `dgemm_nt` and the second using `dsyrk_ln`, which only computes the lower triangular part of `tmp_nv_nv` (and therefore saving half the flops in the second product compared to using `dgemm_nt`).
Also, in general in the BLASFEO API `dgemm_nt` is more efficient than `dgemm_nn`.

giaf: If you need the full `tmp_nv_nv`, you can copy and transpose the lower triangular part in the upper triangular part with `dtrtr_l`.

FreyJo: Thanks for checking!
You are right, the symmetry should be exploited there.
However, since this part is not actually tested. I just added the more efficient way as a comment for now, to be on the saver side.

PR #610, opened 2020-06-02
Add option to use numerical hessian approximation for contribution of external cost module

FreyJo:
add option to use custom hessian approximation for external cost module

### DIFF
--- a/acados/ocp_nlp/ocp_nlp_cost_external.c
+++ b/acados/ocp_nlp/ocp_nlp_cost_external.c
@@ -172,8 +172,8 @@ void *ocp_nlp_cost_external_model_assign(void *config_, void *dims_, void *raw_m
 
     // blasfeo_mem align
     align_char_to(64, &c_ptr);
-    // custom_hessian
-    assign_and_advance_blasfeo_dmat_mem(nx+nu, nx+nu, &model->custom_hessian, &c_ptr);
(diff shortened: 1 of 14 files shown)


acados/ocp_nlp/ocp_nlp_cost_external.c
```diff
@@ -217,6 +219,15 @@ int ocp_nlp_cost_external_model_set(void *config_, void *dims_, void *model_,
     {
         model->ext_cost_fun_jac_hess = (external_function_generic *) value_;
     }
+    else if (!strcmp(field, "ext_cost_fun_jac"))
+    {
+        model->ext_cost_fun_jac = (external_function_generic *) value_;
+    }
+    else if (!strcmp(field, "cost_custom_hess"))
```
giaf, line 226: maybe `ext_cost_custom_hess` is more in line with the other namings

giaf: or even maybe just 'ext_cost_hess'

FreyJo: I think I renamed all the fields as we discussed.

giaf: yes that was before our discussion

PR #619, opened 2020-07-29
Interface fixes

FreyJo:
Mostly added support for external cost for MEX templating, see
https://discourse.acados.org/t/use-solver-in-c-environment/206


giaf:
The removal of the `cost_` prefix in the matlab interface goes against the "convention" of having the module as prefix.
This convention is exploited in several places to e.g. route options to the correct path though the hierarchy of modules, and in the specific case of `cost_` it is used here
https://github.com/acados/acados/blob/master/interfaces/acados_matlab_octave/acados_ocp_model.m#L82

Sure this specific change only regards the name of the generated functions, but it goes against the direction of trying to unify the naming acados-wide.
So I would be more in favor of keeping `cost_ext_` prefix and possibly go from `cost_ext_cost_...` to `cost_ext_...` if the issue is the unpleasant repetition of `cost` in the name.

FreyJo:
> The removal of the `cost_` prefix in the matlab interface goes against the "convention" of having the module as prefix.
> This convention is exploited in several places to e.g. route options to the correct path though the hierarchy of modules, and in the specific case of `cost_` it is used here
> https://github.com/acados/acados/blob/master/interfaces/acados_matlab_octave/acados_ocp_model.m#L82
> 
> Sure this specific change only regards the name of the generated functions, but it goes against the direction of trying to unify the naming acados-wide.
> So I would be more in favor of keeping `cost_ext_` prefix and possibly go from `cost_ext_cost_...` to `cost_ext_...` if the issue is the unpleasant repetition of `cost` in the name.

Yes, the duplication of `cost` in the name is not that nice in my opinion.
However, I now changed it back in Matlab and made the templates follow the same naming convention.
I would like to rename the cost type `ext_cost` to `external` to avoid this duplication in the name at some point.

giaf:
Thanks for considering my opinion.
Nothing against changing `ext_cost` to `external` or whatever you like, my comment was only about the prefix part and its use in acados.

PR #621, opened 2020-07-31
update hpipm & blasfeo

giaf:


giaf:
This should provide some bug fixes.

PR #622, opened 2020-08-10
Support for discrete dynamics in template workflow

FreyJo:
only from Matlab for now.
Thanks to @martkir0919

TODO:
- [x] discrete dynamics for python


FreyJo:
Can be merged like this IMO.

giaf:
I just checked the matlab part, it looks good to me.

PR #626, opened 2020-08-22
Missing Jacobians in make_sfun.in.m

mindThomas:
Added missing CasADi-generated Jacobian files for the external cost in the MATLAB template.


FreyJo:
Thanks for contributing! :ok_hand:

PR #620, opened 2020-07-31
Race cars july2020

FreyJo:
Changed some settings in the race car example.
Use new file, to keep the settings file match the settings in the publication.
https://cdn.syscop.de/publications/Kloeser2020.pdf

PR #628, opened 2020-08-25
Fix racecars example by using CasADi function

mindThomas:
When defining the dynamics the built-in CasADi functions should be used instead of the numeric functions available in numpy.


FreyJo:
That change is also contained in https://github.com/acados/acados/pull/620
I would like to merge this instead.
I hope that's fine with you.

nutonomy-thomas:
No worries at all.

nutonomy-thomas:
No worries at all.

mindThomas:
No worries at all.

PR #629, opened 2020-08-25
Ported race-cars example to MATLAB

mindThomas:
Here is the first take on a port of the race-cars example to MATLAB.


FreyJo:
Merged via https://github.com/acados/acados/pull/631

PR #640, opened 2020-09-28
Add cost computation for ocp nlp, fix cost computation for NLS & LLS

FreyJo:


acados/ocp_nlp/ocp_nlp_common.c
```diff
@@ -2593,3 +2593,25 @@ void ocp_nlp_res_compute(ocp_nlp_dims *dims, ocp_nlp_in *in, ocp_nlp_out *out, o
...
+    {
+        config->cost[ii]->compute_fun(config->cost[ii], dims->cost[ii], in->cost[ii],
+                    opts->cost[ii], mem->cost[ii], work->cost[ii]);
+        tmp_cost = config->cost[ii]->memory_get_fun_ptr(mem->cost[ii]);
+        // printf("cost at stage %d = %e\n", ii, *tmp_cost);
+        total_cost += *tmp_cost;
+    }
+    mem->cost_value = total_cost;
+
+    printf("\ncomputed total cost: %e\n", total_cost);
```
giaf, line 2615: I guess this should be commented out

acados/ocp_nlp/ocp_nlp_common.h
```diff
@@ -89,13 +89,16 @@ typedef struct
...
+    void (*eval_param_sens)(void *config, void *dims, void *opts_, void *mem, void *work,
+                            char *field, int stage, int index, void *sens_nlp_out);
     // prepare memory
     int (*precompute)(void *config, void *dims, void *nlp_in, void *nlp_out, void *opts_, void *mem, void *work);
     // initalize this struct with default values
     void (*config_initialize_default)(void *config);
     // general getter
     void (*get)(void *config_, void *dims, void *mem_, const char *field, void *return_value_);
+    void (*opts_get)(void *config_, void *dims, void *opts_, const char *field, void *return_value_);
+    void (*work_get)(void *config_, void *dims, void *work_, const char *field, void *return_value_);
```
giaf, line 101: why do you need to get opts and work of the nlp?

FreyJo: I needed them here:
https://github.com/acados/acados/blob/3ac5ab40a79ada831e0c6f8d9e2c2617f3a04bab/interfaces/acados_c/ocp_nlp_interface.c#L913-L926

acados/ocp_nlp/ocp_nlp_common.c
FreyJo: Right!

acados/ocp_nlp/ocp_nlp_common.h
giaf: I was just wondering why get opts and work has never been necessary, probably there is another way of doing it without using them.

giaf: I guess the other way of doing it would be to do it inside the sqp and sqp_rti solvers, instead of exposing the nlp_opts, nlp_mem, nlp_work as stand-alone.

FreyJo: exactly, the alternative was to duplicate the code for the function I mentioned for SQP and SQP_RTI.

giaf: Not necessarily. The function could be implemented once in nlp_common, and there would just be two wrappers in SQP and SQP_RTI.

PR #641, opened 2020-10-02
OpenMP parallelization in CMake build system

FreyJo:
Added `OpenMP` parallelization in `CMake` build system.

If `ACADOS_WITH_OPENMP` is `ON`, cmake will try to find OpenMP (default is `ON`).
If OpenMP is found and `ACADOS_NUM_THREADS` is provided, it will use that many threads.
If OpenMP is found and `ACADOS_NUM_THREADS` is NOT provided, it will call `omp_get_max_threads()` to get the number of threads.

If `ACADOS_WITH_OPENMP` is `ON` (default), we now require cmake >= 3.9.
If you want to use an older `CMake` version, turn `ACADOS_WITH_OPENMP`off.


CMakeLists.txt
```diff
@@ -216,3 +242,15 @@ message(STATUS "Octave Templates (${ACADOS_OCTAVE_TEMPLATE})")
 message(STATUS "System name:version ${CMAKE_SYSTEM_NAME}:${CMAKE_SYSTEM_VERSION}")
 message(STATUS "Build type is ${CMAKE_BUILD_TYPE}")
 message(STATUS "Installation directory is ${CMAKE_INSTALL_PREFIX}")
+if(ACADOS_WITH_OPENMP)
+    message(STATUS "OpenMP parallelization is ON")
+    if(ACADOS_NUM_THREADS)
+        message(STATUS "OpenMP number of threads ${ACADOS_NUM_THREADS}")
+    else()
+        message(STATUS "OpenMP number of threads not set, will use omp_get_max_threads")
```
giaf, line 250: That's the only remark: to be more specific staying something like "ACADOS_NUM_THREADS not set, will use omp_get_max_threads instead"

giaf: Otherwise it can be merged.

PR #642, opened 2020-10-07
Interface improvements

FreyJo:
- MEX interface:
    - detect all dimensions (also slacks) like in Python.
    - remove setting dimensions in examples and tests (they would be overwritten anyway by the interface now)
- MEX templating: 
     - fix dimension bug
     - add GNSF support
- Python: added closed loop OCP + MHE example, in colaboration with Katrin
- Update Roadmap


giaf:
It looks fine for me.

PR #644, opened 2020-10-13
QP solvers: work on qpDUNES, OSQP

FreyJo:
Main features of this PR:
- C: make QP solver enum independent of compiler flags
- C: wrap qpDUNES and OSQP options properly
- Add support and test for SQP with qpDUNES and OSQP through Matlab & Python interfaces.
- These tests are run on travis

Thanks to @asparc for 68c3ffb 
I guess this fixes https://github.com/acados/acados/issues/651

There were some weird problems with Octave, that I did not have time to investigate.
Now every MEX test has it's own run file that can exit with status 1, if it fails.
This is unfortunate code duplication. But I dont have more time to spend on debugging travis right now.


interfaces/acados_c/ocp_qp_interface.h
```diff
@@ -63,25 +63,39 @@ typedef enum {
     PARTIAL_CONDENSING_HPIPM,
 #ifdef ACADOS_WITH_HPMPC
     PARTIAL_CONDENSING_HPMPC,
+#else
+    PARTIAL_CONDENSING_HPMPC_NOT_AVAILABLE,
```
giaf, line 67: I think this should work fine, ATM I don't see cases where this would create issues.

asparc:
Thanks for all the work here! Unfortunately, I'm afraid that the changes don't solve #651 (I also don't have a satisfactory solution for this one, only the quick patch that I mentioned in #651). When trying this PR, I also still needed most of the changes in #654 to run the minimal examples on Windows.

acados/ocp_qp/ocp_qp_osqp.c
```diff
@@ -658,35 +659,51 @@ void ocp_qp_osqp_opts_set(void *config_, void *opts_, const char *field, void *v
...
         int *tmp_ptr = value;
         opts->osqp_opts->max_iter = *tmp_ptr;
     }
     else if (!strcmp(field, "tol_stat"))
     {
-        // TODO set solver exit tolerance
+        double *tol = value;
+        // opts->osqp_opts->eps_abs = *tol;
+        opts->osqp_opts->eps_rel = *tol;
+        opts->osqp_opts->eps_dual_inf = *tol;
```
giaf, line 674: If this is the tolerance before the polishing step, it could take a _lot_ of iterations to get there.
The most efficient way of using OSQP is to leave a rather low tolerance (as I guess the default is), and hope for the polishing step to correctly guess the active set and consequently return an high-accuracy solution based on the solution of an additional equality-constrained QP.

giaf:
> Thanks for all the work here! Unfortunately, I'm afraid that the changes don't solve #651 (I also don't have a satisfactory solution for this one, only the quick patch that I mentioned in #651). When trying this PR, I also still needed most of the changes in #654 to run the minimal examples on Windows.

I agree this does not solve #651, see my comment there.

But it does anyway some useful polishing work :)

acados/ocp_qp/ocp_qp_osqp.c
FreyJo: As you maybe saw in the commented lines below, I already played around with that.
On the example I tested it with, I reduced the total number of QP iterations from 2500 to 1220.
Full output:
[New options](https://pastebin.com/LnH3eyZZ)
[Old](https://pastebin.com/RzB53bZT)

FreyJo:
@asparc Right! Sorry, think I didn't check your issue correctly.
I will try to address it in another PR and cherry-pick more from #654

PR #645, opened 2020-10-13
avoid SX/MX confusion

jgillis:
While playing around with acados,
there was sometimes an incorrect inference of SX/MX.

Proposed fix is probably not complete, but at least it covered the cases I encountered


FreyJo (approved):
Looks good, thank you Joris!

PR #649, opened 2020-10-15
Minor UI improvements

jgillis:


FreyJo:
Thanks!
Just spotted a minor error in the Tex andmerged #653  instead.

PR #654, opened 2020-10-16
Fixes for Windows 10 + Matlab

asparc:
I addressed a couple of issues to be able to run the minimal_example_ocp/sim.m in Windows + Matlab. I did not touch the octave variations, even though they probably require very similar fixes


interfaces/acados_matlab_octave/acados_ocp.m
```diff
@@ -120,8 +120,11 @@
                     elseif ~isempty(strfind(obj.opts_struct.qp_solver,'osqp'))
                         flag_file = fullfile(obj.opts_struct.output_dir, '_compiled_with_osqp.txt');
                         compile_interface = ~exist(flag_file, 'file');
+                    elseif ~isempty(strfind(obj.opts_struct.qp_solver,'hpipm'))
```
FreyJo, line 123: acados is always be compiled with HPIPM, so this is unnecessary.

interfaces/acados_matlab_octave/ocp_compile_interface.m
```diff
@@ -125,14 +134,26 @@ function ocp_compile_interface(opts)
...
 if with_qp_hpmpc
-    % flag file to remember if compiled with OSQP
+    % flag file to remember if compiled with HPMPC
     flag_file = fullfile(opts.output_dir, '_compiled_with_hpmpc.txt');
     flagID = fopen(flag_file, 'w');
     fclose(flagID);
+    % additional compilation flag
+    FLAGS = [FLAGS, ' -DACADOS_WITH_HPMPC'];
+end
+% is HPIPM?
```
FreyJo, line 150: See above.

interfaces/acados_matlab_octave/sim_generate_casadi_ext_fun.m
```diff
@@ -99,7 +99,8 @@ function sim_generate_casadi_ext_fun(model_struct, opts_struct)
 end
 
 if ispc
-    mbuild(c_files_path{:}, '-output', lib_name, 'CFLAGS="$CFLAGS"', 'LDTYPE="-shared"', ['LDEXT=', ldext]);
+    % mbuild(c_files_path{:}, '-output', lib_name, 'CFLAGS="$CFLAGS"', 'LDTYPE="-shared"', ['LDEXT=', ldext]);
```
FreyJo, line 102: I am wondering, why `mbuild` was used here instead of the other command in the first place.

FreyJo (commented):
Thanks a lot for this PR! Looks good overall! :ok_hand: 
I would like to test it on Windows myself before merging.
Not sure when I have time for it though.
No need to address the comments, they are just notes.

interfaces/acados_matlab_octave/acados_ocp.m
asparc: If I understand you correctly, C code compiled with qpOASES will also work if you set HPIPM as your solver. I just wonder if the compiled code would be smaller or more efficient in some way if you build with just HPIPM, i.e. without qpOASES.

interfaces/acados_matlab_octave/sim_generate_casadi_ext_fun.m
asparc: I wondered as well. I just know that mbuild is part of a very expensive Matlab toolbox that not many users will have access to. It might however be included in an academic license, explaining why it was used without further considerations.

interfaces/acados_matlab_octave/acados_ocp.m
FreyJo: It is mandatory to compile acados with HPIPM, since HPIPM is used in the core of acados, see e.g.
https://github.com/acados/acados/blob/master/acados/ocp_qp/ocp_qp_common.c#L45
For this reason, there is no need to keep track, if it was compiled with HPIPM.

Then also the `else` branch can set `compile_interface = false;` IMO.
It is basically for the case that, no external QP solver is used, but HPIPM and the interface doesnt have to be recompiled.

The compiled code should be smaller when not compiling with external QP solvers, but not (significantly) more efficient.

asparc: My changes were mainly focused on this last line in your answer. If you choose HPIPM and it detects existing compilation results with an external QP solver, it will rebuild to remove the parts related to the external solver.
But this is clearly not critical.

interfaces/acados_matlab_octave/sim_generate_casadi_ext_fun.m
giaf: On the other hand the `system(...)` command requires `gcc` to be installed system-wide on windows, which is surely not as common and straightforward as on a linux system...

giaf: All in all, IMO the `mbuild(...)` command is more likely to work out-of-the box on a matlab installation on windows system.

asparc: 'gcc' comes with `mingw` right? And `mingw` is part of the installation instructions of acados

FreyJo:
I merged most of the changes here:
https://github.com/acados/acados/pull/659

PR #657, opened 2020-10-23
Update utils.py from acados_template

gmsanchez:
Add CasADi versions to match the condition >=3.5.1 from 8361dea6868744eb921c24327de02223183b98e1


FreyJo:
Thanks, @gmsanchez 
I also did this yesterday here https://github.com/acados/acados/pull/656
So, I merged that instead.
Thanks anyway!

PR #658, opened 2020-10-26
Change constr_lh and constr_uh via MEX interface

martkir0919:


FreyJo:
I merged the commit in https://github.com/acados/acados/pull/659, don't know why travis was failing here..

PR #660, opened 2020-11-04
CMake option to disable console status output

martkir0919:
Default build is still with console output


FreyJo:
Looks fine to me.
Any remarks from `Make` side? @giaf

PR #661, opened 2020-11-08
Fix automatic acadosConfig.cmake generation

mindThomas:
Only look for the external modules needed by the compiled acados installation.


FreyJo:
Just to understand:
Is there an issue using `find_package`?
Or is this just to reduce overhead in the build process?

mindThomas:
I was getting the errors that some of the unused external libraries, which was not installed on the target system, could not be found.
If the user decides to leave them out of the acados build one can not expect these libraries to be present on the system and the acadosConfig file should thus not be looking for them.

PR #662, opened 2020-11-08
Python: fixed bug in passing non-fortran ordered matrices

jgillis:


FreyJo:
Thanks!
Also made this change for `cost_set`.

Just for reference:
https://numpy.org/doc/stable/reference/internals.html

PR #663, opened 2020-11-08
Python: sim -- don't assume lbx_0 is of length nx and ordered

jgillis:
Exposed by 53e06fd90ea706dda5a17bd7b0bd30f6a80c86d7


FreyJo:
Thanks for the fix!

PR #664, opened 2020-11-08
Unload dll at AcadosOcpSolver destruction

jgillis:


jgillis:
[Failing unittest](https://travis-ci.org/github/acados/acados/jobs/742297366#L4265) looks suspicious. Is this an existing bug that got exposed?

FreyJo:
The test was calling the old DLL before your fix, because the solvers to be compared have the same name.
Thanks a lot for this! @jgillis :pray:

PR #669, opened 2020-11-24
Test qpDUNES on Github

FreyJo:
Remove qpDUNES from C mass spring examples/tests.


FreyJo:
Test failed because Octave 4.2.2 was used, see https://github.com/acados/acados/pull/670

PR #671, opened 2020-11-26
Devel

martkir0919:
Reduces the minimum required version of cmake for OpenMP support.


FreyJo:
All changes apart from the ones in eb805c1 which breaks Python are merged via #673

PR #674, opened 2020-11-27
Thread safe templates

FreyJo:
https://discourse.acados.org/t/thread-safe-version/123


giaf:
So the key idea is to remove all the global variables from the header file, and instead define a `capsule` struct to contain them. This capsule would be the first thing to be created and contain the solver and all the external model/cost/constraints functions pointers.

Is this correct? Maybe do you want to add a few comments?

FreyJo:
@giaf pretty well summarized!
I think the templated header file gives a good summary of the commit.

Basically, these 2 functions to create/free the capsule where added.
```
     nlp_solver_capsule *acados_ocp_capsule = {{ model.name }}_acados_create_capsule();
     status = {{ model.name }}_acados_free_capsule(capsule);
```
and all the functions defined in the template to interact with the solver take a pointer to the capsule as an argument.

PR #685, opened 2020-12-23
[bugfix!] adding missing arguments to param update function

nirajrathod:
The declaration of {model.name}_sim_update_param() was missing its arguments so I fixed it on the fly and thought it would be useful to let you know


FreyJo:
Cherry-picked & merged here https://github.com/acados/acados/pull/684
Thanks @nirajrathod !

PR #687, opened 2021-01-06
Hotfixes, Python: store and load current iterate

FreyJo:


FreyJo:
contained here https://github.com/acados/acados/pull/684

PR #689, opened 2021-01-24
change return-type for all *calculate_size functions to configurable type

schwieni:
This allows for changing the type of the return type. 
Previously, the return type was int which limited the max memory to approx 2GB and the solver crashed with larger
problems.
By changing typedef in the acados/utils/type.h the behavior can be changed.

### DIFF
--- a/acados/dense_qp/dense_qp_common.c
+++ b/acados/dense_qp/dense_qp_common.c
@@ -63,9 +63,9 @@
  * config
  ************************************************/
 
-int dense_qp_solver_config_calculate_size()
+acados_size_t dense_qp_solver_config_calculate_size()
 {
-    int size = 0;
+    acados_size_t size = 0;
 
     size += sizeof(qp_solver_config);
 
@@ -90,9 +90,9 @@ qp_solver_config *dense_qp_solver_config_assign(void *raw_memory)
  * dims
  ************************************************/
 
-int dense_qp_dims_calculate_size()
+acados_size_t dense_qp_dims_calculate_size()
 {
-    int size = sizeof(dense_qp_dims);
+    acados_size_t size = sizeof(dense_qp_dims);
 
     size += d_dense_qp_dim_memsize();
 
@@ -135,9 +135,9 @@ void dense_qp_dims_set(void *config_, void *dims_, const char *field, const int*
  * in
  ************************************************/
 
-int dense_qp_in_calculate_size(dense_qp_dims *dims)
+acados_size_t dense_qp_in_calculate_size(dense_qp_dims *dims)
 {
-    int size = sizeof(dense_qp_in);
+    acados_size_t size = sizeof(dense_qp_in);
     size += sizeof(dense_qp_dims);
     size += d_dense_qp_memsize(dims);
     return size;
@@ -179,9 +179,9 @@ dense_qp_in *dense_qp_in_assign(dense_qp_dims *dims, void *raw_memory)
  * out
  ************************************************/
 
-int dense_qp_out_calculate_size(dense_qp_dims *dims)
+acados_size_t dense_qp_out_calculate_size(dense_qp_dims *dims)
 {
-    int size = sizeof(dense_qp_out);
+    acados_size_t size = sizeof(dense_qp_out);
     size += d_dense_qp_sol_memsize(dims);
     size += sizeof(qp_info);
     size += 8;
@@ -235,9 +235,9 @@ void dense_qp_out_get(dense_qp_out *out, const char *field, void *value)
  * res
  ************************************************/
 
-int dense_qp_res_calculate_size(dense_qp_dims *dims)
+acados_size_t dense_qp_res_calculate_size(dense_qp_dims *dims)
 {
-    int size = sizeof(dense_qp_res);
+    acados_size_t size = sizeof(dense_qp_res);
     size += d_dense_qp_res_memsize(dims);
 
     make_int_multiple_of(8, &size);
@@ -263,9 +263,9 @@ dense_qp_res *dense_qp_res_assign(dense_qp_dims *dims, void *raw_memory)
 
 
 
-int dense_qp_res_workspace_calculate_size(dense_qp_dims *dims)
+acados_size_t dense_qp_res_workspace_calculate_size(dense_qp_dims *dims)
 {
-    int size = sizeof(dense_qp_res_ws);
+    acados_size_t size = sizeof(dense_qp_res_ws);
 
     size += d_dense_qp_res_ws_memsize(dims);
     make_int_multiple_of(8, &size);
--- a/acados/dense_qp/dense_qp_common.h
+++ b/acados/dense_qp/dense_qp_common.h
@@ -59,15 +59,15 @@ typedef struct d_dense_qp_res_ws dense_qp_res_ws;
 typedef struct
 {
     void (*dims_set)(void *config_, void *dims_, const char *field, const int* value);
-    int (*opts_calculate_size)(void *config, void *dims);
+    acados_size_t (*opts_calculate_size)(void *config, void *dims);
     void *(*opts_assign)(void *config, void *dims, void *raw_memory);
     void (*opts_initialize_default)(void *config, void *dims, void *args);
     void (*opts_update)(void *config, void *dims, void *args);
     void (*opts_set)(void *config_, void *opts_, const char *field, void* value);
-    int (*memory_calculate_size)(void *config, void *dims, void *args);
+    acados_size_t (*memory_calculate_size)(void *config, void *dims, void *args);
     void *(*memory_assign)(void *config, void *dims, void *args, void *raw_memory);
     void (*memory_get)(void *config_, void *mem_, const char *field, void* value);
-    int (*workspace_calculate_size)(void *config, void *dims, void *args);
+    acados_size_t (*workspace_calculate_size)(void *config, void *dims, void *args);
     int (*evaluate)(void *config, void *qp_in, void *qp_out, void *args, void *mem, void *work);
     void (*eval_sens)(void *config, void *qp_in, void *qp_out, void *opts, void *mem, void *work);
 } qp_solver_config;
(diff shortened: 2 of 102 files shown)


FreyJo:
Hi,
thanks for the PR, I think this makes sense!
Probably, you also want to apply the corresponding change to the functions in [interfaces/acados_c](https://github.com/acados/acados/tree/master/interfaces/acados_c)
Otherwise, this doesn't solve your issue, right?

schwieni:
Hi,
as suggested ;-)
Can you test this? I did run the test/unit_tests without issues.

schwieni:
Hi,
I digged a little deeper and I found that external solvers have a similar issue.
Example: acados/acados/dense_qp/dense_qp_common.c: dense_qp_in_calculate_size() calls d_dense_qp_memsize() which is referenced with acados/external/hpipm/dense_qp/x_dense_qp.c:38 int DENSE_QP_MEMSIZE. So, in this case we would need the same type. However, since it is an external library this might not be so easy.
I found that this seems to be an issue with large problems and full condensing.
Example:
`acados_ocp = AcadosOcp()`
`sol_opts = acados_ocp.solver_options`
`sol_opts.qp_solver = 'FULL_CONDENSING_QPOASES'`
while
`sol_opts.qp_solver = 'PARTIAL_CONDENSING_HPIPM'`
does not seem to pose such a problem.
I am afraid that I will be unable to solve such issues (int type of sizes in external libs), since I am not too deep in the code, sorry. And it also depends on external repositories. It seems to be a design "issue" also of HPIPM.
Do you have an idea?

FreyJo:
Yes, I HPIPM has the same issue and acados is using HPIPM for the QP formulation.
For me, I ran into the issue, when compiling acados with openMP parallelization and attempting to solve a huge problem.
When compiled with openMP, acados does not reuse as much memory across the modules and thus needs more memory overall.
So you still have the issue after the latest commits depending on the QP solver?
Your problem must be really huge, because as I understand, the QP structs borrowed from HPIPM need more than 2GB in your case.
If this is the case, I guess similar changes could be done in HPIPM.

schwieni:
Indeed I used OpenMP (cmake ACADOS_WITH_OPENMP was on). However, I tried to deactivate it and it still crashes but only with `sol_opts.qp_solver = 'FULL_CONDENSING_QPOASES'` (see above).
The problem becomes big when I use longer horizons.
I don't know how to check how large the structs borrowed from HPIPM become. Any hints? How would you do that?

I managed to improve the situation also for 'PARTIAL_CONDENSING_HPIPM' when just changing the type from `int memsize;` to `size_t` in structs `d_part_cond_qp_arg` and `d_part_cond_qp_ws` in file `external/hpipm/include/hpipm_d_part_cond.h` (and `include/hpipm/include/hpipm_d_part_cond.h`?). "Improve" means that even longer horizons are possible, otherwise the messages "Create_cond_qp_ocp2ocp_arg: outside memory bounds!" appears and the program stops.

Regarding changing the size type in HPIPM:
A simple `grep -r size *` in the `acados/include/hpipm/include` directory shows many references to `int`. So the impact of changing all `int memsize` in HPIPM seems very intrusive. Moreover, it is also a lot of work and the chances to introduce new bugs are high.

I guess at one point solving the whole OCP becomes very inefficient and a hybrid approach towards MPC would be better suited (which also reduces the memory footprint).
What is your opinion?

FreyJo:
>  I don't know how to check how large the structs borrowed from HPIPM become. Any hints? How would you do that?

I just added some print statements, when I had the issue to verify.

I am not really familiar with the condensing routines implemented in HPIPM, but from that error, I suspect the memory for the condensing routine becomes too big.
I suggest to open a separate issue/PR at https://github.com/giaf/hpipm/
I would merge this one here, if you agree?

> I guess at one point solving the whole OCP becomes very inefficient and a hybrid approach towards MPC would be better suited (which also reduces the memory footprint).

I guess for such a huge problem there might be other software with lower memory requirements.
acados & HPIPM are developed to allow for very fast solution and feedback times.
I guess the problems you are considering are not typical for optimization based control and have been out of the scope of the previous development.

schwieni:
> I would merge this one here, if you agree?
Sure I agree!
The only thing where I am unsure are the examples (commit 1a85d9a). At least for me using CMake they did not seem to be included in the build process. Maybe you might want to check them!?
> I suggest to open a separate issue/PR at https://github.com/giaf/hpipm/
I will try to do that!

> I just added some print statements, when I had the issue to verify.
All right, the good old print method is also what I used ;-) I just thought there is some build in statistics which I overlooked.

> I guess for such a huge problem there might be other software with lower memory requirements.
Do you have some hints for that software (candidates)?

> I guess the problems you are considering are not typical for optimization based control and have been out of the scope of the previous development.
Understood. I am still learning and I thought that starting from an OCP instead of jumping right to the end might be a good Idea to get some experience.
Actually, I must say that after clearing all my modelling errors acados (and CasADi and the solvers) gives very good results already and HPIPM is very fast! Congrats! :)
As a next step I would of course like to make steps towards MPC and real-time execution by reducing the horizon length.
Currently, I am using python but I believe I would have to follow the c-examples (model export in Matlab) and then use c/c++ code for running the control loop, since in my case the python-part inside the control loop for filling the yref and preparing the next step seem to be the bottleneck.

schwieni:
FYI
The PR for HPIPM regarding the same configurable size is:
https://github.com/giaf/hpipm/pull/97

FreyJo:
> The only thing where I am unsure are the examples (commit 1a85d9a). At least for me using CMake they did not seem to be included in the build process. Maybe you might want to check them!?

The `C` examples are not maintained, especially the ones in the `no_interface_examples` are rather there for historic reasons.
Thus, I reverted the changes in these files.

> As a next step I would of course like to make steps towards MPC and real-time execution by reducing the horizon length.
Currently, I am using python but I believe I would have to follow the c-examples (model export in Matlab) and then use c/c++ code for running the control loop, since in my case the python-part inside the control loop for filling the yref and preparing the next step seem to be the bottleneck.

That sounds good!
Just to clarify and connected to my previous point. The generated `C` code, i.e. especially the `acados_solver*` files, should be used as a reference on how to use `acados` from `C` rather than the C examples.
For reference, the generated files (that are not from CasADi) are based on these templates:
https://github.com/acados/acados/tree/master/interfaces/acados_template/acados_template/c_templates_tera

I guess this is just the standard overhead of calling `C` from `Python`, or did you experience unusual delays when "preparing the next step"?

> Do you have some hints for that software (candidates)?

Unfortunately not, but it also seems that you are going into another direction now, shortening the horizon looking at real-time feasibility.

schwieni:
First, thank you for the merge! Cool :-)

> The generated C code, i.e. especially the acados_solver* files, should be used as a reference on how to use acados from C rather than the C examples.

Thank you for pointing that out. I was not aware of that. In my opinion the documentation is relatively sparse on that.
Currently, some Matlab export code seems to be located in a directory marked as "experimental". Is there a reason for that? To me this was a hint, that one should use python, since the Matlab part is more experimental. So for my problem I started with the python interface.
I believe new users would definitely benefit from some hints in doc/README files. I certainly would appreciate that.
IMHO the pdf-file in the doc-directory seems a little out-dated here and there and a revision would probably help the users to understand the overall concept better.
Do you plan some activities on documentation in the future?

FreyJo:
> Currently, some Matlab export code seems to be located in a directory marked as "experimental". Is there a reason for that?

Everything in the `experimental` folder can be ignored IMO.
Basically the Matlab interface consists of the Native MEX interface, which is [here](https://github.com/acados/acados/tree/master/interfaces/acados_matlab_octave) and a wrapper to the template based interface [here](https://github.com/acados/acados/tree/master/interfaces/acados_matlab_octave/acados_template_mex/%2Bacados_template_mex)

For documentation, I mainly focus on improving the documentation on docs.acados.org over README files.
The Sections on Matlab and Python [here](https://docs.acados.org/interfaces/) should give a good overview on the interfaces.

If you have more precise points, what is outdated, I will definitely try to update it.

schwieni:
> If you have more precise points, what is outdated, I will definitely try to update it.

Regarding the documentation at https://docs.acados.org/interfaces/, this is very helpful, indeed.

Imho there is still some room for improvement:

https://docs.acados.org/interfaces/#optimal-control-problem-description 
“The Python interface relies on the same problem formulation as the MATLAB interface see here.”
Under “see here” is a pdf-file. Some things in the pdf are probably missing or different, examples:
- The dynamics is described as f_{impl}. This might suggest that only implicit dynamics is possible, which for the python version is not true.
- The “*_type” definitions are different between Matlab and Python. In the pdf there are no acados_template.acados_ocp.AcadosOcpConstraints::idx[bu|bx|bx_0|bx_e|…] described. That was a bit confusing at first.
- some other python-properties are missing in the pdf, e.g., all hints to phi (Jsphi, Jsphi_e, idxsphi, idxsphi_e, lsphi, lsphi_e, uphi, uphi_e, usphi, usphi_e)

Some properties already contain more "meta"-info but most do not have that.

https://docs.acados.org/interfaces/#module-acados_template.acados_ocp and the following:
Just a few examples of what additional information would be helpful to a user for each property imho:
- Which values are optional or set automatically (and maybe under which conditions)?
- If applicable, what is the default value?
- If applicable, what is the matrix/vector dimension? (partially covered in the pdf)
- sometimes the elements of a diagonal matrix have to be given as a vector of the diagonals, e.g., 
- What are the allowed values/strings? Especially for “type” properties,  e.g., globalization, integrator_type, constr_type, cost_type.
- Is there a value range? Example: boolean: True/False, float: 0...3.14, int: -10...10.
- Hints about what are good values or give an example about the influence, e.g., parameters „alpha_min“, „alpha_reduction“, “levenberg_marquardt”.

Some specifics (not an exhaustive list):
- https://docs.acados.org/interfaces/#acados_template.acados_model.AcadosModel:
“disc_dyn_expr” seems to be missing (in conjunction with acados_ocp.solver_options.integrator_type == 'DISCRETE')
- https://docs.acados.org/interfaces/#acados_template.acados_ocp.AcadosOcpCost.yref_e is the description „Vex - x matrix coefficient for Mayer term“ correct?
- https://docs.acados.org/interfaces/#acados_template.acados_sim_solver.AcadosSimSolver the set function seems to be replaced by the cost_set() and constraints_set() functions. These, however, are not mentioned. Some other functions are not mentioned, e.g., print_statistics(), store_iterate(). The description of the get()-function seems outdated (can have the stage number now).

Examples section:
https://docs.acados.org/examples/index.html this section might benefit from more description about the examples.

I can try to help out here and there if you want. I would need advice on how to test the doc system on my computer.

schwieni:
@FreyJo FYI: the PR giaf/hpipm#97 mentioned above is now merged into hpipm master. So, hpipm has also a configurable size type now.

giaf:
The version of HPIPM with the above fixes is used as default in acados with this PR https://github.com/acados/acados/pull/701

PR #697, opened 2021-02-28
acados_template: configurable code export directory

schwieni:
The code export directory can be set in AcadosOcp.code_export_directory and
AcadosSim.code_export_directory.
The constructor of AcadosSimSolver transfers the setting from AcadosOcp to
AcadosSim if initialized with an AcadosOcp object.


FreyJo:
Thanks @schwieni!
That looks good to me and is definitely useful.
We should probably also add this option in Matlab at some point.

Issue #599, opened 2020-04-24
Support 32 Bit Architectures


FreyJo:
I put a bit of work into this, see #597 but there are still some issues left.
For now, acados is not fully supported on 32 Bit architectures, mainly because of memory alignment issues.

In order to be able to make such a statement, the following should be done on a 32 Bit machine:
- [ ] make C tests / examples work
- [ ] make the Python tests work
- [ ] ideally have some CI to check some form of 32 Bit compatibility.

I hope this Guideline will be helpful for future developers/maintainers:
https://github.com/acados/acados/blob/master/docs/memory_management.md

giaf:
A step in the direction of fixing 32-bit support would be to edit these asserts
https://github.com/acados/acados/blob/master/acados/utils/mem.c#L95
https://github.com/acados/acados/blob/master/acados/utils/mem.c#L108
https://github.com/acados/acados/blob/master/acados/utils/mem.c#L121
https://github.com/acados/acados/blob/master/acados/utils/mem.c#L134
https://github.com/acados/acados/blob/master/acados/utils/mem.c#L147
such that they check for 8-byte alignment on 64-bit systems and 4-byte alignment on 32-bit systems.
(BLASFEO structs only contain int and pointers, so they are also good with the alignment for the pointers).

FreyJo:
Just to update this issue:
This PR applied the changes suggested in the comment above: https://github.com/acados/acados/pull/706


> A step in the direction of fixing 32-bit support would be to edit these asserts
> https://github.com/acados/acados/blob/master/acados/utils/mem.c#L95
> https://github.com/acados/acados/blob/master/acados/utils/mem.c#L108
> https://github.com/acados/acados/blob/master/acados/utils/mem.c#L121
> https://github.com/acados/acados/blob/master/acados/utils/mem.c#L134
> https://github.com/acados/acados/blob/master/acados/utils/mem.c#L147
> such that they check for 8-byte alignment on 64-bit systems and 4-byte alignment on 32-bit systems.
> (BLASFEO structs only contain int and pointers, so they are also good with the alignment for the pointers).

PR #704, opened 2021-04-03
Correct pointer alignment checks to support 32-bit architectures

mindThomas:
As part of an effort in getting ACADOS up and running on a 32-bit ARM Cortex-M7 device, the STM32H743, a few missing pointer alignments were identified.
Since the 32-bit architecture uses 32-bit long pointers any memory allocation for pointers would only require 4 bytes which would lead to 8-byte misaligned pointers.
The function call `align_char_to(8, &c_ptr);` has been added to the necessary places to ensure 8-byte pointer alignment where needed.

With the changes ACADOS ran successfully on the embedded ARM processor for which more details can be found here: https://discourse.acados.org/t/working-example-for-stm32h7-cortex-m7-32-bit/406

Related to https://github.com/acados/acados/issues/599

### DIFF
--- a/acados/ocp_nlp/ocp_nlp_common.c
+++ b/acados/ocp_nlp/ocp_nlp_common.c
@@ -861,12 +861,16 @@ ocp_nlp_out *ocp_nlp_out_assign(ocp_nlp_config *config, ocp_nlp_dims *dims, void
     // blasfeo_dvec_struct
     // ux
     assign_and_advance_blasfeo_dvec_structs(N + 1, &out->ux, &c_ptr);
+    align_char_to(8, &c_ptr);
     // z
     assign_and_advance_blasfeo_dvec_structs(N + 1, &out->z, &c_ptr);
+    align_char_to(8, &c_ptr);
     // pi
     assign_and_advance_blasfeo_dvec_structs(N, &out->pi, &c_ptr);
+    align_char_to(8, &c_ptr);
     // lam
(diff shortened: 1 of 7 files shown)


FreyJo:
Hi @mindThomas 
thanks for the PR!

I see that the extra alignment calls here are to pass the assert calls in
- assign_and_advance_double_ptrs
- assign_and_advance_int_ptrs
- assign_and_advance_blasfeo_dvec_structs
- assign_and_advance_blasfeo_dmat_structs
- assign_and_advance_blasfeo_dmat_ptrs
which are there to ensure pointer alignments.

On the 32bit architecture it is sufficient that pointers are aligned to 4 bytes, see https://github.com/acados/acados/issues/599#issuecomment-815854164

Based on this comment, I made the PR https://github.com/acados/acados/pull/706  which I suggest to merge instead of this one here.
If you have time to test the corresponding branch on your hardware and can confirm that it works, that would be awesome.

Best,
Jonathan

mindThomas:
I have tested your code and updated my PR to reflect the same since your PR did not include the extra asserts to check whether memory was actually allocated (this is a crucial check on embedded platforms who has limited memory which would also make troubleshooting easier).

FreyJo:
alright, I don't have a strong opinion on the asserts.
I included them with a more descriptive commit message.
https://github.com/acados/acados/pull/706/commits/f5d29f571d0584cd33d831d879b41d458be9b972

PR #706, opened 2021-04-09
C: pointer alignment check for 4 / 8 byte alignment depending on sizeof(void *)

FreyJo:


mindThomas:
Instead of wasting calls to `acados_get_pointer_size` you should make this function constexpr (if we were using C++) or into a compile directive (#define). In the latter case the check should be added elsewhere.

FreyJo:
I am not familiar with constexpr. But the core of acados should stay pure C.
What do you think is wasted? CPU time for the memory assignment is not really important and the overhead of a C function call is marginal.

PR #711, opened 2021-04-20
Nicer simulink examples 

sandmaennchen:
thanks a lot, @asparc!

.slx files stored in Matlab2013a format for backwards compatibility


FreyJo:
For reference:
https://discourse.acados.org/t/simulink-advanced-closed-loop-example/413

Issue #651, opened 2020-10-16
Linker problem after building acados with multiple qp solvers


asparc:
I'm experiencing linker issues on Windows 10 + Matlab2017b with acados compiled with multiple qp solvers when running the minimal example in `acados_matlab_octave\getting\started\minimal_example_ocp.m`.

More specifically: 
I compiled acados with qpOASES and OSQP, by turning on their flags in the CMakeLists.txt. When I set the default qp solver (`partial_condensing_hpipm`), `minimal_example_ocp.m` fails because MinGW64 complains about undefined references to qpOASES and OSQP, while these solvers shouldn't be used.

```
Building with 'MinGW64 Compiler (C)'.
C:\ProgramData\MATLAB\SupportPackages\R2017b\3P.instrset\mingw_w64.instrset\bin\gcc -c -DUSE_MEX_CMD   -m64 -DMATLAB_MEX_FILE  -I"C:\Users\gruijthuijsenc\Drive\Documents\MATLAB\acados\examples\acados_matlab_octave\..\.." -I"C:\Users\gruijthuijsenc\Drive\Documents\MATLAB\acados\examples\acados_matlab_octave\..\..\interfaces" -I"C:\Users\gruijthuijsenc\Drive\Documents\MATLAB\acados\examples\acados_matlab_octave\..\..\external" -I"C:\Users\gruijthuijsenc\Drive\Documents\MATLAB\acados\examples\acados_matlab_octave\..\..\external\blasfeo\include" -I"C:\Users\gruijthuijsenc\Drive\Documents\MATLAB\acados\examples\acados_matlab_octave\..\..\external\hpipm\include"  -I"C:\Program Files\MATLAB\R2017b/extern/include" -I"C:\Program Files\MATLAB\R2017b/simulink/include" -I"C:\Program Files\MATLAB\R2017b/extern\lib\win64\mingw64" -fexceptions -fno-omit-frame-pointer -std=c99 -fopenmp -O -DNDEBUG "C:\Users\gruijthuijsenc\Drive\Documents\MATLAB\acados\interfaces\acados_matlab_octave\ocp_create.c" -o C:\Users\GRUIJT~1\AppData\Local\Temp\mex_136896563278865_18692\ocp_create.obj
[...]

PR #712, opened 2021-04-21
Improve MEX interface

FreyJo:
This PR is replacing https://github.com/acados/acados/pull/710 and applying the changes carefully, e.g. compatibility with Octave is ensured.

- [x] no recompilation of interface needed when switching QP solvers
   - [x] CMake & Make build system
- [x] addresses issue https://github.com/acados/acados/issues/651 
- [ ] On my Windows system, I tested the solvers from Matlab:
  - [x] HPIPM
  - [x] above in Simulink
  - [x] qpOASES
  - [x] above in Simulink
  - [x] OSQP
  - [x] above in Simulink
  - [ ] qpDUNES; current error: https://pastebin.com/M1FuUy1i
  - [ ] above in Simulink
- [ ] make mex wrapper around templated solver work on Windows: https://github.com/FreyJo/acados/blob/master/examples/acados_matlab_octave/test/test_template_pendulum_exact_hess.m
  - current error: https://pastebin.com/hPCTV6KY
  - probably templated Makefile needs to be adapted
- [x] improved installation documentation with respect to what I tested.

Note, I am not sure, if I will fix all the unchecked points above, since I would like to merge this soon. This is just an overview of what would be desireable.

Big thanks to @asparc :pray:


interfaces/acados_matlab_octave/ocp_compile_interface.m
```diff
@@ -112,82 +121,41 @@ function ocp_compile_interface(opts)
...
-% is HPMPC?
-with_qp_hpmpc = ~isempty(strfind(opts.qp_solver, 'hpmpc'));
-if with_qp_hpmpc
-    % flag file to remember if compiled with HPMPC
-    flag_file = fullfile(opts.output_dir, '_compiled_with_hpmpc.txt');
-    flagID = fopen(flag_file, 'w');
-    fclose(flagID);
-    % additional compilation flag
-    FLAGS = [FLAGS, ' -DACADOS_WITH_HPMPC'];
+if ~is_octave()
```
FreyJo, line 124: @asparc your PR didn't include these flags as far as I see, which is weird because I needed them.

interfaces/acados_matlab_octave/ocp_compile_interface.m
```diff
@@ -112,82 +121,41 @@ function ocp_compile_interface(opts)
     LDFLAGS = 'LDFLAGS=$LDFLAGS -fopenmp';
 end
 
-% remove flag file, if present, before creating a new one
```
asparc, line 115: I think these flag files were used to support the `auto` mode for the option `compile_interface`. Are you sure you want all of this?

interfaces/acados_matlab_octave/ocp_set_ext_fun.m
```diff
@@ -615,13 +616,15 @@
...
+            % load linking information
+            libs = loadjson(fileread(fullfile(acados_folder, 'lib', 'link_libs.json')));
             mex(mex_flags, FLAGS, LDFLAGS, ['-DSETTER=', setter{ii}],...
                 ['-DSET_FIELD=', set_fields{ii}], ['-DMEX_FIELD=', mex_fields{ii}],...
                 ['-DFUN_NAME=', fun_names{ii}], ['-DPHASE=', num2str(phase{ii})],...
                 ['-DN0=', num2str(phase_start{ii})], ['-DN1=', num2str(phase_end{ii})],...
                 acados_include, acados_interfaces_include, external_include, blasfeo_include,...
                 hpipm_include, acados_lib_path, acados_matlab_octave_lib_path, model_lib_path,...
-                '-lacados', '-lhpipm', '-lblasfeo', ['-l', model_name], mex_files{ii});
+                '-lacados', '-lhpipm', '-lblasfeo', libs.osqp, libs.qpoases, libs.hpmpc, ['-l', model_name], mex_files{ii});
```
asparc, line 627: Here you might need to add  `libs.qpdunes` and `libs.ooqp` as well?

interfaces/acados_matlab_octave/ocp_solve.c
```diff
@@ -40,10 +40,19 @@
 // mex
 #include "mex.h"
 
-
+#if defined(_WIN32) || defined(__WIN32__) || defined(WIN32) || defined(_WIN64)
```
asparc, line 43: While all of these changes are necessary to show the verbose output on Windows Matlab, they are annoying when you don't want verbose output, especially when you are running a closed-loop controller: it will open and close a terminal window for each call to `ocp_solve`, thus showing a flickering terminal window during the entire closed loop simulation (which also slows down the computation). If you don't feel like implementing to activate these lines only when the print_level > 0, I'd leave them commented out by default

interfaces/acados_matlab_octave/acados_ocp.m
```diff
@@ -142,28 +142,32 @@
...
+                        end
                     end
                 else
                     compile_interface = true;
                 end
             else
-                obj.model_struct.cost_type
                 error('acados_ocp: field compile_interface is %, supported values are: true, false, auto', ...
                         obj.opts_struct.compile_interface);
             end
```
asparc, line 173: At line 187, this function throws an error when you selected a qp_solver that acados was not compiled with. The error message is not correct. E.g.: 
```
Error using acados_ocp (line 187)
Exception:

    ocp_create: field qp_solver does not support partial_condensing_osqp, supported values are:
partial_condensing_hpipm, full_condensing_hpipm, full_condensing_qpoases, partial_condensing_osqp, partial_condensing_hpmpc,
partial_condensing_qpdunes

```

interfaces/acados_matlab_octave/ocp_compile_interface.m
FreyJo: oops, right..

interfaces/acados_matlab_octave/ocp_compile_interface.m
FreyJo: Yes, now the `auto` mode is done via the `link_libs.json`, thus I removed the flag file stuff here.

interfaces/acados_matlab_octave/ocp_solve.c
FreyJo: I agree! I committed it in uncommented form because I thought it would pipeline it to the "normal" Matlab terminal.

interfaces/acados_matlab_octave/acados_ocp.m
FreyJo: Good point, I added the following note to the error message to address this:
```
NOTE: acados needs to be compiled explicitly with external QP solvers!
If the qp_solver value is listed as supported, please recompile acados with the desired QP solver.
```

interfaces/acados_matlab_octave/ocp_compile_interface.m
asparc: I mean something else. The code used to put a flag file in de build folder, to remember for what qp_solver the ocp mex files were built. In `auto` mode, it read the name of that flag file, to see if it has to rebuild to ocp. At first sight, this flag file is not placed in the build folder anymore. Or can you just change the selected solver for the compiled ocp, without rebuilding the mex files?

FreyJo: exaclty, now the Mex interface is built with all the external QP solvers specified in the `link_libs.json` file. So there is no need to recompile the interface when switching between different external QP solvers.

asparc: I see. Perfect!

PR #713, opened 2021-04-23
Dev generic dyn template

martkir0919:
C-code can be used to specify discrete-time dynamics in Matlab (standard and template interface). An example is included.


FreyJo:
closing, replaced by #717

PR #719, opened 2021-05-06
Wip msvc

asparc:
Enabled MSVC compiler for Matlab interface, including multiple solver support and sfunction compilation.

Compilation (assuming MSVC 2017 compiler, 64 bit)
- cmake:

        cmake -G "Visual Studio 15 2017 Win64" -DBLASFEO_TARGET=GENERIC -DACADOS_INSTALL_DIR=.. -DBUILD_SHARED_LIBS=OFF ..

- build (don't use PowerShell, as that creates some issues with env vars): 

        cmake --build . -j10 --target INSTALL --config Release

- Matlab: change mex setup to MSVC via `mex -setup C`

Remarks:
- untested for 32 bit (I think I'd need a 32-bit Matlab version to compile 32-bit mex files)
- untested for e.g. MSVC 2015 or 2019 compiler. Biggest challenge here is the location of the `.bat` file that sets the env vars that MSVC's `cl.exe` and `link.exe` need. On my system, I can find this `.bat` file as follows (but I see warnings throughout the internet that say that this location might not be fixed across MSVC versions :roll_eyes:):

        mexOpts = mex.getCompilerConfigurations('C', 'Selected');
        msvc_env = fullfile(mexOpts.Location, 'VC\Auxiliary\Build\vcvars64.bat');

- it seems that MSVC (as opposed to mingw/gcc) does not require the `-openmp` (not `-fopenmp`) linker flag, although I did add it everywhere.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -281,7 +281,7 @@ message(STATUS " ")
 
 # Write libraries to be linked into a json file
 if(${ACADOS_WITH_OPENMP})
-    set(LINK_FLAG_OPENMP -fopenmp)
+    set(LINK_FLAG_OPENMP ${OpenMP_C_FLAGS})
 endif()
 if(${ACADOS_WITH_QPOASES})
     set(LINK_FLAG_QPOASES -lqpOASES_e)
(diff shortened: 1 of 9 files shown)


FreyJo:
Closing, replaced by https://github.com/acados/acados/pull/721

PR #724, opened 2021-05-20
Solver parameter value initialization from Matlab

ivobatkovic:
A way to pass the parameter values to the generated c code from the Matlab interface.

- Included a parameter_values field for acados_opts
- Store the parameter values in the json file
- Updated template files for acados_solver and main

### DIFF
--- a/interfaces/acados_matlab_octave/acados_ocp_opts.m
+++ b/interfaces/acados_matlab_octave/acados_ocp_opts.m
@@ -55,6 +55,7 @@ classdef acados_ocp_opts < handle
             % set one of the following for nonuniform grid
             obj.opts_struct.shooting_nodes = [];
             obj.opts_struct.time_steps = [];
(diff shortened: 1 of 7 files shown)


FreyJo:
Hi @ivobatkovic 
thanks for the PR.
However, I would not like to have the default parameters to be varying over the stages for now.
Setting the parameters stage wise can be done by interacting with the solver after creation.
I think otherwise, it is not clear which parameter values are initially used when creating an integrator from an ocp nlp solver.
Thus, I would apply these changes instead:
https://github.com/acados/acados/pull/725

ivobatkovic:
Hi @FreyJo, thanks for the quick reply.

I understand your reasoning. The idea behind setting parameters for each stage was to quickly verify that the generated C code behaves in the same way as the Matlab version, e.g., you won't need to modify the generated main file.

I'll close this PR.

PR #730, opened 2021-06-17
acados_template: make tera path configurable

adeebshihadeh:
Since `ACADOS_SOURCE_DIR` is current only used to find the tera path, we could remove it, but not sure if it's wanted for future use.

### DIFF
--- a/interfaces/acados_template/acados_template/utils.py
+++ b/interfaces/acados_template/acados_template/utils.py
@@ -55,8 +55,10 @@ def get_acados_path():
 
 
 def get_tera_exec_path():
-    ACADOS_PATH = get_acados_path()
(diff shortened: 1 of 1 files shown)


FreyJo:
`ACADOS_SOURCE_DIR` is not only used to find the tera path.
See, e.g.
https://github.com/acados/acados/blob/0334f7c7f67a52ee511037fa691552c1805493ea/interfaces/acados_template/acados_template/acados_ocp.py#L2775

adeebshihadeh:
Ah, missed that. Anything else needed to get this merged?

FreyJo:
Done, I guess this can be useful, if you have to compile tera renderer, when on a system that is not compliant with the default binaries as described here https://discourse.acados.org/t/problems-with-t-renderer/438

PR #732, opened 2021-07-16
fix warning strict-prototypes

krepa098:
Hello,

```void foo(void)``` and ```void foo()``` are not the same in plain C. This causes the compile to emit warnings resp. errors depending on the context (e.g. ```-Werror -Wstrict-prototypes```). This PR addresses the issue.

### DIFF
--- a/interfaces/acados_template/acados_template/c_templates_tera/acados_sim_solver.in.h
+++ b/interfaces/acados_template/acados_template/c_templates_tera/acados_sim_solver.in.h
@@ -88,7 +88,7 @@ sim_opts * {{ model.name }}_acados_get_sim_opts(sim_solver_capsule *capsule);
 sim_solver * {{ model.name }}_acados_get_sim_solver(sim_solver_capsule *capsule);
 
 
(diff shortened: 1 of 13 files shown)


FreyJo:
Hi,
that's interesting.
There are quite a few similar function declarations in the acados core though.
So, compiling with `-Werror -Wstrict-prototypes` will also not work after merging this.
Are you planning to change these too?

krepa098:
There is no real incentive in compiling the acados core library with those flags, at least in my case. I guess it would make sense to be stricter about the generated code that gets included in all kinds of projects (using all kinds of flags), unlike the code of lib acados itself. I guess it would be the right thing to do this for the core library as well, but in practice there is very little to be gained here.

So, no I'm currently not planning to do those changes.

FreyJo:
Okay, thanks!
I rebased the commit here: https://github.com/acados/acados/pull/736/commits/b54276eef41fb19206626062bce2cb396c88209d
There was another external function added before this was merged (https://github.com/acados/acados/pull/734)

PR #735, opened 2021-07-20
Fast get set

HaraldSchafer:


HaraldSchafer:
@FreyJo Is this something you'd be interested in merging? It just makes running with the python api a lot faster.

FreyJo:
Hi @HaraldSchafer ,

I really like the `build` option for the Python interface and would like to merge that.

About the `*_slice` stuff, I am not so sure.
In the `acados` core, it is possible that the dimensions vary over the shooting nodes.
That's why it is not enough to get the dimension of a field only at one shooting node and then assume this dimension will be constant for all subsequent nodes.
This is for example happening here:
https://github.com/acados/acados/pull/735/files#diff-c4d1815bb1d3e25ee5d33f0159d3a6ccfe2b93cc38915b1c6d458e6a22b3d591R911

Also, this is just a computational overhead for prototyping the OCP solver in Python.
In time critical applications one would always directly use the `C` interface of acados.

I thus suggest to just cherry-pick this commit:
https://github.com/acados/acados/pull/735/commits/2d6d9dcb0811bf192fe659ba252e137e6364d33f

Is that fine with you?

HaraldSchafer:
Oh I forgot about the build stuff, made a new PR for that.

https://github.com/acados/acados/pull/737

HaraldSchafer:
But yeah I was mostly asking if you are interested in merging a fast python interface. For our purposes in openpilot we want a fast python interface. The current one is very slow, but there is a lot of room for improvement.

If you are not interested, we will just write a minimal fast python interface for our purposes.

FreyJo:
It would definitely be nice to have a faster Python interface that can be used in openpilot!
Just this dimension stuff has to be thought of a bit more carefully.
I suggest you fix that and the unit tests then ping me again, then I review a bit more in depth.
Also let me know if you have problems doing that.

> The setting and getting slices was a start, I can fix the dimension stuff too.

Do you have something else in mind to speed up the Python interface?
There actually is not much going on besides getting and setting after solver creation.

HaraldSchafer:
Okay I'll get it working correctly.

I assume all the dimension checking also slows things down. That could all be statically done? I'll work on it more and let you know what I find.

FreyJo:
> I assume all the dimension checking also slows things down. That could all be statically done? I'll work on it more and let you know what I find.

I am not exactly sure how you are planning to handle this.
But that sounds good, keep me posted! :ok_hand:

FreyJo:
Closing, work on a fast Python interface is continued using the Cython approach
https://github.com/acados/acados/pull/766

PR #744, opened 2021-08-12
Defines in a header file such that they are known at compile time

roangel:
If I try to add the `c_generated_code` folder from a python project to a C++ project, the dimensions of the problem are not known at compile time, which was the case with ACADO (`acado_common.h`). The only way was to include the `acados_solver_<project>.c`, which is rather ugly. This tries to fix that, but not sure if this is the preferred way. Feel free to suggest changes!


FreyJo:
Hi @roangel 
Thanks for the PR:
I suggest to merge this commit instead, which would allow to use multiple acados OCP solvers (e.g. for MPC + MHE) together.
https://github.com/acados/acados/pull/745

Does that work for you?

FreyJo:
merged #745 instead.

PR #745, opened 2021-08-23
Templates: define ocp solver dimensions in header file such that they…

FreyJo:
… are known at compile time

Compare https://github.com/acados/acados/pull/744

I suggest to merge this commit instead:
The name `ACADOS_N*` could cause problems, if one tries to use multiple acados OCP solvers together.


roangel:
Hi @FreyJo , thanks for the suggestion, that works nicely for me as well!

PR #747, opened 2021-08-29
acados_template improvements

schwieni:
This PR adds some improvements to the exported code

1. help to identify constant time-steps with numerical differences (improves reliability when distinguishing code paths in template)
2. ocp & main: help reuse exported code, since the model is independent of the number of shooting nodes
3. sim & main: use dimension defines and help with code readability (use parameters instead of numbers in code)
4. example minimal_example_ocp_reuse_code.py added to ctest

### DIFF
--- a/interfaces/acados_template/acados_template/acados_ocp_solver.py
+++ b/interfaces/acados_template/acados_template/acados_ocp_solver.py
@@ -831,6 +831,10 @@ class AcadosOcpSolver:
         assert getattr(self.shared_lib, f"{model.name}_acados_create")(self.capsule)==0
         self.solver_created = True
 
+        # create solver with another number of stages
+        getattr(self.shared_lib, f"{model.name}_acados_create_w_stages").argtypes = [c_void_p, c_int, c_void_p]
+        getattr(self.shared_lib, f"{model.name}_acados_create_w_stages").restype = c_int
+
         # get pointers solver
         getattr(self.shared_lib, f"{model.name}_acados_get_nlp_opts").argtypes = [c_void_p]
         getattr(self.shared_lib, f"{model.name}_acados_get_nlp_opts").restype = c_void_p
--- a/interfaces/acados_template/acados_template/acados_ocp_solver.py
+++ b/interfaces/acados_template/acados_template/acados_ocp_solver.py
@@ -783,7 +783,9 @@ class AcadosOcpSolver:
         if simulink_opts == None:
             acados_path = get_acados_path()
             json_path = os.path.join(acados_path, 'interfaces/acados_template/acados_template')
-            with open(json_path + '/simulink_default_opts.json', 'r') as f:
+            json_filename = json_path + '/simulink_default_opts.json'
+            print(f'using simulink json-file: "{json_filename}"')
+            with open(json_filename, 'r') as f:
                 simulink_opts = json.load(f)
 
         # make dims consistent
@@ -832,7 +834,17 @@ class AcadosOcpSolver:
         assert getattr(self.shared_lib, f"{model.name}_acados_create")(self.capsule)==0
         self.solver_created = True
 
+        self.acados_ocp = acados_ocp
+
         # get pointers solver
+        self.__get_pointers_solver()
+
+    def __get_pointers_solver(self):
+        """
+        Private function to get the pointers for solver
+        """
+        # get pointers solver
+        model = self.acados_ocp.model
         getattr(self.shared_lib, f"{model.name}_acados_get_nlp_opts").argtypes = [c_void_p]
         getattr(self.shared_lib, f"{model.name}_acados_get_nlp_opts").restype = c_void_p
         self.nlp_opts = getattr(self.shared_lib, f"{model.name}_acados_get_nlp_opts")(self.capsule)
@@ -857,9 +869,6 @@ class AcadosOcpSolver:
         getattr(self.shared_lib, f"{model.name}_acados_get_nlp_solver").restype = c_void_p
         self.nlp_solver = getattr(self.shared_lib, f"{model.name}_acados_get_nlp_solver")(self.capsule)
 
-        self.acados_ocp = acados_ocp
-
-
     def solve(self):
         """
         Solve the ocp with current input.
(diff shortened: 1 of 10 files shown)


interfaces/acados_template/acados_template/c_templates_tera/acados_solver.in.h
```diff
@@ -144,6 +144,12 @@ nlp_solver_capsule * {{ model.name }}_acados_create_capsule(void);
 int {{ model.name }}_acados_free_capsule(nlp_solver_capsule *capsule);
 
 int {{ model.name }}_acados_create(nlp_solver_capsule * capsule);
+/**
+ * More generic acados_create function, since the number of stages usually is independent of the model and its
```
FreyJo, line 148: I suggest changing the comment to:
"Generic version of {{ model.name }}_acados_create which allows to use a different number of shooting intervals than the number used for code generation."

interfaces/acados_template/acados_template/acados_ocp_solver.py
```diff
@@ -825,6 +831,10 @@ def __init__(self, acados_ocp, json_file='acados_ocp_nlp.json', simulink_opts=No
         assert getattr(self.shared_lib, f"{model.name}_acados_create")(self.capsule)==0
         self.solver_created = True
 
+        # create solver with another number of stages
```
FreyJo, line 834: I think these lines are not needed, since the solver in Python is still created via
the line above:
```
         assert getattr(self.shared_lib, f"{model.name}_acados_create")(self.capsule)==0
```

interfaces/acados_template/acados_template/c_templates_tera/acados_solver.in.c
```diff
@@ -122,8 +122,23 @@ int {{ model.name }}_acados_free_capsule(nlp_solver_capsule *capsule)
 
 
 int {{ model.name }}_acados_create(nlp_solver_capsule * capsule)
+{
+    int n_stages = {{ model.name | upper }}_N;
+    double* const_time_step = NULL; // NULL -> don't alter the exported value
+    return {{ model.name }}_acados_create_w_stages(capsule, n_stages, const_time_step);
+}
```
FreyJo, line 129: I suggest to change the lines above to:
```
    int N_shooting_intervals = {{ model.name | upper }}_N;
    double* const_time_step = NULL; // NULL -> use time step from code generation
    return {{ model.name }}_acados_create_with_discretization(capsule, N_shooting_intervals, const_time_step); 
```
I think the function name: `{{ model.name }}_acados_create_with_discretization` would better describe what it does.

FreyJo (changes requested):
Hi @schwieni ,

thanks for the PR!
I spotted one error and would like to rather use the term shooting intervals than stages. Just because number of stages is also used in the integrator settings.
In addition to the comments, I think it would be nicer to have an argument 
```
 int {{ model.name }}_acados_create_w_stages(nlp_solver_capsule * capsule, int N, double* time_steps)
```
where `time_steps` is a vector of time steps of length `N`.
What do you think?
Could you implement that and address the other comments?

Best,
Jonathan

schwieni:
Hi @FreyJo,
thank you for looking into this PR!
I will integrate your suggestions and make a commit.
Sorry about the (N+1)! Good that you are having a good eye on it. :-)

interfaces/acados_template/acados_template/c_templates_tera/main.in.c
```diff
@@ -42,12 +42,45 @@
...
+#define NY     {{ model.name | upper }}_NY
+#define NYN    {{ model.name | upper }}_NYN
+#define NH     {{ model.name | upper }}_NH
+#define NPHI   {{ model.name | upper }}_NPHI
+#define NHN    {{ model.name | upper }}_NHN
+#define NPHIN  {{ model.name | upper }}_NPHIN
+#define NR     {{ model.name | upper }}_NR
+
+
+int main(int argc, char** argv)
```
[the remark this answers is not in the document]
schwieni: You are right. In a previous conversion we had, I understood that this file is considered a starting point for a c-based solver. So I was under the impression, that this would make sense. I will remove the signature again.

schwieni:
Hi @FreyJo,
I made the changes you requested above. Are you ok with the changes?
If you want I can also try to add an example in python also to test the feature (change N after compilation).
This could also be integrated in the ctests.

interfaces/acados_template/acados_template/c_templates_tera/main_sim.in.c
```diff
@@ -41,6 +41,36 @@
...
+#define NGN    {{ model.name | upper }}_NGN
+#define NY0    {{ model.name | upper }}_NY0
+#define NY     {{ model.name | upper }}_NY
+#define NYN    {{ model.name | upper }}_NYN
+#define NH     {{ model.name | upper }}_NH
+#define NPHI   {{ model.name | upper }}_NPHI
+#define NHN    {{ model.name | upper }}_NHN
+#define NPHIN  {{ model.name | upper }}_NPHIN
+#define NR     {{ model.name | upper }}_NR
+
```
FreyJo, line 73: I would remove all dimensions but NX, NZ, NU and NP here, since this file is used to just test the standalone acados integrator.

FreyJo (approved):

FreyJo:
Thanks for addressing everything!
If you have such an example available and can add it to the `ctests` that would be nice.
I will wait for it and then merge this :+1:

FreyJo:
>  If you want I can also try to add an example in python also to test the feature (change N after compilation).

Do you still want to do this?
I can also merge it as is.

schwieni:
> > If you want I can also try to add an example in python also to test the feature (change N after compilation).
> 
> Do you still want to do this?
> I can also merge it as is.

Hi @FreyJo,
I added an example (minimal_example_ocp_reuse_code.py). I had to change some interfaces as well. I also added it to the list of ctest. See my last two commits.
I think now with the example and test the whole PR makes more sense. :-)
Can you review again, please?

FreyJo:
I would just "squash and merge" as-is, to keep the acados history linear.
Is that fine with you?

schwieni:
Hi @FreyJo,
> I really like the example.
cool :-)

> I would just "squash and merge" as-is, to keep the acados history linear.
> Is that fine with you?
Of course! Thanks for merging! :+1:

PR #766, opened 2021-11-12
cython interface

jwooning:
This is an alternative approach to a fast python interface, instead of using slices (#735) it uses cython. 

A cython interface is created in `acados_ocp_solver_pyx.pyx` with the same API as `acados_ocp_solver.py`. 
Not using ctypes, and compiling the code, this is much faster than the original python interface, and it should be just as safe.

@FreyJo what do you think about this approach?

### DIFF
--- a/examples/acados_python/cython/example.py
+++ b/examples/acados_python/cython/example.py
@@ -1,0 +1,218 @@
+#
+# Copyright 2019 Gianluca Frison, Dimitris Kouzoupis, Robin Verschueren,
+# Andrea Zanelli, Niels van Duijkeren, Jonathan Frey, Tommaso Sartor,
+# Branimir Novoselnik, Rien Quirynen, Rezart Qelibari, Dang Doan,
+# Jonas Koenemann, Yutao Chen, Tobias Schöls, Jonas Schlagenhauf, Moritz Diehl
+#
+# This file is part of acados.
+#
+# The 2-Clause BSD License
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+# 1. Redistributions of source code must retain the above copyright notice,
+# this list of conditions and the following disclaimer.
+#
+# 2. Redistributions in binary form must reproduce the above copyright notice,
+# this list of conditions and the following disclaimer in the documentation
+# and/or other materials provided with the distribution.
+#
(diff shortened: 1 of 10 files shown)


FreyJo:
I was not really aware of cython, but I really like that approach!

I didn't check the changes closely yet.
Could you include a working example with the new interface and ping me again?
Then I can go through the workflow a bit more carefully.

I think this should then fully replace the existing ctypes based interface, so only one Python interface has to be maintained.
- obviously some functionalities need to be added in the cython based interface
- the documentation generation should work for the cython interface equally well
- all examples should still work

A minor remark: I saw that you are using `clang` in the new targets of the templated Makefile. Was there a specific reason for that?
I think it should be the same compiler for all the targets there, but which one can of course be made an option.

jwooning:
@FreyJo alright, included a small example in `examples/acados_python/cython`, it's just the same as the `crane` example but it uses the cython interface.

If you think this workflow is ok, then I will fix up the non implemented functions, to fully replace the ctypes interface. 

The documentation generation will probably require a very minimal example that can be imported for the generation to work, but that also shouldn't be a problem.

FreyJo:
I tried the example but had to do some changes to make it work on my machine, which should be done in a more clean fashion I guess.
See here:
https://github.com/commaai/acados/compare/cython2...FreyJo:wip_cython?expand=1

I think it would be consistent if rendered files include the model name https://github.com/acados/acados/blob/34eb6db2780f07b2291a2a2ae831a97ed8a2890f/interfaces/acados_template/acados_template/acados_ocp_solver.py#L657-L659

This import would then have to be changed. I actually find it a bit ambiguous now.
https://github.com/acados/acados/blob/34eb6db2780f07b2291a2a2ae831a97ed8a2890f/interfaces/acados_template/acados_template/acados_ocp_solver_pyx.pyx#L42

This is related to supporting multiple different acados OCP solvers in one script.
I think it makes sense to think about that next and adapt this example to the other interface:
https://github.com/acados/acados/blob/master/examples/acados_python/getting_started/mhe/closed_loop_mhe_ocp.py
One option would also be to adapt it and generate the code corresponding to the different solvers in separate solvers (which is also what is done in openpilot I guess).

FreyJo:
Just looking through the `NotImplementedError`s, the functions to implement in the Cython interface:
- [ ] set_new_time_steps
- [ ] print_statistics -> can just call the function in the generated solver
- [ ] store_iterate
- [ ] load_iterate
- [ ] get_stats
- [ ] get_residuals

PR #772, opened 2021-12-09
Work on Globalization

FreyJo:
- add option to check Armijo condition
- simplifed SQP loop
- add option to do Second Order Correction (SOC) -- following Nocedal2006: p.557, eq. (18.51) -- (18.56)
- SOC is done once if step is rejected or constraint violation would increase when using non-corrected step, as described in IPOPT paper (Waechter2006)

replacing https://github.com/acados/acados/pull/764

Example [`examples/acados_python/non_ocp_nlp/marathos_test_problem.py`](https://github.com/acados/acados/blob/4c75f8885e58a15406464a1bd42558804a6d8bd0/examples/acados_python/non_ocp_nlp/marathos_test_problem.py) is used to test the new globalization.

Plots produced with this example:
- line search without SOC:
marathos_no_SOC
- line search with SOC:
marathos_with_SOC


acados/utils/print.c
```diff
@@ -402,6 +402,39 @@ void print_ocp_qp_out(ocp_qp_out *qp_out)
     return;
 }
 
+
+void print_ocp_qp_out_to_file(FILE *file, ocp_qp_out *qp_out)
```
giaf, line 406: I would move the actual printing routine for an ocp_qp_sol in HPIPM and here just call that one, since the actual QP solution struct is defined in HPIPM and here there is only a typedef.
So the printing routine would remain up to date in case there are changes in HPIPM.

acados/utils/print.c
```diff
@@ -533,6 +566,108 @@ void print_ocp_qp_res(ocp_qp_res *qp_res)
...
+            fprintf(file, "%d ", *(A+i+j*lda));
+        }
+        fprintf(file, "\n");
+    }
+    fprintf(file, "\n");
+}
+
+
+
+void print_ocp_qp_in_to_file(FILE *file, ocp_qp_in *qp_in)
```
giaf, line 586: Same as for the QP sol.

acados/utils/print.h
```diff
@@ -75,8 +75,12 @@ void print_ocp_qp_dims(ocp_qp_dims *dims);
 
 void print_ocp_qp_in(ocp_qp_in *qp_in);
```
giaf, line 76: In general all printing routines that are printing HPIPM structures I would move them in HPIPM.
Maybe it's unrelated to this PR but it would make sense to do at some point.

acados/ocp_nlp/ocp_nlp_sqp_rti.c
```diff
@@ -516,10 +511,6 @@ void ocp_nlp_sqp_rti_preparation_step(void *config_, void *dims_,
     mem->time_lin = 0.0;
     mem->time_reg = 0.0;
 
-    int N = dims->N;
-
-    int ii;
-
 #if defined(ACADOS_WITH_OPENMP)
```
giaf, line 514: better move also the beginning and end of parallel region in the new routine to alias memory, otherwise there is the future risk of calling it from somewhere else without them

giaf: To be more precise, the issue is that the `nowait` in the `omp for` means that there is not an explicit barrier after the parallelized for loops, and the implicit barrier is at the end of the parallel region.
So either the beginning and end of parallel regions are also moved inside the new routine, or at least an explicit barrier is added at the end of the new routine.

acados/ocp_nlp/ocp_nlp_common.c
```diff
@@ -2115,11 +2242,208 @@ void ocp_nlp_embed_initial_value(ocp_nlp_config *config, ocp_nlp_dims *dims,
...
+
+    /* dynamics */
+    double merit_grad_dyn = 0.0;
+    for (i=0; i<N; i++)
+    {
+        // get shooting node gap x_next(x_n, u_n) - x_{n+1};
+        tmp_vec = config->dynamics[i]->memory_get_fun_ptr(mem->dynamics[i]);
+
+        /* compute directional derivative of xnext with direction y -> tmp_vec_nxu */
+        blasfeo_dvecse(nx[i], 0.0, &dxnext_dy, 0);
```
giaf, line 2283: I don't really understand these two loops.
Aren't you just trying to compute what dgemv_t would do at once?
And in any case, the index of the matrix BAbt is missing (i.e. `mem->qp_in->BAbt+i`), and the size of the operation is wrong, since it should be the next value of nx (i.e. `nx[i+1]`), at least if I got it correctly.

acados/ocp_nlp/ocp_nlp_common.c
```diff
@@ -2115,11 +2242,208 @@ void ocp_nlp_embed_initial_value(ocp_nlp_config *config, ocp_nlp_dims *dims,
...
+        }
+        for (j = nu[i]; j < nu[i] + nx[i]; j++)
+        {
+            double alpha = BLASFEO_DVECEL(mem->qp_out->ux+i, j);
+            // Sx
+            blasfeo_drowex(nx[i], 1.0, mem->qp_in->BAbt, j, 0, &tmp_vec_nxu, 0);
+            blasfeo_daxpy(nx[i], alpha, &tmp_vec_nxu, 0, &dxnext_dy, 0, &dxnext_dy, 0);
+        }
+        /* add merit gradient contributions depending on sign of shooting gap */
+        for (j = 0; j < nx[i]; j++)
```
giaf, line 2299: also here it should be `nx[i+1]` as this is the (state) size of `pi+i` and `ux+i+1`

acados/ocp_nlp/ocp_nlp_common.c
```diff
@@ -2115,11 +2242,208 @@ void ocp_nlp_embed_initial_value(ocp_nlp_config *config, ocp_nlp_dims *dims,
...
+    {
+        tmp_vec = config->constraints[i]->memory_get_fun_ptr(mem->constraints[i]);
+        int *idxb = mem->qp_in->idxb[i];
+        if (ni[i] > 0)
+        {
+            // printf("constraints fun = \n");
+            // blasfeo_print_exp_dvec(2*ni[i], tmp_vec, 0);
+            // print_ocp_qp_in(mem->qp_in);
+            // for (j = 0; j < 2 * ni[i]; j++) // 2 * ni
+            for (j = 0; j < 2 * (nb[i] + ng[i]); j++) // 2 * ni
```
giaf, line 2339: Since everything is two sided, I would rather loop with j from 0 to nb[i]+ng[i] and process the upper and the lower bounds at once merged in the same loop.
This would simplify indexing and give extra independent work for each loop iteration.

acados/ocp_nlp/ocp_nlp_sqp.c
```diff
@@ -785,76 +701,271 @@ int ocp_nlp_sqp(void *config_, void *dims_, void *nlp_in_, void *nlp_out_,
...
+            // Do SOC
+            // 1.if "the first trial step size alpha_k,0 has been rejected and
+            // 2. if the infeasibility would have increased when accepting the previous step
+
+            // preliminary line search
+            alpha = ocp_nlp_line_search(config, dims, nlp_in, nlp_out, nlp_opts, nlp_mem, nlp_work, 1);
+
+            if (alpha < 1.0)
+            {
+                // Second Order Correction (SOC): following Nocedal2006: p.557, eq. (18.51) -- (18.56)
```
giaf, line 732: It looks like the page and equations are different in my version of the book, can you maybe write the chapter number and title too?

acados/ocp_nlp/ocp_nlp_sqp.c
```diff
@@ -785,76 +701,271 @@ int ocp_nlp_sqp(void *config_, void *dims_, void *nlp_in_, void *nlp_out_,
...
+                    tmp_fun_vec = config->constraints[ii]->memory_get_fun_ptr(nlp_mem->constraints[ii]);
+                    /* SOC for bounds can be skipped (because linear) */
+                    // NOTE: SOC can also be skipped for truely linear constraint, i.e. ng of nlp, now using ng of QP = (nh+ng)
+
+                    // upper & lower
+                    blasfeo_dveccp(ng[ii], tmp_fun_vec, nb[ii], qp_in->d+ii, nb[ii]); // lg
+                    blasfeo_dveccp(ng[ii], tmp_fun_vec, 2*nb[ii]+ng[ii], qp_in->d+ii, 2*nb[ii]+ng[ii]); // ug
+                    // general linear / linearized!
+                    // d[nb:nb+ng] += D * u + C * x (lower)
+                    blasfeo_dgemv_t(nu[ii]+nx[ii], ng[ii], 1.0, qp_in->DCt+ii, 0, 0, qp_out->ux+ii, 0,
```
giaf, line 805: Since the constraints are two sided, you only need to perform the matrix-vector multiplication (the most expensive operation) once and cache it in a working vector, to reuse for both upper and lower constraint.

acados/ocp_nlp/ocp_nlp_sqp_rti.c
giaf: Same for ocp_nlp_sqp.c.

acados/ocp_nlp/ocp_nlp_common.c
```diff
@@ -1106,6 +1107,9 @@ void ocp_nlp_opts_initialize_default(void *config_, void *dims_, void *opts_)
     // globalization
     opts->alpha_min = 0.05;
     opts->alpha_reduction = 0.7;
+    opts->full_step_dual = 0;
+    opts->line_search_use_sufficient_descent = 0;
```
giaf, line 1111: Can you explain somewhere what this option does?

acados/ocp_nlp/ocp_nlp_common.c
```diff
@@ -1106,6 +1107,9 @@ void ocp_nlp_opts_initialize_default(void *config_, void *dims_, void *opts_)
...
+    opts->glob_SOC = 0;
```
giaf, line 1112: Same for this: SOC is second order correction, but what is `glob`?

giaf:
Ok I checked the PR, but it is quite large and now I'm not familiar with the algorithm you implemented, so I can't really judge their correctness.
On the other hand, I tried to check more possible implementation issues, and I found some regarding indexing and parallelization: please double check if there could be other similar ones still out there, at the end you know better what changes you did.

acados/ocp_nlp/ocp_nlp_common.c
FreyJo: You're right!
It is fixed in https://github.com/acados/acados/pull/772/commits/880ab5790b574d9f177c64a45e241b22cb804c01

acados/ocp_nlp/ocp_nlp_common.c
FreyJo: Indeed I found some more related errors.
Fixed in https://github.com/acados/acados/pull/772/commits/61cd5f7a7ff691da282dab3364038e269d480ee6

acados/ocp_nlp/ocp_nlp_common.c
FreyJo: I think here, there is no performance difference, since this is just done for either upper or lower constraints or none of them.
So, I just left a note https://github.com/acados/acados/pull/772/commits/c1f10500125b8bdd6f86a44c745a22957e58cae2

acados/ocp_nlp/ocp_nlp_sqp.c
FreyJo: Right!
Addressed in: https://github.com/acados/acados/pull/772/commits/39e7718c14bcbb204ae0ae0b6e9c5c1fd799fbc7

acados/ocp_nlp/ocp_nlp_common.c
FreyJo: I added and documented the options to the Python interface.
https://github.com/acados/acados/blob/8d2766b4ebaa321e55e10ad8b818bf0dc991d6f7/interfaces/acados_template/acados_template/acados_ocp.py#L2374-L2409
The docstrings there are used to generate the documentation on the website.

acados/ocp_nlp/ocp_nlp_common.c
FreyJo: I renamed the option to `globalization_use_SOC`.
I guess that is more clear.

acados/utils/print.c
FreyJo: I decided to just leave a note, stating that these routines should be moved to HPIPM, for now.

acados/utils/print.h
FreyJo: I decided to just leave a note, stating that these routines should be moved to HPIPM, for now.

acados/ocp_nlp/ocp_nlp_sqp_rti.c
FreyJo: I addressed this in https://github.com/acados/acados/pull/772/commits/d66ce703d708eaaa88b8943b11844a987b1c272d
Playing a bit with openmp, I found an issue, but this is definitely not introduced in the current PR:  https://github.com/acados/acados/issues/782

FreyJo:
Thanks a lot for checking!
I addressed all of your comments, interfaced the new options.
I double checked the indexing in the new code and found some bugs similar to the ones you pointed to.
I think that it can be merged as it is now.

PR #780, opened 2021-12-24
Fix build=false when creating AcadosOcpSolver

gijskoning:
The export directory was not set when not building. This commits fixes it.


FreyJo:
I fixed this in https://github.com/acados/acados/pull/791
It is now also tested and the solver creation process is split into: generate, build and create

PR #781, opened 2022-01-04
templates: simplify Makefile.in with implicit rules and patterns

schwieni:
Some useful patterns and implicit rules from
 http://www.gnu.org/software/make/manual/make.html
are applied here.
The explicit definition of all objects files is replaced by make's implicit rules.
This makes the resulting Makefile also more human readable.

### DIFF
--- a/interfaces/acados_template/acados_template/c_templates_tera/Makefile.in
+++ b/interfaces/acados_template/acados_template/c_templates_tera/Makefile.in
@@ -125,20 +125,6 @@
 	{%- endif %}
 {%- endif %}
 
-{# acados flags #}
-ACADOS_FLAGS = -fPIC -std=c99 {{ openmp_flag }} #-fno-diagnostics-show-line-numbers -g
-{%- if qp_solver == "FULL_CONDENSING_QPOASES" %}
-ACADOS_FLAGS += -DACADOS_WITH_QPOASES
-{%- endif %}
-{%- if qp_solver == "PARTIAL_CONDENSING_OSQP" %}
-ACADOS_FLAGS += -DACADOS_WITH_OSQP
-{%- endif %}
-{%- if qp_solver == "PARTIAL_CONDENSING_QPDUNES" %}
-ACADOS_FLAGS += -DACADOS_WITH_QPDUNES
-{%- endif %}
-# # Debugging
-# ACADOS_FLAGS += -g3
-
 # define sources and use make's implicit rules to generate object files (*.o)
 
 # model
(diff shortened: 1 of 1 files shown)


FreyJo:
Thanks for the PR!
The separation into `MODEL_OBJ` and `OCP_OBJ` makes sense. 
`MODEL_*` being used for sim example / solver, while `OCP_*` is additionally needed for the OCP targets, the more granular separation that was there before seems indeed redundant.
Also there is obviously less repetition now.

I am not sure, how far the current templated Makefile is Windows compatible and how it compares to the new one.
Do you know something about it?

I read a bit about `CFLAGS` and `CPPFLAGS`
https://stackoverflow.com/questions/2754966/cflags-vs-cppflags
I am not sure if what currently is in `ACADOS_FLAGS` is separated correctly into these two.

schwieni:
Thanks for looking into this PR!

In my opinion the current Makefile is unnecessarily complex. Example: a developer (mostly you?) has to add both .c and .o files manually in the template. In case of the CasADi files they have to be added once with and once without directory name.
I wanted to address this here.

### Regarding your Comments
1. > I am not sure, how far the current templated Makefile is Windows compatible and how it compares to the new one.
Do you know something about it?

I did not check the Windows compatibility of this PR. I guess it should work equally to the current version in master.
I should have access to a Visual Studio installation to check it if you want.
Did you or any other user encounter issues with the Makefile in the past?

Another option is to use a Makefile generator, e.g., cmake. I can try to create a template for a CMakeList.txt file if you want. Maybe in another PR? However, when working under Linux with Python or Matlab the direct creation and "execution" of a Makefile seems to be the most efficient option.
What is your opinion?

2. > I read a bit about `CFLAGS` and `CPPFLAGS` https://stackoverflow.com/questions/2754966/cflags-vs-cppflags I am not sure if what currently is in `ACADOS_FLAGS` is separated correctly into these two.

You are correct. However, since the implicit rule is `$(CC) $(CPPFLAGS) $(CFLAGS) -c` (See: http://www.gnu.org/software/make/manual/make.html#Catalogue-of-Rules) it is equivalent to the current version.
My suggestion: remove `ACADOS_FLAGS` and fill `CFLAGS` and `CPPFLAGS` directly. I will do another push for that.

FreyJo:
>  I guess it only works with a Linux-emulation (cygwin, mingw, Linux-subsystem for Windows).

I guess you are right! Thanks for trying though!
I actually don't really remember using the templated Makefile on Windows.
The way I used the template based solvers on Windows in Matlab is to do it through Simulink
https://github.com/acados/acados/blob/master/interfaces/acados_template/acados_template/c_templates_tera/make_sfun.in.m
This is actually also what most Matlab/Windows users want to do in my experience.

> Another option is to use a Makefile generator, e.g., cmake. I can try to create a template for a CMakeList.txt file if you want. Maybe in another PR? However, when working under Linux with Python or Matlab the direct creation and "execution" of a Makefile seems to be the most efficient option.
What is your opinion?

I agree. A CMakeLists file would be the way to go for good cross OS support.
On Linux, using the Make workflow for the template based solver is preferable.

Thanks for the answer and PR!
Merging this now!

PR #783, opened 2022-01-13
Templates: simplify acados_solver.in.c and code reuse

schwieni:
The goal:
1) Split the long function *_acados_create_with_dicretization() into several (numbered) sub-functions to give it some structure.
2) That also facilitates to set "qp_cond_N" afterwards using new function {{ model.name }}_acados_update_qp_solver_cond_N(...). This also allows for better code reuse after code export.
3) Macro MAP_CASADI_FNC to simplify the CasADi function mapping.


FreyJo:
No complaints, good job! :)

Issue #777, opened 2021-12-22
[Errno 9] in Python when installing acados with qpDUNES


FreyJo:
I had weird errors using the test files added in PR #772

I install the acados shared library with CMake on Ubuntu 20.04.
When I install acados without qpDUNES the newly added example / test files, work as expected.
When installing with qpDUNES, I get the following error.
```
  File "marathos_test_problem.py", line 243, in <module>
    main()
  File "marathos_test_problem.py", line 59, in main
    solve_marathos_problem_with_setting(setting)
  File "marathos_test_problem.py", line 171, in solve_marathos_problem_with_setting
    ocp_solver.print_statistics()
  File "/home/oj/acados/interfaces/acados_template/acados_template/acados_ocp_solver.py", line 1052, in print_statistics
    print('\niter\tres_stat\tres_eq\t\tres_ineq\tres_comp\tqp_stat\tqp_iter\talpha')
OSError: [Errno 9] Bad file descriptor
Exception ignored in: <_io.TextIOWrapper name='<stdout>' mode='w' encoding='utf-8'>
OSError: [Errno 9] Bad file descriptor
```
Somehow stdout is broken, when executing the second test run.
Basically, any Python `print` after exiting the function in which `AcadosOcpSolver` lives, fails on my machine.

The same error also occurred on Github Actions, see https://github.com/acados/acados/runs/4605115813?check_suite_focus=true
```
Traceback (most recent call last):
  File "marathos_test_problem.py", line 243, in <module>
    main()
  File "marathos_test_problem.py", line 59, in main
    solve_marathos_problem_with_setting(setting)
  File "marathos_test_problem.py", line 171, in solve_marathos_problem_with_setting
    ocp_solver.print_statistics()
  File "/home/runner/work/acados/acados/acadosenv/lib/python3.8/site-packages/acados_template/acados_ocp_solver.py", line 1056, in print_statistics
    print(f'{int(stat[0][jj]):d}\t{stat[1][jj]:e}\t{stat[2][jj]:e}\t{stat[3][jj]:e}\t' +
OSError: [Errno 9] Bad file descriptor
```

Both locally and on Github Actions, this is fixed when compiling `acados` without qpDUNES.
https://github.com/acados/acados/pull/772/commits/cce9d8bbb0e47f5e942c2ea98c990b367b3c9a9e

Any ideas what could be the issue are appreciated!

giaf:
I'm not too deep in python so not really idea why this happens.
[...]

Issue #782, opened 2022-01-05
Python script testing multiple solvers with OpenMP parallelization fail


FreyJo:
I realized that on the current master branch https://github.com/acados/acados/commit/91a01d4c8db13b5de0d7c600eec72a96b3b5d771
When compiling acados with openmp, the Python test scripts that test multiple solvers are failing, e.g.
https://github.com/acados/acados/blob/91a01d4c8db13b5de0d7c600eec72a96b3b5d771/examples/acados_python/tests/main_test.py
I also went back to the version 0.1.6 and got the same behavior.
It seems like the number of threads is restored on every exit of the SQP method.
Maybe someone with more experience with openmp knows what could be the issue.

schwieni:
@FreyJo: this issue should be fixed with PR #785

FreyJo:
Thanks!

PR #785, opened 2022-02-13
Fix OpenMP segfault and qpDUNES print issue

schwieni:
This PR should fix issues #777 and #782.
Explanation: both issues seem to have the same root cause, since `libacados.so` is loaded by the solver. Upon destruction of the solver it seems that also `libacados.so` is unloaded and that causes the issues.
- In #777 `qpDUNES` seems to close `stdout` during destruction.
- In #782 the threads created by `OpenMP` are spin waiting in the background and cause a segmentation fault.

The solution is to load `libacados.so` separately and avoid unloading upon destruction of the solver.

This PR also allows the two tests `main_test.py` (external call to python) and `armijo_test.py` (multiple creations/destructions of solvers in the same python session) under `examples/acados_python/tests` to run correctly.

### DIFF
--- a/acados/ocp_nlp/ocp_nlp_common.c
+++ b/acados/ocp_nlp/ocp_nlp_common.c
@@ -59,8 +59,6 @@
 
 acados_size_t ocp_nlp_config_calculate_size(int N)
 {
-    int ii;
-
     acados_size_t size = 0;
 
     // self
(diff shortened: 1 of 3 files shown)


FreyJo:
Thanks a lot for this PR and the explanations! 🙏 
This all makes sense!
I need a bit more time to check it more closely and verify that both issues are fixed by this!

I am currently doing an internship (for one more month), so I am doing acados stuff in my free time.
Thus, I am a bit slower, reviewing/merging PRs etc (see also your other one).
Thanks for the patience!

schwieni:
@FreyJo cool, thanks!
Have a good internship! :-)

PR #786, opened 2022-02-16
Template CMakeList.txt and CMake build of solvers - add MS Windows support

schwieni:
As discussed in PR #781 `make` can't be used under MS Windows/MS Visual Studio and therefore `acados` is currently not supported with MSVC.
This PR is an (initial?) implementation to solve that by (optionally) using `CMake` instead.
It was successfully tested under MSVC 2015 (not tested under Apple MacOS, since no HW available).
A positive observation is that `CMake` does not add too much overhead to the compile process here, since most things are predefined (no configuration checking and library searches involved). The compile times under Linux are comparable (within  fractions of a second).

### Overview of this PR
1. The default behavior does not change (Makefile is still used per default). The `CMake` pipeline is used only if a `CMakeBuilder` object with the appropriate settings (a default exists) is given to the constructors of `AcadosOcpSolver` or `AcadosSimSolver`. The `CMakeBuilder` object handles the build process.
2. MS Windows requires symbol export (see macro `ACADOS_SYMBOL_EXPORT` in `acados/utils/types.h`), used in the according `interfaces/acados_c` and `interfaces/acados_template/acados_template/c_templates_tera` files.
3. CTests adopted from existing tests and added (so it is tested with the github actions):
 `examples/acados_python/getting_started/ocp/example_ocp_dynamics_formulations_cmake.py`
 `examples/acados_python/getting_started/ocp/minimal_example_ocp_cmake.py` 
 `examples/acados_python/getting_started/sim/minimal_example_sim_cmake.py`
4. documentation supplemented

### Limitations
- Since `blasfeo` seems to not have accelerated support under MS Windows the observed runtimes might not be too great (at least on tests on different PCs, so no direct comparison possible).
- It is assumed that the user adapts the compiler flags, since these were not touched (MSVC only gives warnings, if fed with the `gcc` compiler flags, i.e., the `-ab` instead of `/XY` notation).

Some improvements could be made in the future:
1. default usage of `CMake` under MS Windows (and maybe under Linux?)
2. proper compiler flags for MSVC
3. (add github action to test for MSVC, if possible? Maybe with AppVeyor and Python?)
4. (faster `blasfeo` under MSVC?)
5. (test with MacOS?)


acados/utils/types.h
```diff
@@ -35,6 +35,21 @@
 #ifndef ACADOS_UTILS_TYPES_H_
 #define ACADOS_UTILS_TYPES_H_
 
+/* Symbol visibility in DLLs */
```
FreyJo, line 38: I see that you adapted this part from the `CasADi` exported functions.

`GCC_HASCLASSVISIBILITY` seems to be SWIG related https://github.com/swig/swig/blob/b4a92ae34f6d39825bc0eba056d6983575db4695/Lib/swiglabels.swg#L71
So I am not sure, if this is relevant / would work here.
Maybe these conditions should be checked instead.
https://github.com/swig/swig/blob/b4a92ae34f6d39825bc0eba056d6983575db4695/Lib/swiglabels.swg#L69

If you have a better understanding of this, a comment would be nice.

FreyJo (commented):
Thanks for the PR and fixing the conflicts!
In general looks good to me!
I added some comments and also pushed minor changes to your branch.

I merged the two example files https://github.com/acados/acados/pull/786/commits/1b843c0a1c27723b27b3361da363ee874894eb84
I think it is always nice to have a feature as an option in an existing example to see what actually has to change to use it.

I left the minimal examples though, since it is good if they stay as simple as possible.

acados/utils/types.h
schwieni: It can be that I "borrowed" this from CasADi (it is already some time ago). I probably thought if it works there it should work here as well. There is not more behind it. The main part for MSVC compatibility is
`#define ACADOS_SYMBOL_EXPORT __declspec(dllexport)`
I can remove lines 46 and 47, if you want.

interfaces/acados_template/acados_template/builders.py
```diff
@@ -0,0 +1,82 @@
...
+            cmd_str = self.get_cmd3_install()
+            print(f'call("{cmd_str}")')
+            retcode = call(cmd_str, shell=True)
+            if retcode != 0:
+                raise RuntimeError(f'Install command "{cmd_str}" was terminated by signal {retcode}')
+        except OSError as e:
+            print("Execution failed:", e, file=sys.stderr)
+        except Exception as e:
+            print("Execution failed:", e, file=sys.stderr)
+            exit(1)
```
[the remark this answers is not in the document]
schwieni: To be honest I am not entirely sure about this anymore. The main difference is that an OSError will not exit the program, while every other Exception does. This includes the RuntimeErrors which are raised in the try-block. I can remove lines 78 & 79

schwieni:
Hi @FreyJo 
thanks for looking into this!
I think the additions you made in https://github.com/acados/acados/commit/1b843c0a1c27723b27b3361da363ee874894eb84 are good and I agree with you.
You raised some questions. I am not sure it you now merged everything or are there open topics left?

PR #790, opened 2022-03-03
Rename *plan to *plan_t to fix compiliation with -Wpedantic

roangel:
When compiling with -Wpedantic, it complains that the name of the types `ocp_qp_solver_plan`, `sim_solver_plan` are the same as their respective fields in `ocp_nlp_plan` struct. I changed it to `ocp_qp_solver_plan_t` and `sim_solver_plan_t` respectively.


CMakeLists.txt
```diff
@@ -48,7 +48,11 @@ set(CMAKE_POSITION_INDEPENDENT_CODE TRUE)
 ### Defaults
 
 set(ACADOS_INSTALL_DIR "${PROJECT_SOURCE_DIR}" CACHE PATH  "Installation path to PROJECT_SOURCE_DIR")
-set(BLASFEO_TARGET "X64_AUTOMATIC" CACHE STRING "BLASFEO Target architecture")
+if ("${CMAKE_HOST_SYSTEM_PROCESSOR}" STREQUAL "aarch64")
+  set(BLASFEO_TARGET "ARMV8A_ARM_CORTEX_A57" CACHE STRING "BLASFEO Target architecture")
```
FreyJo, line 52: There are two possible BLASFEO targets, when on `aarch64`:
`ARMV8A_ARM_CORTEX_A57` and `ARMV8A_ARM_CORTEX_A53`.
https://github.com/giaf/blasfeo/blob/5346cfc600ca58e540107d45e5e0be5b7c0b0fd2/CMakeLists.txt#L54-L55

I do think it makes sense to set the default to A57, the more powerful one.
That's why we also had `X64_INTEL_HASWELL` as a default, before there was `X64_AUTOMATIC`.

Ideally, one would detect if A57 or A53 target should be used in CMake.

FreyJo:
Hey @roangel
Thanks for the PR!
I think for consistency `ocp_nlp_plan` should also be renamed to `ocp_nlp_plan_t` then.
Can you also do this, before I merge it?

roangel:
> Hey @roangel Thanks for the PR! I think for consistency `ocp_nlp_plan` should also be renamed to `ocp_nlp_plan_t` then. Can you also do this, before I merge it?

Done :)

PR #796, opened 2022-03-15
Add option to reset OCP solver

FreyJo:
Changes in C:
- update BLASFEO to version that does not propagate NaN when adding 0.0 * NaN vector, instead of the following (reverted commits)
- <del>multiply 0 with inputs instead of old values, old values might contain NaNs and 0*NaN gives NaN! </del>
- <del>add return type `ACADOS_NAN_IN_RESIDUAL` </del>
- <del>check for NaNs in residual when exiting SQP based on residual condition</del>
- fix bug in z setter
- remove empty loop

Changes in template Workflow:
- add `reset` function to acados OCP solver
- also in Simulink

Tests:
- added `reset_test.py` running solver with NaN and check if it can recover.
--> Failing before this PR and now running on Github Actions.

It is still recommended to never initialize acados with NaNs!!!


acados/ocp_nlp/ocp_nlp_common.c
```diff
@@ -2930,6 +2930,71 @@ void ocp_nlp_res_get_inf_norm(ocp_nlp_res *res, double *out)
 }
 
 
+// NOTE: if NaNs are in residual vector, blasfeo_inf_norm might return 0.
```
giaf, line 2933: The NaN case was not considered when implementing the infinity norm routine.
Namely, the routine is currently implemented using fmax, which doesn't propagate NaN.
It can be reimplemented using proper checks such that NaN is propagated.

More specifically, one way to check for NaN (also in your code below) is to check whether
`if(res_entry!=res_entry) then res_entry is NaN;`
This works on IEEE floating point standard complying compilers.
Otherwise, in C99 it is defined the function `isnan()`.

acados/ocp_nlp/ocp_nlp_cost_ls.c
```diff
@@ -793,11 +793,11 @@ void ocp_nlp_cost_ls_update_qp_matrices(void *config_, void *dims_,
 
         // tmp_ny = W * res
         blasfeo_dsymv_l(ny, 1.0, &model->W, 0, 0, &memory->res,
-                0, 0.0, &work->tmp_ny, 0, &work->tmp_ny, 0);
+                0, 0.0, &model->y_ref, 0, &work->tmp_ny, 0);
```
giaf, line 796: I also came across the issue of having NaN propagate from the y vector into the z vector even if beta==0.
This has been fixed for dgemv and dsymv in a recent commit https://github.com/giaf/blasfeo/commit/0bdc0ceb0bc51ed78802620eabd7c805044ae07e
The single-precision has not been fixed yet, this is still WIP.

I would suggest to revert these changes in acados (in order to make the code less confusing) and update BLASFEO with the fix in the dgemv and dsymv routines.

giaf: Is the issue appearing in other routines other than `dgemv` and `dsymv` ?

FreyJo: I think there was also an issue related to a `dgemm` routine in GNSF.

I can revert those commits and update BLASFEO instead.

Just realized that the commit you pointed to is only on your `dev` branch.
Are you going to merge that soon?
I just tried updating BLASFEO to your `dev` branch, but I don't know if you have a compatible HPIPM version, or which one to use.

acados/ocp_nlp/ocp_nlp_common.c
FreyJo: yes, I can also just do the `!=` comparison instead if you prefer that.
I thought, just checking for NaN when exiting SQP because of the residuals is ok.
But I can also directly check for it in res_compute if you have no concerns regarding performance.
Or do you want to change `VECNRM_INF` in Blasfeo?

acados/ocp_nlp/ocp_nlp_cost_ls.c
giaf: I will merge dev into master in a matter of days at most, I'm almost done there.

If I remember correctly dgemm should not suffer from the same issue if there is NaN in the C matrix.
But if you say it does, I can double check.

acados/ocp_nlp/ocp_nlp_common.c
giaf: I can update the `vecnrm_inf` in BLASFEO, if this PR can wait a few more days.

I didn't get the concern about performance: how would checking `!=` instead affect performance?

FreyJo: Ok, I will wait until you update the BLASFEO master.

> I didn't get the concern about performance: how would checking != instead affect performance?

I just meant checking in every residual computation vs only when exiting SQP.
But I agree checking `!=` is less confusing.

PR #797, opened 2022-03-16
Fixed minor issues in python interface for win

dkouzoup:
Suggestions for docmentation in https://docs.acados.org/python_interface/index.html:

- Step 5: path has changed from /getting_started/minimal_example_ocp.py to /getting_started/ocp/minimal_example_ocp.py
- Step 5: explicitly mention to run the example from _inside_ the folder since it imports 'common' using relative path

I still did not manage to run the example natively on windows (had to switch to WSL which works fine) due to other errors I could not fix. @FreyJo are you aware if this is even supported?

### DIFF
--- a/interfaces/acados_template/acados_template/acados_ocp_solver.py
+++ b/interfaces/acados_template/acados_template/acados_ocp_solver.py
@@ -812,7 +812,8 @@ class AcadosOcpSolver:
     """
     if sys.platform=="win32":
         from ctypes import wintypes
-        dlclose = ctypes.WinDLL('kernel32', use_last_error=True).FreeLibrary
+        from ctypes import WinDLL
+        dlclose = WinDLL('kernel32', use_last_error=True).FreeLibrary
         dlclose.argtypes = [wintypes.HMODULE]
     else:
         dlclose = CDLL(None).dlclose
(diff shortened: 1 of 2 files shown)


FreyJo:
Thanks, for the fixes!
I pushed a commit with the documentation changes you suggested.

> I still did not manage to run the example natively on windows (had to switch to WSL which works fine) due to other errors I could not fix. @FreyJo are you aware if this is even supported?

I don't consider it really supported until there is a CI test for it.
Although @schwieni  reported that it worked for him with MSVC, https://github.com/acados/acados/pull/786

@dkouzoup did you try to use the CMake option? Or can you write about where you did get stuck? Was it related to the templated Makefile?

schwieni:
> I don't consider it really supported until there is a CI test for it. Although @schwieni reported that it worked for him with MSVC, #786
> 
> @dkouzoup did you try to use the CMake option? Or can you write about where you did get stuck? Was it related to the templated Makefile?

Yes, I can confirm that it worked for me under Windows 10 using the CMake option and MSVC 2017.
I made some comments in PR #786 on github about the limitations of the PR and its external dependencies.
Maybe here, it concerns mostly the default compiler options which differ between `gcc` and MSCV's `cl.exe` (all options start with `/` instead of `-`). However, `cl.exe` only reported warnings in my case. Also, a CMakeBuilder object can be configured accordingly.
In general, I assumed that once compilation with MSVC becomes more popular that the acados CMake tool chain is extended and / or becomes more user friendly (but it is a start).

dkouzoup:
Thanks @FreyJo for the quick reply and @schwieni for resurrecting the MSVC compilation!

I actually build acados with minGW (most examples work fine, some don't) which creates an acados.lib library. Running the example first fails because it looks for a library without exension (https://github.com/acados/acados/blob/master/interfaces/acados_template/acados_template/acados_ocp_solver.py#L936). I hacked the right extension but then I get the error

`OSError: [WinError 193] %1 is not a valid Win32 application`. I will try again with MSVC

PR #834, opened 2022-07-21
add CSTR example from ELO-X course

FreyJo:


aghezz1 (approved):
Example tested also on MacOS (Intel core)

PR #859, opened 2022-09-15
Simulink: add option to output residuals and cost function value

FreyJo:
- "cost_val"
- "KKT_residuals"


aghezz1 (approved):
`KKT_residual`and `KKT_residuals` is not the naming convention.
It is clear but it might be bug prone.

PR #868, opened 2022-10-07
Matlab: fix algebraic variables offset in detect_cost_type.m

zanellia:
Hey guys! I found a small bug in the MATLAB interface - here is a quick fix.


FreyJo (approved):
Thanks Andrea!

PR #876, opened 2022-11-08
cost module for convex-over-nonlinear cost 

sandmaennchen:
* new cost module for convex-over-nonlinear cost of the form `l(x, u) = phi(y(x,u)-yref)`
* as Hessian approximation a Generalized-Gauss Newton (GGN) Hessian is used
* support for algebraic variables
* support for custom Hessian approximations of the outer loss function 'phi'


FreyJo (approved):
Just did some minor changes. Good job!

PR #878, opened 2022-11-09
Computation of W_chol (W_chol + Hessian) in nonlinear LS (linear LS) only if data changes

sandmaennchen:
* add `precompute` to all cost modules
* `LINEAR_LS`: recompute `W_chol` only if `W` has changed, recompute cost Hessian if `Cyt` or `scaling` have changed
* `NONLINEAR_LS`: recompute `W_chol` only if `W` has changed


FreyJo (approved):
More performance :heart: 
Added some minor simplifications on top :)

PR #882, opened 2022-11-15
Simplify SQP_RTI

FreyJo:
- make feedback and preparation phase functions private
- avoid casting workspace and others multiple times for performance and less code duplication


sandmaennchen (approved):
:+1:

PR #884, opened 2022-11-17
Add option to use a custom update function in Template workflow, can be called between OCP solver calls

FreyJo:
Custom C function to update solver data and parameters in between solver calls efficiently.
Example is provided.
All fields `custom_update*` are documented with Python docstrings.


sandmaennchen (approved):
:octopus:

PR #888, opened 2022-11-25
Work on core for speedups

FreyJo:
- **Cast highest level workspace (SQP/SQP-RTI) only in precompute for more speed**:
This still allows to reuse workspace across modules. This can only break if one tries to reuse the acados workspace memory from outside.
- **Add CasADi sparsity detection** that is called when creating an external function.
The CasADi sparsity struct in the generated functions contains an `int` which always say inputs are sparse, often this is not the case. Thus, we check now for speed!
- Added `sim_in_assign_and_advance`, this can be called to avoid calling `calculate_size` twice. Doing this everywhere is rather tedious and does not save very much.
- **Breaking**: precompute is now required!

On a small benchmark of an RTI with ERK and the cartpole this gives me >20% speedup, from 78μs to 62μs.

### DIFF
--- a/acados/sim/sim_erk_integrator.c
+++ b/acados/sim/sim_erk_integrator.c
@@ -440,10 +440,11 @@ acados_size_t sim_erk_workspace_calculate_size(void *config_, void *dims_, void 
 
 
 
-static void *sim_erk_cast_workspace(void *config_, void *dims_, void *opts_, void *raw_memory)
+static void *sim_erk_cast_workspace(void *config_, void *dims_, void *opts_, void *raw_memory, void *mem_)
 {
     sim_opts *opts = opts_;
     sim_erk_dims *dims = (sim_erk_dims *) dims_;
+    sim_erk_memory *mem = mem_;
 
     int ns = opts->ns;
 
@@ -478,7 +479,6 @@ static void *sim_erk_cast_workspace(void *config_, void *dims_, void *opts_, voi
         //assign_and_advance_double((num_steps + 1) * nX, &workspace->out_forw_traj, &c_ptr);
         work->out_forw_traj = d_ptr;
         d_ptr += (num_steps+1)*nX;
-        
     }
     else
     {
@@ -518,7 +518,7 @@ static void *sim_erk_cast_workspace(void *config_, void *dims_, void *opts_, voi
     // update c_ptr
     c_ptr = (char *) d_ptr;
 
-    assert((char *) raw_memory + sim_erk_workspace_calculate_size(config_, dims, opts_) >= c_ptr);
+    assert((char *) raw_memory + mem->workspace_size >= c_ptr);
 
     return (void *) work;
 }
@@ -530,6 +530,8 @@ static void *sim_erk_cast_workspace(void *config_, void *dims_, void *opts_, voi
 int sim_erk_precompute(void *config_, sim_in *in, sim_out *out, void *opts_, void *mem_,
                        void *work_)
 {
+    sim_erk_memory *mem = mem_;
+    mem->workspace_size = sim_erk_workspace_calculate_size(config_, in->dims, opts_);
     return ACADOS_SUCCESS;
 }
 
@@ -555,7 +557,7 @@ int sim_erk(void *config_, sim_in *in, sim_out *out, void *opts_, void *mem_, vo
     void *dims_ = in->dims;
     sim_erk_dims *dims = (sim_erk_dims *) dims_;
 
-    sim_erk_workspace *work = sim_erk_cast_workspace(config, dims, opts, work_);
+    sim_erk_workspace *work = sim_erk_cast_workspace(config, dims, opts, work_, mem_);
 
     int i, j, s, istep;
     double a = 0, b = 0;  // temp values of A_mat and b_vec
--- a/acados/sim/sim_erk_integrator.c
+++ b/acados/sim/sim_erk_integrator.c
@@ -623,7 +623,8 @@ int sim_erk(void *config_, sim_in *in, sim_out *out, void *opts_, void *mem_, vo
     ext_fun_arg_t expl_vde_type_out[3];
     void *expl_vde_out[3];
 
-    int nx_squared = nx * nx;
+    int nx_squared_plus_nx = nx * nx + nx;
+    int nx_times_nu = nx * nu;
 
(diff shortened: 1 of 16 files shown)


acados/sim/sim_erk_integrator.c
```diff
@@ -616,6 +618,38 @@ int sim_erk(void *config_, sim_in *in, sim_out *out, void *opts_, void *mem_, vo
     ext_fun_arg_t ext_fun_type_out[3];
     void *ext_fun_out[3];
 
+    ext_fun_arg_t expl_vde_type_in[4];
+    void *expl_vde_in[4];
+    ext_fun_arg_t expl_vde_type_out[3];
+    void *expl_vde_out[3];
+
+    int nx_squared = nx * nx;
```
sandmaennchen, line 626: you could also precompute `nx_squared_plus_nx` :laughing:

sandmaennchen (approved):

acados/sim/sim_erk_integrator.c
```diff
@@ -440,10 +440,11 @@
 
 
 
-static void *sim_erk_cast_workspace(void *config_, void *dims_, void *opts_, void *raw_memory)
+static void *sim_erk_cast_workspace(void *config_, void *dims_, void *opts_, void *raw_memory, void *mem_)
 {
     sim_opts *opts = opts_;
     sim_erk_dims *dims = (sim_erk_dims *) dims_;
+    sim_erk_memory *mem = mem_;
```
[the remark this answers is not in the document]
FreyJo: used in assert.

acados/sim/sim_erk_integrator.c
```diff
@@ -616,6 +618,39 @@ int sim_erk(void *config_, sim_in *in, sim_out *out, void *opts_, void *mem_, vo
...
+    int nx_times_nu = nx * nu;
+
+    if (opts->sens_forw)
+    {  // simulation + forward sensitivities
+        expl_vde_type_in[0] = COLMAJ;
+        expl_vde_in[0] = rhs_forw_in;  // x: nx
+        expl_vde_type_in[1] = COLMAJ;
+        expl_vde_in[1] = rhs_forw_in + nx;  // Sx: nx*nx
+        expl_vde_type_in[2] = COLMAJ;
+        expl_vde_in[2] = rhs_forw_in + nx_squared_plus_nx;  // Su: nx*nu
```
giaf, line 636: Out of curiosity, are these optimization making any difference in performance?
I would expect the bottleneck to be in the function evaluations.

If you get started with such optimizations, you can also consider using a temporary pointer to advance rhs_forw_in, so you avoid to keep performing the same additions.

giaf:
Can you explain a bit how it works the "Cast highest level workspace (SQP/SQP-RTI) only in precompute for more speed: This still allows to reuse workspace across modules".
In which sense is the workspace reused across modules?
Is the chunk of raw memory used in the workspace zeroed out again by each module?

acados/sim/sim_erk_integrator.c
FreyJo: It is hard to measure and I don't remember exactly.
But these computations were inside a double for a triple for loop, so I think the speedup of the ERK change overall is measurable.

giaf: In the speedup evaluation above, is the casadi function actually having "fake sparse" arguments?
If so, how much does it account for the speedup in the evaluation?

The general feeling is that in ERK everything other than function evaluations is pretty much negligible, and it has been coded in a "textbook" way and the compiler should anyway make a good enough job with that.
If you found this is not the case it would be interesting to investigate if there are other opportunities for code optimization, e.g. also in the floating point algebra.

acados/utils/external_function_generic.c
```diff
@@ -192,24 +190,72 @@ static int casadi_nnz(const int *sparsity)
...
+            return 1;
+        }
+        else
+        {
+            // check for casadi fake sparse
+            const int *idxcol = sparsity + 2;
+            const int *row = sparsity + ncol + 3;
+            for (jj = 0; jj < ncol; jj++)
+            {
+                if (idxcol[jj + 1] - idxcol[jj] != nrow)
```
giaf, line 214: In principle, generally speaking in the sparse matrix format the elements do not need to be ordered within a column, so here just checking that each column has exactly `nrow` elements would fail if that is not the case.

In practice, is there any CasADi documentation about assuming column elements ordering in their sparse matrix format?

acados/sim/sim_erk_integrator.c
FreyJo: Yes, there are fake sparse arguments.
I ran the same benchmark with a branch excluding the commit [b8c7034386854d6d78f2732e7c6757984a525526](https://github.com/acados/acados/pull/888/commits/b8c7034386854d6d78f2732e7c6757984a525526)
Adding more shooting intervals for more consistent timings.

| Variant | integrators | Remaining Linearization | QP Solver | Remaining QP | Remaining | 
|---------|---------|---------|---------|---------|---------|
| master                      | 26 | 46 | 48 | 34 | 36 |
| dev_c                       | 27 | 36 | 48 | 33 | 11 |
| dev_c_without_fake_sparse   | 32 | 39 | 47 | 34 | 11 |

acados_FULL_CONDENSING_DAQP_commit_comparison

acados/utils/external_function_generic.c
FreyJo: This if only says that a matrix is not dense, if there is a structural zero element in a column.
The next `for`-`if` loop additionally checks that the indices are also increasing.
I do think that this next check should not be necessary in practice, but I did not find any guarantee on CasADi side and it is a one time offline check, so better to put it there.

giaf: OK perfect then, maybe a comment explaining what the code does would be handy :)

acados/sim/sim_erk_integrator.c
giaf: Ok I see that timings are quite erratic, except there is a clear speedup in "remaining" (that is avoiding the cast of workspace I guess), and to a smaller extent the "fake sparse" casadi, even if master branch is actually faster in this benchmark.

FreyJo: Oh right, yes timing is hard :sweat_smile:

PR #890, opened 2022-11-28
Refactor Python interface

FreyJo:
- Single header file for cost and constraints, independent of type
- Structure template directory by putting Matlab related files in subfolder
- simplify CasADi function generating scripts


sandmaennchen (approved):
looks much better now! :+1:

PR #896, opened 2022-12-06
Matlab remove os path dependency

JacobDGrunnet:
When generating a Simulink s-function a number of c-files are compiled using the mex compilation setup and these are linked with a library containing the casadi_ext_fun. The compilation of this library does not use mex compilation, but is rather a system call directly to the compiler. For this to work it is required that the compiler used by mex is added to the OS PATH environment variable.

As we need to have another (incompatible) version of gcc in our OS PATH environment variable acados will use this incompatible gcc version to compile the cadasdi_ext_fun library and the linking will subsequently fail.

### DIFF
--- a/interfaces/acados_matlab_octave/ocp_generate_casadi_ext_fun.m
+++ b/interfaces/acados_matlab_octave/ocp_generate_casadi_ext_fun.m
@@ -197,19 +197,26 @@ if (strcmp(model_struct.dyn_type, 'discrete') && strcmp(model_struct.dyn_ext_fun
     c_files_path{end+1} = model_struct.dyn_generic_source;
 end
 
+%Store the current PATH environment variable value
+origEnvPath=getenv('PATH');
+
 % check compiler
 use_msvc = false;
 if ~is_octave()
     mexOpts = mex.getCompilerConfigurations('C', 'Selected');
     if contains(mexOpts.ShortName, 'MSVC')
         use_msvc = true;
+    else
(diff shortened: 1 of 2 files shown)


interfaces/acados_matlab_octave/ocp_generate_casadi_ext_fun.m
```diff
@@ -223,6 +226,13 @@ function ocp_generate_casadi_ext_fun(model_struct, opts_struct)
     % build
     compile_command = sprintf('"%s" & %s', msvc_env, build_cmd);
 else % gcc
+    %Read the mex C compiler configuration and extract the location
+    cCompilerConfig = mex.getCompilerConfigurations('C');
```
JacobDGrunnet, line 230: This command is not available in Octave.

However, it seems that Octave automatically adds a path to compatible gcc version to the PATH environment variable when Octave starts, so it should be possible just to skip all of this if octave is used

FreyJo (changes requested):
Hi, that's good to know.
I have to admit that I don't have a Casadi compatible Octave version on my machine right now, so I can't really test it.
However, that is why we have the Github tests in place.
Can you just add some `if`s to make this work for Octave?
Then the tests should also pass and I will test it for Matlab locally as well.

JacobDGrunnet:
Seems I fixed the remaining bug :D

FreyJo (approved):
Thanks for fixing the Octave version! Now all tests pass.
I just did some minor edits and will merge this.

PR #897, opened 2022-12-06
Automate matlab install

JacobDGrunnet:
Install script for automating acados installation using Matlab.

Also includes minor changes to enable full installation without user interaction.

Note: A few files were moved from examples to interfaces as they are actually needed to use acados and are thus part of the interface and not an example.

Furthermore, the include directory references were changes to point at the include directory created automatically when installing acados instead of directly from the source. This enables management of the acados install used on an enterprise level where the responsible software engineer can generate an acados package with a specific configuration and share a minimal and compiled acados distribution to the engineers using the solver generation.

### DIFF
--- a/interfaces/acados_matlab_octave/acados_template_mex/+acados_template_mex/set_up_t_renderer.m
+++ b/interfaces/acados_matlab_octave/acados_template_mex/+acados_template_mex/set_up_t_renderer.m
@@ -1,0 +1,44 @@
+function set_up_t_renderer(t_renderer_location,force)
+arguments
+    t_renderer_location
+    force=true
+end
+    message = ['\nDear acados user, we could not find t_renderer binaries,',...
+        '\n which are needed to export templated C code from ',...
+        'Matlab.\n Press any key to proceed setting up the t_renderer automatically.',...
+        '\n Press "n" or "N" to exit, if you wish to set up t_renderer yourself.\n',...
+        '\n https://github.com/acados/tera_renderer/releases'];
+    
+    if(~force)
+        In = input(message,'s');
+        if strcmpi( In, 'n')
+            error('Please set up t_renderer yourself and try again');
+        end
+    end
+	t_renderer_version = 'v0.0.34';
+	if ismac()
+		suffix = '-osx';
+	elseif isunix()
+		suffix = '-linux';
+	elseif ispc()
+		suffix = '-windows';
+	end
+	acados_root_dir = getenv('ACADOS_INSTALL_DIR');
+
+	tera_url = ['https://github.com/acados/tera_renderer/releases/download/', ...
+			t_renderer_version '/t_renderer-', t_renderer_version, suffix];
+	destination = fullfile(acados_root_dir, 'bin');
+	tmp_file = websave(destination, tera_url);
+
+	if ~exist(destination, 'dir')
+		[~,~] = mkdir(destination);
+	end
+
+	movefile(tmp_file, t_renderer_location);
+
+	if isunix()
+		% make executable
+		system(['chmod a+x ', t_renderer_location]);
+	end
+	fprintf('\nSuccessfully set up t_renderer\n')
+end
--- a/interfaces/acados_matlab_octave/acados_template_mex/+acados_template_mex/set_up_t_renderer.m
+++ b/interfaces/acados_matlab_octave/acados_template_mex/+acados_template_mex/set_up_t_renderer.m
@@ -1,44 +1,56 @@
-function set_up_t_renderer(t_renderer_location,force)
-arguments
-    t_renderer_location
-    force=true
-end
+function set_up_t_renderer(t_renderer_location,varargin)
+% set_up_t_renderer(t_renderer_location,[force])
+% If force is not provided it is default set to true
+
+    %The first variable parameter determines whether to force install
+    switch(nargin)
+        case 1
+            force=true;
+        case 2
+            force=varargin{1};
+        otherwise
+            error('function called with %d parameters, was expecting max 2',nargin);       
+    end
+
+
     message = ['\nDear acados user, we could not find t_renderer binaries,',...
         '\n which are needed to export templated C code from ',...
         'Matlab.\n Press any key to proceed setting up the t_renderer automatically.',...
         '\n Press "n" or "N" to exit, if you wish to set up t_renderer yourself.\n',...
         '\n https://github.com/acados/tera_renderer/releases'];
-    
+
     if(~force)
         In = input(message,'s');
         if strcmpi( In, 'n')
             error('Please set up t_renderer yourself and try again');
         end
     end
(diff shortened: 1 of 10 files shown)


FreyJo (changes requested):
Thanks! This seems convenient!

However, I do not want to break Octave compatibility here as well:
The tests are now failing with the error:
`scalar cannot be indexed with .`
Could you look into that and ideally fix it?

Moreover: Could you update the docs, to describe how the interface should be used after your proposed change?
Like: How to call the `acados_install` what are prerequisites to do so?

https://github.com/acados/acados/blob/master/docs/matlab_octave_interface/index.md

JacobDGrunnet:
Rebased and now all tests passed here as well.

Be aware that I have been using the argument block in Matlab which was not introduced untill version 2019b

@FreyJo please have a look and if you are a happy with the approach I will update the index.md file as requested.

The automated installation works for windows with minGW following the guide given on the below URL

https://docs.acados.org/installation/index.html#windows-for-use-with-matlab 

The above documentation seems more complete than the .md file.

interfaces/acados_matlab_octave/ocp_set_ext_fun.m
```diff
@@ -43,11 +43,11 @@
 
 % set paths
 acados_mex_folder = fullfile(acados_folder, 'interfaces', 'acados_matlab_octave');
-acados_include = ['-I' acados_folder];
+acados_include = ['-I' fullfile(acados_folder,'include')];
```
FreyJo, line 46: Good catch, the folders here should be correct!
Probably, it should be adjusted in `interfaces/acados_matlab_octave/sim_set_ext_fun.m` as well

interfaces/acados_matlab_octave/acados_template_mex/+acados_template_mex/set_up_t_renderer.m
```diff
@@ -0,0 +1,44 @@
...
+        'Matlab.\n Press any key to proceed setting up the t_renderer automatically.',...
+        '\n Press "n" or "N" to exit, if you wish to set up t_renderer yourself.\n',...
+        '\n https://github.com/acados/tera_renderer/releases'];
+    
+    if(~force)
+        In = input(message,'s');
+        if strcmpi( In, 'n')
+            error('Please set up t_renderer yourself and try again');
+        end
+    end
```
FreyJo, line 17: Indent is weird here.

FreyJo (changes requested):
I just went through the changes.

> Be aware that I have been using the argument block in Matlab which was not introduced untill version 2019b

I would like to not make this restriction. There are people stuck with older Matlab versions/licenses I guess.
I see that `arguments` might be more readable, but would like to ask you to replace this using `varargin`.

> The above documentation seems more complete than the .md file.

Seems like I linked the wrong file above.
It indeed makes more sense to document it here.
https://github.com/acados/acados/blob/master/docs/installation/index.md
Overall, the https://docs.acados.org page is built from the files in https://github.com/acados/acados/blob/master/docs

I'll try to test this new installation soon myself before merging.

interfaces/acados_matlab_octave/ocp_set_ext_fun.m
JacobDGrunnet: I updated the above mentioned filed, but I have not tested this my self. Hopefully this is covered by the CI-test

JacobDGrunnet:
> I just went through the changes.
> 
> > Be aware that I have been using the argument block in Matlab which was not introduced untill version 2019b
> 
> I would like to not make this restriction. There are people stuck with older Matlab versions/licenses I guess. I see that `arguments` might be more readable, but would like to ask you to replace this using `varargin`.

It is a lot more readable :D 

I have changed to varargin.

> > The above documentation seems more complete than the .md file.
> 
> Seems like I linked the wrong file above. It indeed makes more sense to document it here. https://github.com/acados/acados/blob/master/docs/installation/index.md Overall, the https://docs.acados.org page is built from the files in https://github.com/acados/acados/blob/master/docs

When you have approved the code-changes I will make and update to the documentation before you merge.

> 
> I'll try to test this new installation soon myself before merging.

interfaces/acados_matlab_octave/acados_template_mex/+acados_template_mex/set_up_t_renderer.m
JacobDGrunnet: Fixed

JacobDGrunnet: How is it wierd? 

There is a nested if statement indentation follows the blocks.

FreyJo: Right, sorry, I don't know why I was confused..

FreyJo:
I pushed a commit updating the acados_install_windows file.
Feel free to make the documentation commit before I test it myself, then I can directly merge.

FreyJo (approved):
A colleague of mine just tested it successfully.

The only pitfall was providing the acadosPath as an optional argument here:
```
    switch(nargin)
        case 0
            acadosPath=pwd;
        case 1
            acadosPath=varargin{1};
```
I think it could also be a non optional argument, or you specify how to use the script precisely in the docu to be added.
Otherwise, all good. Thanks! :raised_hands:

JacobDGrunnet:
> A colleague of mine just tested it successfully.
> 
> The only pitfall was providing the acadosPath as an optional argument here:
> 
> ```
>     switch(nargin)
>         case 0
>             acadosPath=pwd;
>         case 1
>             acadosPath=varargin{1};
> ```
> 
> I think it could also be a non optional argument, or you specify how to use the script precisely in the docu to be added. Otherwise, all good. Thanks! 🙌

I see. This was a choice I made when the install script was in the acados root :D

I would like to make a small update such that it is possible to specify custom cmake options as an optional input. However, I can make the acados path default path a lot smarter though....

Give me a little time and I will push a suggested change

JacobDGrunnet:
I added a smarter way of autodetecting the acados root folder (just relative to the locatio of the file)

Also it is now possible to add a user defined CMAKE configuration as a input parameter for the installs script

If this interface change is approved I can update the documentation

FreyJo (approved):
Given that you detect the acados path now, I would suggest to remove it as an argument from the function.
As it is now, the user needs to provide the path if they want to change the default CMake config, which is not so convenient.
Or is there a reason why you think the path detection fails?

Otherwise, it all looks good to me!

JacobDGrunnet:
I kept the path as optional as I thought it might fail in the case where a developer has multiple acados repositories checked out.

However, on further reflection I do not think it can fail unless you have another acados installation in your Matlab path and I guess that failure in this case is a user error.

JacobDGrunnet:
And the force push was to remove the bug I introduced :D

PR #900, opened 2022-12-11
DAQP: Abort solve if setup fails

darnstrom:
If the problem setup fails (for example due to trivial infeasibility or nonconvex hessian), abort the qp solve.


FreyJo (approved):
Thanks! :+1:

PR #903, opened 2023-01-03
update docs with DYLD_LIBRARY_PATH, remove duplicate / outdated documentation

FreyJo:


aghezz1 (approved):
To edit the DYLD_LIBRARY_PATH on MacOS you need to disable the "SIP" (system integrity protection). I recently discovered that one could edit the "DYLD_FALLBACK_LIBRARY_PATH" which does not required to disable anything and it should be safer. 
I haven't tested it though. Should we change the instructions?

PR #907, opened 2023-01-10
Add __init__.py for gnsf subpackge in acados_template

lukasfro:
Dear Acados developers,

first of all, hats off to your efforts in creating this amazing library! I have been using Acados over the last couple of weeks, and I'm impressed by the performance and ease of use via the Python interface. However, when integrating Acados into our CI pipeline, I realized that the `gnsf` subpackage in the `acados_template` package is not properly installed. I narrowed the issue down to a missing `__init__.py`, so that `gnsf` is not detected as subpackage under certain circumstances. 

I hope that with this small PR, I can contribute to making Acados ever so slightly better.

Cheers


FreyJo (approved):
Happy to hear you are enjoying acados!
Thanks for the fix!

PR #908, opened 2023-01-13
Update non-.py file list for python installation

lukasfro:
Hi @FreyJo,

I found another small issue with the `acados_template` package. It seems that the file list in `setup.py` has not been updated accordingly during the refactoring of the Python interface from PR #890.

Cheers


interfaces/acados_template/setup.py
```diff
@@ -64,27 +64,20 @@
...
-        'c_templates_tera/p_constraint.in.h',
-        'c_templates_tera/h_constraint.in.h',
-        'c_templates_tera/h_e_constraint.in.h',
-        'c_templates_tera/phi_e_constraint.in.h',
-        'c_templates_tera/phi_constraint.in.h',
+        'c_templates_tera/constraints.in.h',
+        'c_templates_tera/cost.in.h',
+        'c_templates_tera/main.in.c',
+        'c_templates_tera/main_sim.in.c',
         'c_templates_tera/make_sfun.in.m',
```
FreyJo, line 79: The file is now in `interfaces/acados_template/acados_template/c_templates_tera/matlab_templates/make_sfun.in.m`.
Probably all templates should be included here, i.e. also `matlab_templates`.
I think the failing CI was unrelated to this change.
The mex interface does not use
`interfaces/acados_template/setup.py`.

lukasfro: > Probably all templates should be included here, i.e. also matlab_templates.

Ah, I've overlooked these files. Just one quick question: what's the reasoning to add the Matlab/Simulink related files in the Python interface? Wouldn't it be quite odd that someone uses the Python interface to generate mex files?

> I think the failing CI was unrelated to this change.

I believe that archive.ubuntu.com had some issues last week, which might explain the failing pipelines.

Anyway, thanks for your feedback. I will make the proposed change later this week.

FreyJo: > Just one quick question: what's the reasoning to add the Matlab/Simulink related files in the Python interface? Wouldn't it be quite odd that someone uses the Python interface to generate mex files?

There might be people prototyping in Python that have the need to export an Sfunction block of their solver, since there is some deployment workflow for their embedded target.
But it doesn't seem very convenient I agree :D

FreyJo (approved):
:+1:

PR #910, opened 2023-01-16
RK3 implementation and sanity check for number of RK stages of plant and model

sourishpramanick:


FreyJo (approved):
Looks good! Thanks for the contribution!

Since there are quite a few explicit RK methods of RK3, I just wanted to note that you went for:
"Kutta's_third-order_method"
https://en.wikipedia.org/wiki/List_of_Runge%E2%80%93Kutta_methods#Kutta's_third-order_method

PR #917, opened 2023-03-09
Add acados_env_variables_windows.m in examples folder

FreyJo:
see:

https://discourse.acados.org/t/acados-example-in-matlab/957/4

https://github.com/acados/acados/pull/897


sandmaennchen (approved):
thanks!

PR #928, opened 2023-05-18
added verbose option to AcadosOcpSolver build

tudoroancea:
This PR answers my own question on the [discourse forum](https://discourse.acados.org/t/disabling-compilation-stdout-output/1019?u=tudoroancea).
It simply adds a verbose argument to some of the functions of the Python interface to disable the compiler output. 
I achieved this by replacing the calls to `os.system()` by calls to `subprocess.call()` which can properly redirect the output to `/dev/null` as indicated in this [StackOverflow answer](https://stackoverflow.com/a/5597017).
I have added this option for the `CMakeBuilder` and `AcadosSimSolver` classes as well, but I may still have missed some parts of the code to change for consistency, please let me know.


FreyJo (approved):
Thanks a lot for the PR!
It is really useful :)
I pushed some very minor improvements.

Issue #765, opened 2021-11-02
Make MEX wrapper around templated solver work on Windows


FreyJo:
Just to keep track of this, since it would be useful for all Simulink and Windows users.
current error: https://pastebin.com/hPCTV6KY
probably templated Makefile needs to be adapted

From https://github.com/acados/acados/pull/712

FreyJo:
TODO:
try switching to templated CMake and see if / how compatible it is with Windows matlab!

AkashJohnSubash:
@sourishpramanick You got this.

FreyJo:
Proof of concept PR https://github.com/acados/acados/pull/930 is merged.
Big thanks @yzuuang :fire_extinguisher: 
Let's remove the backend of the old Matlab interface completely and make it use the templates!
For reference on the duplicated backends we have now:
https://github.com/acados/acados/blob/2a0c1e20072a7014a74e337728ee13a01c3a2204/docs/matlab_octave_interface/mex_interface_overview.png
https://docs.acados.org/matlab_octave_interface/index.html#interface-structure

PR #934, opened 2023-06-02
Matlab full template based OCP solver

yzuuang:
The Matlab OCP solver interface is now also template based.
This closes https://github.com/acados/acados/issues/765


FreyJo (approved):
It looks very good! Happy that it works!
We discussed the cleanup of external QP solver flags to be pushed still.

PR #939, opened 2023-06-30
NLS cost module: Change W_changed flag if cost matrix was changed

lukasfro:
Hi @FreyJo 

I recently switched from a `LINEAR_LS` to a `NONLINEAR_LS` cost module in my problem formulation. I realized that the cost
 matrices were not properly updated when I set them online. It appears that there was a minor mistake in #878, i.e., the `W_changed` flag is only set to true for the linear least squares module: https://github.com/acados/acados/blob/master/acados/ocp_nlp/ocp_nlp_cost_ls.c#L301

Cheers and kind regards from Freiburg ✌🏻


FreyJo (approved):
Good catch! :+1: 
Kind regards from Freiburg too :slightly_smiling_face:

PR #940, opened 2023-07-03
Example to create Simulink wrapper from Python

FreyJo:
As requested here:
https://discourse.acados.org/t/create-simulink-wrapper-from-python/1120/2


sandmaennchen (approved):
nice! :yum:

PR #942, opened 2023-07-06
Matlab: add option to name the c_generated_code folder

sourishpramanick:
- Added options for user to define the folder name for c_generated_code  for a particular OCP, thus enabling the user to execute 2 or more OCPs in the same simulink model
- extensive_example_ocp.m : clear all at the top erases the simulink_opts defined in simulink_example_advanced.m - hence removed;  U_max is used in the simulink model, but missing in the script, hence added; added code for renaming c-generated code folder; added simulink_opts to acados_ocp call
- simulink_example_advanced.m : introduced the use of ocp_opts.opts_struct.code_export_directory instead of hardcoded 'c_generated_code'

### DIFF
--- a/examples/acados_python/multiphase_nonlinear_constraints/main.py
+++ b/examples/acados_python/multiphase_nonlinear_constraints/main.py
@@ -1,0 +1,140 @@
+import numpy as np
+import casadi as ca
+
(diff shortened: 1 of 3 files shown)


FreyJo (approved):
I fixed the conflict and did a minor change. Will merge if tests still pass.

PR #944, opened 2023-07-13
MATLAB: use template-based code to interface acados sim solver

FreyJo:
- made Mex wrapper around templated integrator
- modify `acados_sim` to use template-code based integrator
- fix the error of compiling with openmp on Windows


yzuuang:
MATLAB: use template-based code to interface acados sim solver

interfaces/acados_matlab_octave/detect_dims_sim.m
```diff
@@ -52,12 +52,18 @@
         model.dim_np = 0;
     end
 
+    if isfield(model, 'sym_xdot')
+        if numel(model.sym_xdot) ~= model.dim_nx
+            warning('sym_xdot is not of shape nx');
```
sandmaennchen, line 57: error instead of warning here?

sandmaennchen (approved):

interfaces/acados_matlab_octave/detect_dims_sim.m
FreyJo: `xdot` is not needed for explicit integrator.
So, I went for warning.

PR #945, opened 2023-07-14
Feature/colcon compatible cmake

nielsvd:
This rework of the CMake build pipeline has two aims:
* Introduce compatibility with [colcon](https://colcon.readthedocs.io/en/released/) and thereby ROS2.
* Be able to build `acados` without `colcon`.
* Otherwise minimize the number of overall changes to not break other use cases where `acados` is currently deployed.

High-level changes:
* We introduce three packages: `acados`, `acados_matlab_octave`, and `acados_template`. To this end we have simply moved the matlab and python interfaces out of the `interfaces` folder. Each with their own `package.xml` file.
* Each of these packages have their own subdirectory, effectively that means that most of the source files of `acados` have been moved into a subdirectory of the repository.
* Do not build the external packages by source by default, treat them as proper external packages. Only implemented for `blasfeo` and `hpipm` so far, also see: [boschresearch/blasfeo-vendor](https://github.com/boschresearch/blasfeo-vendor) and [boschresearch/hpipm-vendor](https://github.com/boschresearch/hpipm-vendor).

TODO:
* Re-ntegrate Python and Matlab tests.
* Test other dependencies as external packages.
* Update README.md.


nielsvd:
Opened PR in wrong place: see https://github.com/boschresearch/acados/pull/1

PR #952, opened 2023-08-02
MEX: remove pointers to external functions in Matlab

mohammed-hababeh:
Pointers were only needed to set parameters in the old interface.
Now they are internally in the solver capsule.

Removed and adapted some code in:

-  acados_mex_create.in.c
- acados_mex_set.in.c
- mex_solver.in.m


FreyJo (approved):
Looks good, thanks!

PR #953, opened 2023-08-09
Use PROJECT_SOURCE_DIR to allow for better submodulization

apozharski:
When using cmake `add_subdirectory` if using acados as a submodule, the use of `CMAKE_SOURCE_DIR` adds the wrong directories. Using `PROJECT_SOURCE_DIR` instead produces the same results when building acados on its own but also adds the correct directories when included in other CMAKE projects.


FreyJo (approved):
Makes sense, thanks!

PR #954, opened 2023-08-09
Update problem_formulation_ocp_mex.tex with J_sbx^e

jnz:
Is this correct? Adding ^e to J_sbx here.


FreyJo (approved):
That is correct, thanks!

PR #958, opened 2023-08-15
Minimal closed-loop example using RTI

sandmaennchen:


FreyJo (approved):
Looks good. Users will love it! :heart:

PR #961, opened 2023-08-24
Python get_stats returns float or int instead of single element numpy array and fix cython for custom path

aghezz1:
- Get statistics with their correct types in both python and cython interface - avoid creation of 1-dim numpy array
- Make cython interface working with custom path of the `code_export_directory`

For Release notes:
- BREAKING: change return types of `get_stats` in Python


sandmaennchen (approved):
:+1:

PR #968, opened 2023-09-20
Windows python interface

asparc:
Enable python interface on Windows 

As the python interface requires acados to be compiled as a shared library,
MinGW & MSVC compilation with the shared library options needed to be fixed.

Required/implemented changes:

- Export shared library symbols on windows
- Prevent acados linking to math libary `m` on windows
- Set consistent library suffixes, prefixes and folders for windows
- Get python DLL loader/reader to work on windows
- Fix template renderer for subfolders on windows 
- Add support for specifying host architecture in CMakeBuilder python class
- Reset generated cmake cache to enable consecutive cmake runs


FreyJo (approved):
Thanks for the PR, great that you made it work :)
The changes look good to me!

I or a colleague of mine will test it before merging.

asparc:
Sounds good :-)

interfaces/acados_template/acados_template/utils.py
```diff
@@ -165,14 +165,25 @@ def make_model_consistent(model):
...
+        return '.dylib'
     elif os.name == 'nt':
-        lib_ext = ''
+        return '.dll'
+    else:
+        return '.so'
+
+def get_shared_lib_dir():
+    if os.name == 'nt':
+        return 'bin'
```
FreyJo, line 178: I am a bit confused by this. Can you explain?
We should also test Windows on Matlab before merging this.
As far as I remember, it expects the libs to be in `lib`.

asparc: Windows, and particularly MSVC, behaves a bit different than Linux when it comes to shared libraries. Windows also uses so called import libraries, containing stubs to the exported functions of the shared library. 

- Import libraries (`.lib` on windows - confusingly just like static libraries are `.lib`) go to the `lib\` folder and are needed at link time only.
- Shared libraries (`.dll` on windows) go to the `bin\` folder and are needed at runtime only.

Different compilers have different ways of dealing with them:
- MSVC cannot link directly to a `.dll` shared library. At link time it really needs the `.lib` import library.
- MinGW doesn't really need import libraries, as it can directly link to a `.dll`, but - I guess for uniformity - it can also handle import libraries (which it would typically give the extension `.dll.a`, but I renamed that suffix in CMake to `.lib`).

So:
- At link time, the `.lib` import libraries in the `lib\` folder are required, also for e.g. `mex`-ing with Matlab.
- At run time, the `.dll` shared libraries in the `bin\` folder are required. The Python DLL loader (`ctypes.WinDLL`) is reading the shared libraries at run time, so needs to get them from the `bin\` folder.

PR #970, opened 2023-09-26
Fix typo in minimal_example_closed_loop.py

Federico-PizarroBejarano:
The median computation time was accidentally coded as `np.min`.


FreyJo (approved):
Thanks!

PR #971, opened 2023-09-29
Fix dependency handling in CMake config file

Hs293Go:
This PR addresses 3 issues when consuming acados as a CMake dependency from a downstream project. 

1. Currently `find_package(acados)` breaks all following `find_package` calls since `acadosConfig.cmake` adjusts the `CMAKE_MODULE_PATH`. This should be restored to the original state per [this cmake dev](https://discourse.cmake.org/t/install-findpackage-script/5307/2) in [the manner shown here](https://discourse.cmake.org/t/cant-export-top-level-lib-with-dependency-on-subdir-lib/1084/21).
2. `add_library(gfortran)`  and `add_library(openblas)` triggers *target with the same name already exists* errors if the acados config file is processed twice, so they should be [guarded against double `add_library` by if guards](https://github.com/microsoft/vcpkg/issues/24367#issuecomment-1107798861)
3. If acados is build with `ACADOS_WITH_OPENMP`, the dependency on `OpenMP` is not handled in the config file

### DIFF
--- a/cmake/acadosConfig.cmake.in
+++ b/cmake/acadosConfig.cmake.in
@@ -29,6 +29,7 @@
 #
 
 # Enabled external modules
+set(ACADOS_WITH_OPENMP @ACADOS_WITH_OPENMP@)
 set(ACADOS_WITH_HPMPC @ACADOS_WITH_HPMPC@)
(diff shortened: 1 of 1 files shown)


FreyJo (approved):
Makes sense!
Thanks a lot for the PR :+1:

PR #975, opened 2023-10-16
Work on solution sensitivities

FreyJo:
- looked into solution sensitivities
- documented requirements for `eval_param_sens` to work properly
- add getters for HPIPM matrices
- improve Python docstrings


acados/ocp_qp/ocp_qp_common.h
```diff
@@ -69,6 +69,7 @@ typedef struct
     void (*memory_get)(void *config_, void *mem_, const char *field, void* value);
     acados_size_t (*workspace_calculate_size)(void *config, void *dims, void *opts);
     int (*evaluate)(void *config, void *qp_in, void *qp_out, void *opts, void *mem, void *work);
+    void (*solver_get)(void *config_, void *qp_in_, void *qp_out_, void *opts_, void *mem_, const char *field, int stage, void* value);
```
giaf, line 72: I guess this should be initialized also for the QP solvers other than HPIPM, otherwise you would get a segfault or something.

acados/ocp_qp/ocp_qp_xcond_solver.c
```diff
@@ -371,6 +371,21 @@ void ocp_qp_xcond_solver_memory_reset(void *config_, ocp_qp_xcond_solver_dims *d
...
+void ocp_qp_xcond_solver_get(void *config_, ocp_qp_in *qp_in, ocp_qp_out *qp_out, void *opts_, void *mem_, const char *field, int stage, void* value)
+{
+    // cast data structures
+    ocp_qp_xcond_solver_config *config = config_;
+    qp_solver_config *qp_solver = config->qp_solver;
+    ocp_qp_xcond_config *xcond = config->xcond;
+    ocp_qp_xcond_solver_opts *opts = opts_;
+    ocp_qp_xcond_solver_memory *mem = mem_;
+
+    qp_solver->solver_get(qp_solver, qp_in, qp_out, opts->qp_solver_opts, mem->solver_memory, field, stage, value);
```
giaf, line 383: This would need to be implemented for all ocp as well as dense QP solvers.

acados/ocp_qp/ocp_qp_hpipm.c
```diff
@@ -323,6 +323,28 @@ void ocp_qp_hpipm_memory_reset(void *config_, void *qp_in_, void *qp_out_, void
     ocp_qp_hpipm_memory_assign(config_, qp_in->dim, opts_, mem_);
 }
 
+void ocp_qp_hpipm_solver_get(void *config_, void *qp_in_, void *qp_out_, void *opts_, void *mem_, const char *field, int stage, void* value)
```
giaf, line 326: This would work as expected only if there is no partial condensing and no x0 elimination.
E.g. some possible issues:
- in case of partial condensing, it could be stage > N of qp_in   =>   random result or segfault
- in case of partial condensing, it could be nu[stage] in qp_in > nu NLP that is used in ocp_nlp_interface to tell the expected sizes   =>   memory corruption or segfault
- in case of x0 elimination, there would be no K[0] or P[0] returned at all in the value buffer

examples/acados_python/pendulum_on_cart/solution_sensitivities/policy_gradient_example.py
```diff
@@ -0,0 +1,382 @@
...
+    min_abs_eig_P = 1e12
+    for i in range(1, N_horizon):
+        P_mat = acados_solver.get_from_qp_in(i, 'P')
+        B_mat = acados_solver.get_from_qp_in(i, 'B')
+        R_mat = acados_solver.get_from_qp_in(i, 'R')
+        proj_hess_block = R_mat + B_mat.T @ P_mat @ B_mat
+        eigv = np.linalg.eigvals(proj_hess_block)
+        min_eigv = np.min(eigv)
+        min_eig_proj_hess = min(min_eigv, min_eig_proj_hess)
+        min_abs_eig_proj_hess = min(min_abs_eig_proj_hess, np.min(np.abs(eigv)))
```
giaf, line 220: Why do you compute the minimum of the absolute value of the eigenvalues?
As long as they are not numerically very close to zero, the issue is just if there are negative eigenvalues.

acados/ocp_qp/ocp_qp_common.h
FreyJo: Addressed in https://github.com/acados/acados/pull/975/commits/4a8c43cddc7292876b8d5628ad1253f7857703cb

acados/ocp_qp/ocp_qp_hpipm.c
FreyJo: I addressed this here. But only in Python https://github.com/acados/acados/pull/975/commits/6c37ff53bb7c288917dfc4e40652589da946b5d4

examples/acados_python/pendulum_on_cart/solution_sensitivities/policy_gradient_example.py
FreyJo: > As long as they are not numerically very close to zero, the issue is just if there are negative eigenvalues.

I wanted to exactly check if they are close to zero.
I also plotted the sign of the smallest eigenvalue now.

interfaces/acados_c/ocp_nlp_interface.c
```diff
@@ -780,7 +780,7 @@ void ocp_nlp_qp_dims_get_from_attr(ocp_nlp_config *config, ocp_nlp_dims *dims, o
         dims_out[0] = 1;
         dims_out[1] = dims->nx[stage+1];
     }
-    else if (!strcmp(field, "Q"))
+    else if (!strcmp(field, "Q") || !strcmp(field, "P"))
```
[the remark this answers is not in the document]
FreyJo: Also addressed in Python here:
https://github.com/acados/acados/pull/975/commits/6c37ff53bb7c288917dfc4e40652589da946b5d4

acados/ocp_qp/ocp_qp_hpipm.c
giaf: Can you also add a comment in the C interface to warn that this would not work in general, and maybe as a TODO that someone can fix it in the future?

giaf: Actually an easy solution would be to pass two additional arguments to the solver_get function, namely int size1, int size2. These are set in the ocp_nlp_interface.c using the NLP dim there, and compared with the dim in ocp_qp_in. For stage it has to be stage < or <= the N in the ocp_qp_in dims, depending on the field needed.

FreyJo: Done as suggested in https://github.com/acados/acados/pull/975/commits/15b74362f7231824f7caaa7b391fb3cca2b12f61

PR #977, opened 2023-10-23
Add time in IRK, allow to cost integration with time dependent function

FreyJo:
- add time to IRK functions
- add time to `NONLINEAR_LS` and `CONVEX_OVER_NONLINEAR` cost functions
- IRK: remove old implementation to compute hessian with CasADi multiplication
- minor changes in zoRO plots

Notes:
- Integrator now simulates from `t0` to `t0+T`.
-  `t0 = 0` by default in integrator. Can be updated using `integrator.set("t0", value)`.
- In OCP: every integrator integrates from `t0=0` to `dt_i` (ith time step in time discretization grid)
- Time dependent cost functions are only meaningful in case cost integration is chosen.


sandmaennchen (approved):
:+1:

PR #982, opened 2023-10-28
Add CodeQL Workflow for Code Security Analysis

b4yuan:
# Summary

This pull request introduces a CodeQL workflow to enhance the security analysis of this repository.

# What is CodeQL

CodeQL is a static analysis tool that helps identify and mitigate security vulnerabilities. It is primarily intra-function but does provide some support for inter-function analysis. By integrating CodeQL into a GitHub Actions workflow, it can proactively identify and address potential issues before they become security threats.

For more information on CodeQL and how to interpret its results, refer to the GitHub documentation and the CodeQL documentation (https://codeql.github.com/ and https://codeql.github.com/docs/).

# What this PR does

We added a new CodeQL workflow file (.github/workflows/codeql.yml) that
- Runs on every pull request (functionality to run on every push to main branches is included as a comment for convenience).
- Runs daily.
- Excludes queries with a high false positive rate or low-severity findings.
- Does not display results for git submodules, focusing only on our own codebase.

## Validation

To validate the functionality of this workflow, we have run several test scans on the codebase and reviewed the results. The workflow successfully compiles the project, identifies issues, and provides actionable insights while reducing noise by excluding certain queries and third-party code.

## Using the workflow results

If this pull request is merged, the CodeQL workflow will be automatically run on every push to the main branch and on every pull request to the main branch. To view the results of these code scans, follow these steps:
1. Under the repository name, click on the Security tab.
2. In the left sidebar, click Code scanning alerts.

# Is this a good idea?

We are researchers at Purdue University in the USA. We are studying the potential benefits and costs of using CodeQL on open-source repositories of embedded software.

We wrote up a report of our findings so far. The TL;DR is “CodeQL outperforms the other freely-available static analysis tools, with fairly low false positive rates and lots of real defects”. You can read about the report here: https://arxiv.org/abs/2310.00205

# Review of engineering hazards

License: CodeQL operates under the MIT License which grants anyone the right to use, copy, modify, merge, publish, distribute, sublicense, and/or sell (commercially or non-commercially) the software as-is.

False positives: We find that around 20% of errors are false positives, but that these FPs are polarized and only a few rules contribute to most FPs. We find that the top rules contributing to FPs are: cpp/uninitialized-local, cpp/missing-check-scanf, cpp/suspicious-pointer-scaling, cpp/unbounded-write, cpp/constant-comparison, and cpp/inconsistent-null-check. Adding a filter to filter out certain rules that contribute to a high FP rate can be done simply in the workflow file.

### DIFF
--- a/.github/workflows/codeql-buildscript.sh
+++ b/.github/workflows/codeql-buildscript.sh
@@ -1,0 +1,5 @@
+#!/usr/bin/env bash
+
+mkdir build && cd build
+cmake ../
+make
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -14,11 +14,10 @@ name: "CodeQL"
 on:
   push:
     branches: [ "main", "master" ]
-  pull_request:
-    # The branches below must be a subset of the branches above
-    branches: [ "main", "master" ]
   schedule:
-    - cron: '28 21 * * 0'
+    - cron: '0 0 * * *'
+  pull_request:
+    branches: '*'
 
 jobs:
   analyze:
@@ -103,21 +102,25 @@ jobs:
           -**:cpp/suspicious-pointer-scaling
           -**:cpp/suspicious-pointer-scaling-void
           -**:cpp/unsigned-comparison-zero
-          -**/third*party/**
-          -**/3rd*party/**
-          -**/external/**
+          -**/cmake*/Modules/**
         input: ${{ steps.step1.outputs.sarif-output }}/cpp.sarif
         output: ${{ steps.step1.outputs.sarif-output }}/cpp.sarif
 
-    - name: Upload SARIF
+    - name: Upload CodeQL results to code scanning
       uses: github/codeql-action/upload-sarif@v2
       with:
         sarif_file: ${{ steps.step1.outputs.sarif-output }}
         category: "/language:${{matrix.language}}"
 
-    - name: Archive CodeQL results
+    - name: Upload CodeQL results as an artifact
+      if: success() || failure()
       uses: actions/upload-artifact@v3
       with:
         name: codeql-results
         path: ${{ steps.step1.outputs.sarif-output }}
         retention-days: 5
+
+    - name: Fail if an error is found
+      run: |
+        ./.github/workflows/fail_on_error.py \
+          ${{ steps.step1.outputs.sarif-output }}/cpp.sarif
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,123 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ "main", "master" ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ "main", "master" ]
+  schedule:
+    - cron: '28 21 * * 0'
+
+jobs:
+  analyze:
+    name: Analyze
+    # Runner size impacts CodeQL analysis time. To learn more, please see:
+    #   - https://gh.io/recommended-hardware-resources-for-running-codeql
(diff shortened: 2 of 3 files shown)


.github/workflows/codeql-buildscript.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
+mkdir build && cd build
+cmake ../
+make
```
FreyJo, line 5: Maybe a stupid question: (Why) is a full build required for a static analyzer?

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,126 @@
...
+#
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ "main", "master" ]
+  schedule:
+    - cron: '0 0 * * *'
+  pull_request:
+    branches: '*'
```
FreyJo, line 20: I think running on pull requests is enough for us.

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,126 @@
...
+
+    - name: Perform CodeQL Analysis
+      uses: github/codeql-action/analyze@v2
+      with:
+        category: "/language:${{matrix.language}}"
+        upload: false
+      id: step1
+
+    # Filter out rules with low severity or high false positve rate
+    # Also filter out warnings in third-party code
```
FreyJo, line 88: Is it possible to filter everything with severity "note", i.e. all those:
https://github.com/acados/acados/security/code-scanning?query=pr%3A982+is%3Aopen+severity%3Anote+
?

Also, is it possible to filter everything from third-party code, i.e. in `external/*`?

FreyJo (commented):
Hi,
thanks for the PR, I briefly read about CodeQL.
It might be useful to have a static analyzer running in CI.
I think nothing severe was found, although it motivated some minor fixes
https://github.com/acados/acados/pull/984
https://github.com/giaf/hpipm/pull/153
There are a lot of uninitialized warnings about HPIPM and BLASFEO code. Maybe @giaf wants to go through them.
In this report the number of false positives seems higher than 20% to me. But I am not sure how this is exactly defined.
E.g. is this a false positive?
https://github.com/acados/acados/security/code-scanning/2403

I hope you can answer/ address the comments.

.github/workflows/codeql-buildscript.sh
b4yuan: https://docs.github.com/en/code-security/codeql-cli/getting-started-with-the-codeql-cli/preparing-your-code-for-codeql-analysis#creating-databases-for-compiled-languages

.github/workflows/codeql.yml
b4yuan: :+1:, of course, any configs can be changed to specific needs

.github/workflows/codeql.yml
b4yuan: Unfortunately I can't see the results under the security tab, but you can filter out any rules you don't like: https://codeql.github.com/codeql-query-help/cpp/

Code from submodules is not included in analysis

giaf:
I had a quick look now and the first two were indeed issues, now fixed https://github.com/giaf/hpipm/commit/1a8df6456727e500c14b5b01808af5e68a822322

I will take time my next working day to check the others.

giaf:
About this https://github.com/acados/acados/security/code-scanning/2403 , the "issue" is the mix of logical and bitwise operator, but in this case the result should work correctly anyway.
In C `&` is a bitwise operator, not logical.
The logical version is `&&`, but this is a "lazy" operator, that evaluates the second condition only after the first is found true, while the bitwise operator evaluates both left and right at the same time.
Performance wise, logical operator avoids the cost of evaluating the right condition but results in multiple branches (so increases chance of branch misprediction), so it is advantageous only if the second condition is very expensive to evaluate (or if the second condition evaluation would fail and break stuff if the first is false, that is the reason for lazy evaluation), while the bitwise results in a single branch.
So as it is written should be the fastest implementation.

giaf:
I'm done checking the report, pushed one commit in HPIPM, dismissed quite a few important ones, plus there is a bunch of notes that should just be ignored.

.github/workflows/codeql.yml
FreyJo: removed cron in https://github.com/acados/acados/pull/988

.github/workflows/codeql.yml
FreyJo: The errors from compiled submodules are included, the line numbers are also shown just not the source code.

FreyJo:
Replaced by https://github.com/acados/acados/pull/988

PR #989, opened 2023-10-31
fix condition for checking number of stages in detect_dims_ocp.m

Ajin2305:
Fixed if condition for checking number of stages in detect_dims_ocp.m

Before fix
image

After fix
image


FreyJo (approved):
I pushed a different fix and test to your branch.
The point is that these options should be allowed to vary per shooting interval.

Ajin2305:
@FreyJo  From the changes you made it is possible for the user to enter a value between 1 and 2 (By mistake), which is why i chose whitelisting 1,2,3,4 like it was done before.

FreyJo:
Indeed the check now doesn't catch this.
However, it was anyway just implemented for ERK.
Plus there are more checks down the road.

Ajin2305:
Understood, all clear

Issue #981, opened 2023-10-26
Invalid MEX-file when running MATLAB examples


KarasevPavel:
Hello! I have a problem with a warning:
```
Warning: The following error was caught while executing 'pendulum_mex_solver' class destructor:
Error using cd
Path must be a text scalar.

Error in pendulum_mex_solver/delete (line 64)
            cd(obj.code_gen_dir);

Error in pendulum_mex_solver (line 49)
            make_mex_pendulum();

Error in acados_ocp (line 209)
            obj.t_ocp = mex_solver();

Error in minimal_example_closed_loop (line 125)
ocp = acados_ocp(ocp_model, ocp_opts); 
> In pendulum_mex_solver (line 49)
In acados_ocp (line 209)
In minimal_example_closed_loop (line 125) 
```
and than an error:
```
Invalid MEX-file '/home/pavel/acados/examples/acados_matlab_octave/getting_started/c_generated_code/acados_mex_create_pendulum.mexa64': libacados_ocp_solver_pendulum.so: cannot open shared object file: No such file or directory

Error in pendulum_mex_solver (line 50)
            obj.C_ocp = acados_mex_create_pendulum();

Error in acados_ocp (line 209)
            obj.t_ocp = mex_solver();

Error in minimal_example_closed_loop (line 125)
ocp = acados_ocp(ocp_model, ocp_opts);
```
when I run examples from the folder `acados_matlab_octave` from this repository

My system:

Processor: intel core i7-9750H
OS: Ubuntu 22.04
MATLAB: 2022a (2021b the same result)
CasADi: 3.4.0 (as mentioned in the **Matlab + Simulink and Octave Interface** guide)

I have done the installation procedure from the links:
1. [Installation](https://docs.acados.org/installation/index.html)
2. [Matlab + Simulink and Octave Interface](https://docs.acados.org/matlab_octave_interface/index.html)

I did some workaround and my temporary solution now is to add a path to the `c_generated_code` folder, which appears in the Matlab running directory to the `LD_LIBRARY_PATH` variable before launching MATLAB. May be we need to add a path to the `c_generated_code` folder in all `env.sh` scripts of all examples? But I think it is not a good idea

Regards, Pavel.

FreyJo:
Hi Pavel,

I am not working with Maltab these days.
But I think that after the latest changes to the Matlab interfaces, see #934, and especially if the model lib is in `c_generated_code`, this folder should indeed be in `LD_LIBRARY_PATH`.

[...]

PR #991, opened 2023-11-07
Rework CasADi requirements in MEX interface

FreyJo:
Following https://github.com/acados/acados/issues/981
- Tested 3.6 on some examples locally
- Only raise Warning on untested CasADi versions

### DIFF
--- a/interfaces/acados_matlab_octave/dump_gnsf_functions.m
+++ b/interfaces/acados_matlab_octave/dump_gnsf_functions.m
@@ -40,7 +40,7 @@ addpath(fullfile(acados_folder, 'external', 'jsonlab'))
 import casadi.*
 
 casadi_version = CasadiMeta.version();
-if ( strcmp(casadi_version(1:3),'3.5')) % require casadi 3.5 +
+if ( strcmp(casadi_version(1:3),'3.5') || strcmp(casadi_version(1:3),'3.6'))
 else % old casadi versions
     error('Please provide CasADi version or 3.5 to ensure compatibility with acados')
 end
--- a/interfaces/acados_matlab_octave/dump_gnsf_functions.m
+++ b/interfaces/acados_matlab_octave/dump_gnsf_functions.m
@@ -40,9 +40,8 @@ addpath(fullfile(acados_folder, 'external', 'jsonlab'))
 import casadi.*
 
 casadi_version = CasadiMeta.version();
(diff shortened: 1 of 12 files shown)


interfaces/acados_matlab_octave/dump_gnsf_functions.m
```diff
@@ -40,7 +40,7 @@ function dump_gnsf_functions(model)
 import casadi.*
 
 casadi_version = CasadiMeta.version();
-if ( strcmp(casadi_version(1:3),'3.5')) % require casadi 3.5 +
+if ( strcmp(casadi_version(1:3),'3.5') || strcmp(casadi_version(1:3),'3.6'))
 else % old casadi versions
     error('Please provide CasADi version or 3.5 to ensure compatibility with acados')
```
sandmaennchen, line 45: The error message is not really a sentence. And this does not use the function `check_casadi_version()` because GNSF is more limiting regarding the casadi version? If so, could we add that to the error message?

interfaces/acados_matlab_octave/gnsf/determine_trivial_gnsf_transcription.m
```diff
@@ -41,11 +41,8 @@
 import casadi.*
 
 casadi_version = CasadiMeta.version();
```
sandmaennchen, line 43: this line can be removed I guess

interfaces/acados_matlab_octave/generate_c_code_nonlinear_least_squares.m
```diff
@@ -36,11 +36,8 @@ function generate_c_code_nonlinear_least_squares( model, opts, target_dir )
 import casadi.*
 
 casadi_version = CasadiMeta.version();
```
sandmaennchen, line 38: this line can be removed?

sandmaennchen (approved):

interfaces/acados_matlab_octave/dump_gnsf_functions.m
FreyJo: This file is just only here to be able to dump the GNSF structure detected system to then use it in Python.
CasADi introduced the serialize functionality in 3.5. afaik.
I'll improve the message though.

interfaces/acados_matlab_octave/gnsf/determine_trivial_gnsf_transcription.m
FreyJo: agreed, removed where not needed.

PR #994, opened 2023-11-14
Seperate nonlinear constraint h into path and initial stage constraints

FreyJo:
Bug description:
when using con_h_expr to set path constraints for an OCP, these constraints are enforced at nodes (0->N-1). However, typically, we also provide an initial state. This results in having two constraints at the initial node.
This results in problems, such as 

Fix:
- enforce path constraints at nodes (1->N-1)
- create a new variable (con_h_expr_0) to set a constraint expression at node 0 explicitly (in case it is desired)
- added a minimal example to reproduce the issue and show usage of the new added variable `ocp_example_h_init_contraints.py`

add constraint expression test example to CMakeLists.txt

update MATLAB/Octave interface to match the new h constraints


sandmaennchen (approved):
that was a bit too much :stuck_out_tongue_winking_eye:

PR #998, opened 2023-11-21
Some interface refactoring

FreyJo:
- Python: detect constraint type (BGH/BGP) from model
- Split `acados_ocp.py` into multiple files for individual individual classes.
- Make `make_consistent()` function of corresponding class
- Formatting
- Moved layout.json files to Matlab, since they are not used anymore in Python interface.


sandmaennchen (approved):
nice!

PR #1007, opened 2023-12-18
extend AcadosMultiphaseOcp

FreyJo:
This is a follow up of https://github.com/acados/acados/pull/1004 removing the limitations described therein.

An `ocp` instance of the class  `AcadosMultiphaseOcp`, now has the additional field `mocp_opts`.
This field is of the class  `AcadosMultiphaseOptions`, "containing options that might be different for each phase", namely `integrator_type`, `collocation_type`, `cost_discretization`.
All of these fields can be either `None`, then the corresponding value from `ocp.solver_options` is used, or a list of length `n_phases` describing the value for this option at each phase.
This especially allows one to define OCPs with varying state dimensions which can be formulated using a transition stage which uses discrete dynamics to map the state representation of one phase to the next phase.
An example of such an OCP formulation is given in `examples/acados_python/mocp_transition_example/main.py`.

### DIFF
--- a/examples/acados_python/mocp_transition_example/main.py
+++ b/examples/acados_python/mocp_transition_example/main.py
@@ -95,6 +95,7 @@ def get_single_integrator_model() -> AcadosModel:
     model.u = u
     model.f_impl_expr = f_impl
     model.f_expl_expr = f_expl
+
     return model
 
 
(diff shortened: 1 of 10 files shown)


sandmaennchen (approved):
amazing! :rocket:

PR #1021, opened 2024-01-16
Update penalty utilities

FreyJo:
- support MX in penalty formulation
- support one sided Huber penalty
- add plots


sandmaennchen (approved):
nice! :rocket:

PR #1023, opened 2024-01-17
Fix multiphase nonlinear constraints

FreyJo:
Fixes https://discourse.acados.org/t/acados-multiphase-python-with-different-constraint-functions-per-phase/1401


sandmaennchen (approved):
:+1:

PR #1024, opened 2024-01-17
Efficient factorization of diagonal outer Hessian in convex-over-nonlinear and nonlinear least-squares costs

sandmaennchen:
* add flags indicating whether outer hessian is diagonal
* if outer hessian is diagonal and `ny > 4`, its diagonal Cholesky factorization `W_chol` is computed manually and stored as `blasfeo_dvec`
* adjust matrix-matrix/matrix-vector multiplications involving `W_chol` accordingly

### DIFF
--- a/acados/ocp_nlp/ocp_nlp_cost_conl.c
+++ b/acados/ocp_nlp/ocp_nlp_cost_conl.c
@@ -40,6 +40,7 @@
 // blasfeo
 #include "blasfeo/include/blasfeo_d_aux.h"
 #include "blasfeo/include/blasfeo_d_blas.h"
+#include "blasfeo/include/blasfeo_common.h"
 // acados
 #include "acados/utils/mem.h"
 
@@ -675,15 +676,16 @@ void ocp_nlp_cost_conl_update_qp_matrices(void *config_, void *dims_, void *mode
                                                 conl_fun_jac_hess_in, conl_fun_jac_hess_type_out, conl_fun_jac_hess_out);
 
         // factorize hessian of outer loss function
-        if (model->psi_hess_is_diag)
+        // TODO: benchmark whether sparse factorization is faster
+        if (model->psi_hess_is_diag & ny > 4)
         {
-            printf("sparse factorization\n");
-            double diag_val = 0.;
+            // TODO move flag to opts and store both W and W_chol as vec if they are diagonal
+            // and then use diagonal matrix matrix multiplication
+            // void blasfeo_dgemm_dn(int m, int n, double alpha, struct blasfeo_dvec *sA, int ai, struct blasfeo_dmat *sB, int bi, int bj, double beta, struct blasfeo_dmat *sC, int ci, int cj, struct blasfeo_dmat *sD, int di, int dj);
             blasfeo_dgese(ny, ny, 0., &memory->W_chol, 0, 0);
             for (int i = 0; i < ny; i++)
             {
-                diag_val = sqrt(blasfeo_dgeex1(&work->W, i, i));
-                blasfeo_dgese(1, 1, diag_val, &memory->W_chol, i, i);
+                BLASFEO_DMATEL(&memory->W_chol, i, i) = sqrt(BLASFEO_DMATEL(&work->W, i, i));
             }
         }
         else
--- a/acados/ocp_nlp/ocp_nlp_cost_conl.c
+++ b/acados/ocp_nlp/ocp_nlp_cost_conl.c
@@ -182,7 +182,6 @@ void *ocp_nlp_cost_conl_model_assign(void *config_, void *dims_, void *raw_memor
     // default initialization
     model->scaling = 1.0;
     model->t = 0.0;
-    model->outer_hess_is_diag = 0;
 
     // assert
     assert((char *) raw_memory + ocp_nlp_cost_conl_model_calculate_size(config_, dims) >= c_ptr);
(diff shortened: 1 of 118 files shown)


acados/ocp_nlp/ocp_nlp_cost_conl.c
```diff
@@ -667,8 +674,22 @@ void ocp_nlp_cost_conl_update_qp_matrices(void *config_, void *dims_, void *mode
...
+        // factorize hessian of outer loss function
+        if (model->psi_hess_is_diag)
+        {
+            printf("sparse factorization\n");
+            double diag_val = 0.;
+            blasfeo_dgese(ny, ny, 0., &memory->W_chol, 0, 0);
+            for (int i = 0; i < ny; i++)
+            {
+                diag_val = sqrt(blasfeo_dgeex1(&work->W, i, i));
+                blasfeo_dgese(1, 1, diag_val, &memory->W_chol, i, i);
```
giaf, line 686: For rather small matrices, this may be slower than just calling the dense Cholesky factorization, since you have two function calls per diagonal element.
I would suggest to perform some benchmark and only switch to the diagonal factorization for sizes where it is beneficial.

I would also suggest to use the macros in here https://github.com/giaf/blasfeo/blob/master/include/blasfeo_common.h#L112 to access the diagonal elements, instead of the two function calls.
The macro works for any BLASFEO version (e.g. is defined differently for each matrix format, e.g. https://github.com/giaf/blasfeo/blob/master/include/blasfeo_common.h#L159 for another one), and it can be used both at the left and right sides of the assignment.

Alternatively, since this is likely a rather common operation, it would be easy to add the corresponding routine directly to BLASFEO, in the auxiliary folder, taking inspiration from existing routines operating on the diagonal of a matrix as e.g. https://github.com/giaf/blasfeo/blob/master/auxiliary/d_aux_lib4.c#L4465 for the panel-major matrix.
A name could be blasfeo_ddiasqrt or something like that.

interfaces/acados_template/acados_template/acados_ocp_cost.py
```diff
@@ -199,6 +211,14 @@ def cost_ext_fun_type_0(self, cost_ext_fun_type_0):
         else:
             raise Exception('Invalid cost_ext_fun_type_0 value, expected numpy array.')
 
+    @outer_hess_is_diag_0.setter
+    def outer_hess_is_diag_0(self, outer_hess_is_diag_0):
+        outer_hess_is_diag_0 = int(outer_hess_is_diag_0)
+        if outer_hess_is_diag_0 == 0 or outer_hess_is_diag_0 == 1:
+            self.__outer_hess_is_diag_0 = outer_hess_is_diag_0
+        else:
+            raise Exception('Invalid phi_hess_diag_0 value, expected bool or int in [0, 1].')
```
FreyJo, line 220: missed rename here

acados/ocp_nlp/ocp_nlp_cost_nls.c
```diff
@@ -286,6 +288,11 @@ int ocp_nlp_cost_nls_model_set(void *config_, void *dims_, void *model_,
         double *scaling_ptr = (double *) value_;
         model->scaling = *scaling_ptr;
     }
+    else if (!strcmp(field, "outer_hess_is_diag"))
+    {
+        int *outer_hess_is_diag_ptr = (int *) value_;
+        model->outer_hess_is_diag = *outer_hess_is_diag_ptr;
```
FreyJo, line 294: I think that for `CONL` cost, the diagonal detection is pretty bullet-proof.
However, for `NLS` / `LS` cost, we allow updating `W` via `cost_set` in the interface.
I think, we should thus check in C, in this function here, if the `W` supplied is actually diagonal and then assign `outer_hess_is_diag` only internally.
In this case, we could actually remove setting it manually for NLS cost in the templates.

acados/sim/sim_irk_integrator.h
```diff
@@ -71,6 +71,7 @@ typedef struct
     external_function_generic *nls_y_fun;  // evaluation nls function
     external_function_generic *conl_cost_fun_jac_hess;
     external_function_generic *conl_cost_fun;
+    int outer_hess_is_diag;
```
FreyJo, line 74: Can we also use a pointer for this, move it to memory, and do it as for `W_chol`?
Then, when doing the change I suggested for NLS, it would automatically update the flag in the integrator.
Moreover, we would not have to set `outer_hess_is_diag` in the dynamics model in the templates at all.

FreyJo (commented):
I went through the code and left some comments.

acados/sim/sim_irk_integrator.h
FreyJo: addressed in [db8b17f](https://github.com/acados/acados/pull/1024/commits/db8b17f92bcefbea96ba18b3c23524010e80b8dc)

FreyJo:
I made a small comparison with these settings:
```
PendulumOcpOptions(
    qp_solver="FULL_CONDENSING_DAQP",
    cost_variant="CONL_LARGE",
    # cost_variant="CONL",
    nlp_solver="SQP_RTI",
    cost_discretization="INTEGRATOR",
    integrator_type="IRK",
    N=50
)
```
For `cost_variant="CONL"`, with ny=nx+nu=5, there is no speedup:
cost_commit_comparison
For `cost_variant="CONL"`, with ny=2*(nx+nu), there is a big speedup:
cost_commit_comparison_conl_large

Thus, I set  `outer_hess_is_diag = 0 if ny <= 4`, in the last commit.

FreyJo (approved):
I think it will make cost integration for medium+ sized problems much faster.
Great collaboration again :chart_with_upwards_trend:

PR #1029, opened 2024-01-26
np = 0 in secondary instances of AcadosOcpSolver.

Confectio:
Hello there,

I am currently trying to parallelise the solving of ocps using multiprocessing (Basically we want to obtain solutions for a batch of states instead of just one state at a time). 
If I understand correctly, since I generated/compiled everything in the main process already and the model should not change, it is enough to create the instances basically like this in each worker process:

ocp = create_AcadosOcp(n_augmented_state_variables, etc.)
acados_ocp_solver = AcadosOcpSolver(
    ocp, json_file='acados_ocp_' + ocp.model.name + '.json', build=False, generate=False)
    
as suggested here by another user in the forum https://discourse.acados.org/t/parallel-execution-in-python-interface-for-monte-carlo-simulations/1278
(Btw is this correct? Do the instances not interfere with each other's solver state then when solving in parallel? Or do I have to generate a new .json for this? I could not find hard evidence on that, yet, so I decided I will just try it out for now.)

However, when I try to use the other created AcadosOcpSolver instances for solving, I get the following error (when trying to set parameters before actualy solving the ocp):
File ".../expert.py", line 445, in set_constraints
    self.solver.set_params_sparse(stage, np.arange(
File ".../acados/interfaces/acados_template/acados_template/acados_ocp_solver.py", line 1565, in set_params_sparse
    raise Exception(f'idx_values_ contains value >= np = {self.acados_ocp.dims.np}')
Exception: idx_values_ contains value >= np = 0

Upon investigation I saw that the dimensions in the acados_ocp.dims member of the "new" (i.e., created in the subprocesses) AcadosOcp instances were not being set.

I talked to @FreyJo earlier today and he said that it is normal that they are not being set, but it might be necessary for the parameter dimension because parameters are handled somehow differently.
Indeed just following his suggestion seems to fix the problem for me.

### DIFF
--- a/acados/ocp_nlp/ocp_nlp_common.c
+++ b/acados/ocp_nlp/ocp_nlp_common.c
@@ -1280,6 +1280,11 @@ void ocp_nlp_opts_set(void *config_, void *opts_, const char *field, void* value
             }
             opts->print_level = *print_level;
         }
+        else if (!strcmp(field, "fixed_hess"))
+        {
+            int* fixed_hess = (int *) value;
+            opts->fixed_hess = *fixed_hess;
+        }
         else
         {
             printf("\nerror: ocp_nlp_opts_set: wrong field: %s\n", field);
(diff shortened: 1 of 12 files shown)


Confectio:
I see, yes that makes sense. Thanks a lot :).

PR #1030, opened 2024-01-26
AcadosOcpSolver: make ocp consistent if generate==False

FreyJo:


FreyJo:
@Confectio would this work for you as well?

Confectio:
@FreyJo Yes, this works for me, too, thanks.

PR #1038, opened 2024-02-16
Cleanup RTI, fix timings and comments

FreyJo:
- fix timings for RTI: subtimers `time_qp*`, `time_reg` now start at 0 when starting preparation and count operations in both phases
- cleanup comments
- make enum for RTI phase
- precompute integrator workspace_size

### DIFF
--- a/acados/ocp_nlp/ocp_nlp_common.c
+++ b/acados/ocp_nlp/ocp_nlp_common.c
@@ -2204,7 +2204,8 @@ void ocp_nlp_approximate_qp_matrices(ocp_nlp_config *config, ocp_nlp_dims *dims,
 
 
 // update QP rhs for SQP (step prim var, abs dual var)
-// evaluate constraints wrt bounds -> allows to update all bounds between preparation and feedback phase.
+// - use cost gradient and dynamics residual from memory
+// - evaluate constraints wrt bounds -> allows to update all bounds between preparation and feedback phase.
 void ocp_nlp_approximate_qp_vectors_sqp(ocp_nlp_config *config,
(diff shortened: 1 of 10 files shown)


acados/ocp_nlp/ocp_nlp_common.c
```diff
@@ -2197,15 +2199,14 @@ void ocp_nlp_approximate_qp_matrices(ocp_nlp_config *config, ocp_nlp_dims *dims,
         struct blasfeo_dvec *ineq_adj =
             config->constraints[i]->memory_get_adj_ptr(mem->constraints[i]);
         blasfeo_dveccp(nv[i], ineq_adj, 0, mem->ineq_adj + i, 0);
-
     }
 }
 
 
 
 // update QP rhs for SQP (step prim var, abs dual var)
```
giaf, line 2207: The rationale behind the old names was to underline that the RHS was set up explicitly for the SQP case, while in principle the matrices routine could be reused to implement other algorithm, e.g. nonlinear IPM.

FreyJo: I am not sure if this routine would be reusable for nonlinear IPM as it was before.
It essentially also sets up and copies things into the `qp_in` structure, which should not be needed in this case.
The main reason for me to move it, was that the QP vectors were not really computed in the feedback phase, they were only copied from what was computed in the preparation.

giaf: Since the `qp_in` in acados is actually the QP in HPIPM, once this is set up it would be possible to implement an IPM by calling one HPIPM IPM iteration (i.e. inequality elimination + Riccati recursion) on this QP.

PR #1040, opened 2024-02-16
Cleanup SQP submodules - only use single evaluation point at a time

FreyJo:
In some places `tmp_ux` and `ux` where used in a mixed way, i.e. `compute_fun` of constraint modules and what was fixed in #1039 
- fixes bug when using globalization and cost computation
- fixes bug in globalization with slacks. Constraint was evaluated with slack values from SQP iterate instead of values from the line search candidate

Implementation:
- Removed all `tmp_*` pointers in SQP submodules
- Always evaluate at `ux`
- When one wants to evaluate at another iterate, these pointers have to be temporarily set to another iterate and afterwards reset to the nominal SQP iterate, as done in `ocp_nlp_evaluate_merit_fun` now.

### DIFF
--- a/acados/ocp_nlp/ocp_nlp_constraints_bgh.c
+++ b/acados/ocp_nlp/ocp_nlp_constraints_bgh.c
@@ -1080,14 +1080,6 @@ void ocp_nlp_constraints_bgh_memory_set_ux_ptr(struct blasfeo_dvec *ux, void *me
 
 
 
-void ocp_nlp_constraints_bgh_memory_set_tmp_ux_ptr(struct blasfeo_dvec *tmp_ux, void *memory_)
-{
-    ocp_nlp_constraints_bgh_memory *memory = memory_;
-
-    memory->tmp_ux = tmp_ux;
-}
-
-
 
 void ocp_nlp_constraints_bgh_memory_set_lam_ptr(struct blasfeo_dvec *lam, void *memory_)
 {
@@ -1096,14 +1088,6 @@ void ocp_nlp_constraints_bgh_memory_set_lam_ptr(struct blasfeo_dvec *lam, void *
     memory->lam = lam;
 }
 
-
-
-void ocp_nlp_constraints_bgh_memory_set_tmp_lam_ptr(struct blasfeo_dvec *tmp_lam, void *memory_)
-{
-    ocp_nlp_constraints_bgh_memory *memory = memory_;
-
-    memory->tmp_lam = tmp_lam;
-}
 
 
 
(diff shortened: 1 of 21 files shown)


acados/ocp_nlp/ocp_nlp_constraints_bgh.c
```diff
@@ -1599,10 +1609,13 @@ void ocp_nlp_constraints_bgh_compute_fun(void *config_, void *dims_, void *model
     blasfeo_daxpy(nb+ng+nh, -1.0, &model->d, nb+ng+nh, &work->tmp_ni, 0, &memory->fun, nb+ng+nh);
 
     // soft
-    blasfeo_dvecad_sp(ns, -1.0, memory->ux, nu+nx, model->idxs, &memory->fun, 0);
-    blasfeo_dvecad_sp(ns, -1.0, memory->ux, nu+nx+ns, model->idxs, &memory->fun, nb+ng+nh);
+    // subtract slacks from softened constraints
+    // fun_i = fun_i - slack_i for i \in I_slacked
+    blasfeo_dvecad_sp(ns, -1.0, ux, nu+nx, model->idxs, &memory->fun, 0);
+    blasfeo_dvecad_sp(ns, -1.0, ux, nu+nx+ns, model->idxs, &memory->fun, nb+ng+nh);
```
FreyJo, line 1615: @giaf Do you agree with this fix?
Before there was an inconsistent evaluation in the constraint module, `tmp_ux` was used for evaluating the function, but the slack values were taken from `ux`.
This results in a failing test, where the solver converges in less iterations than expected.
I will change the test, if you agree.

giaf: Adding the `bool use_tmp_values` to all interfaces goes against the fundamental principles of acados design, see Verschueren2018b  ;)
```
Moreover, it is an important design choice that all modules are identical in
their signature. That way, all modules look similar to the users of acados. For
developers, it should be straightforward to extend acados with another module.
The signature is as follows (in C syntax):
int <solver>(void *config,
<module>_dims *dims,
<module>_in *in,
<module>_out *out,
void *opts,
void *mem,
void *work);
```
In my opinion the right place of this argument is in the opts of each module.

More in general, what is the aim of the change?
As it was implemented, each module could be called with an arbitrary `evaluation point`, it was just needed to properly set the pointers in the memory to where to find it (actually I don't know why in `ocp_nlp_initialize_t_slacks` it is hard copied instead).
With the newly proposed approach you have only two choices, `ux` and `tmp_ux`, since a bool can only take these two values.

About the inconsistent slack evaluation in the constraint module, to what I also see I agree it is inconsistent and it should be fixed as you propose.

FreyJo: Great, thanks for the feedback!
I did a full overhaul to the PR, completely removing the `tmp_*` pointers.

PR #1044, opened 2024-03-04
Fix a few typos regarding the least squares cost formulation (MATLAB examples, Python docs)

josipkh:


FreyJo (approved):
Looks good to me!

Issue #1009, opened 2023-12-19
Missing documentation for initial constraints constr_expr_h_0 (MATLAB)


Jojo-1000:
At least with discrete dynamics (probably also continuous), `constr_expr_h` does not affect the first control vector. Instead, `constr_expr_h_0` needs to be used along with `constr_lh_0` and `constr_uh_0`.

There is no mention of this anywhere in the [problem formulation PDF](https://github.com/acados/acados/blob/master/docs/problem_formulation/problem_formulation_ocp_mex.pdf) nor the linked google sheets. It took me weeks of debugging my OCP to find this. It would be great if the documentation and examples can be updated to include this variable.

After I knew what to look for, the python documentation does mention this option.

### To reproduce
[disc_dyn_example_ocp.m](https://github.com/acados/acados/blob/master/examples/acados_matlab_octave/generic_dyn_disc/disc_dyn_example_ocp.m) can be used to verify this easily. I would argue this example (and others as well) should be updated, since constraints that are not enforced for the first interval are not intended.
- Run solver (until line 222) and print `utraj`:
```matlab
utraj =

  Columns 1 through 13

    1.5288    0.5000    0.5000    0.5000    0.1715   -0.5000   -0.5000   -0.5000   -0.2936    0.5000    0.5000    0.5000    0.0386
    1.1440    0.5000    0.5000    0.5000    0.5000   -0.5000   -0.5000   -0.5000   -0.5000   -0.3021    0.3061    0.5000    0.5000
   -0.9261   -0.5000   -0.3132    0.5000    0.5000    0.5000    0.2025   -0.5000   -0.5000   -0.5000   -0.5000    0.0384    0.2768

  Columns 14 through 20

   -0.2569   -0.2617   -0.1324   -0.0069    0.0696    0.0984    0.0780
    0.2303   -0.0782   -0.1585   -0.0318    0.1025    0.0937   -0.0101
    0.4592    0.5000    0.5000    0.3507   -0.0953   -0.3909   -0.3711
```
Note `utraj(:,1)` exceeds the bounds of +-0.5.
- Add `h_0` constraints to `ocp_model` in line 131:
```matlab
ocp_model.set('constr_expr_h_0', model.expr_h);
ocp_model.set('constr_lh_0', lh);
ocp_model.set('constr_uh_0', uh);
```
- Now `utraj` is constrained as expected:
```matlab
utraj =

  Columns 1 through 13

[...]

PR #1048, opened 2024-03-14
Update MATLAB examples with general initial input constraints

josipkh:
In this PR the following changes are suggested:
- all MATLAB examples are updated as suggested in #1009 
- some examples were modified to remove errors caused by calling outdated structures (e.g., `sim.C_ocp`)
- additional comments were added in some `getting_started` examples
- some models were updated for improved naming consistency (e.g., using `constr_expr_h` instead of `expr_h`)
- examples and tests were updated to work with the changes in variable naming
- a test script that runs all the examples is added (also contains test results on my laptop)
- a few lines were terminated to suppress unnecessary printing


examples/acados_matlab_octave/test/test_all_examples.m
```diff
@@ -0,0 +1,132 @@
...
+%     '../masses_chain_model/example_ocp.m';
+%     '../pendulum_dae/example_closed_loop.m';  % error while reshaping cost.Vz_0
+%     '../pendulum_dae/example_sim.m';
+%     '../pendulum_on_cart_model/example_closed_loop.m';
+%     '../pendulum_on_cart_model/example_ocp.m';
+%     '../pendulum_on_cart_model/example_ocp_custom_hess.m';
+%     '../pendulum_on_cart_model/example_ocp_param_sens.m';  % Unable to resolve the name ocp.t_ocp.eval_param_sens.
+%     '../pendulum_on_cart_model/example_ocp_reg.m';
+%     '../pendulum_on_cart_model/example_sim.m';
+%     '../pendulum_on_cart_model/example_solution_sens_closed_loop.m';  % Unable to resolve the name ocp.t_ocp.eval_param_sens.
```
FreyJo, line 60: Oh, thanks for reporting this! I was not aware!
I will fix this soon!
https://github.com/acados/acados/issues/1049

FreyJo: Maybe we can use this file to run the CI Matlab tests.
Right now we run them via
`cd examples/acados_matlab_octave/test; run_matlab_tests`, see
https://github.com/acados/acados/blob/f3f1a0f75ce889679f782e0720ad944925aedcbc/.github/workflows/full_build.yml#L143

examples/acados_matlab_octave/getting_started/minimal_example_closed_loop.m
```diff
@@ -70,44 +70,51 @@
...
 
 % dynamics
 ocp_model.set('dyn_type', 'explicit');
-ocp_model.set('dyn_expr_f', model.expr_f_expl);
+ocp_model.set('dyn_expr_f', model.dyn_expr_f_expl);
 
 % constraints
 ocp_model.set('constr_type', 'auto');
-ocp_model.set('constr_expr_h', model.expr_h);
+ocp_model.set('constr_expr_h_0', model.constr_expr_h_0);
```
FreyJo, line 107: I do see that this is an important detail.
But, the goal of the examples in `getting_started` is really to make them as easy as possible to not overwhelm new users.
Thus, I would suggest to add the "simpler version" of the pendulum model file to the `getting_started` folder, i.e. ones where the initial node has the same cost and constraints as the terminal one and just have the one with the separate initial shooting node formulation in `pendulum_on_cart_model`.

What do you think?
Could you take care of this?

FreyJo: On a second thought, the differentiation between `h` and `h_0` is pretty important, and I would keep them also in the getting started examples.
Maybe, we can just get rid of the additional `y_0` definition, an additional cost on `x_0` doesnt really hurt and it simplifies the code quite when we omit it.

examples/acados_matlab_octave/getting_started/simulink_example.m
```diff
@@ -26,5 +26,20 @@
...
+end
+
+try
+    cd c_generated_code
+    sim('simulink_model_closed_loop.slx');
+    cd ..
+catch
+    cd ..
+    error('Simulink closed loop example failed')
+end
```
FreyJo, line 45: Does matlab crash otherwise?
I did a similar thing `examples/acados_matlab_octave/test; simulink_test`, which is run on github actions.

examples/acados_matlab_octave/pendulum_on_cart_model/example_ocp.m
```diff
@@ -212,7 +215,7 @@
 ocp_opts.set('qp_solver_cond_ric_alg', qp_solver_cond_ric_alg);
 ocp_opts.set('qp_solver_warm_start', qp_solver_warm_start);
 ocp_opts.set('qp_solver_iter_max', qp_solver_max_iter);
-if (~isempty(strfind(qp_solver, 'partial_condensing')))
+if (contains(qp_solver, 'partial_condensing'))
```
FreyJo, line 218: I think this is not supported in Octave, thats why I would undo this change.

examples/acados_matlab_octave/getting_started/extensive_example_ocp.m
```diff
@@ -296,60 +293,15 @@
...
-    disp(strcat(field, " at stage ", num2str(stage), " = "));
-    ocp.get(field, stage)
-
-    field = 'qp_lbu';
-    disp(strcat(field, " at stage ", num2str(stage), " = "));
-    ocp.get(field, stage)
-    field = 'qp_ubu';
-    disp(strcat(field, " at stage ", num2str(stage), " = "));
-    ocp.get(field, stage)
+for stage = 0:N-1
```
FreyJo, line 299: I would keep it to stage = [0, N-1] here, since the output is a bit much otherwise :)

FreyJo (commented):
Hi Josip,

thanks for the PR!
I have some minor comments, I would be happy if you could respond/address them :)
I also think that we can close https://github.com/acados/acados/issues/1009 after merging this.

FreyJo:
One more comment: did you change anything in the example simulink blocks?
I would prefer not changing them.
If there is a reason to change them, could you make sure that you save them by exporting them to an older Simulink version?
Otherwise people on older Simulink versions can not run the examples.

FreyJo:
About the remaining failing simulink test.
I guess the change in this value is just be because of the minor change in cost.
You can either update the data file or change the tolerance here:
https://github.com/acados/acados/blob/master/examples/acados_matlab_octave/test/simulink_test.m

examples/acados_matlab_octave/test/test_all_examples.m
josipkh: You're welcome :) and thanks for the quick fix.
Yes, I was hoping the script I added could be useful for CI since the examples can be tested directly in that case. It should also work for the existing tests, perhaps with some minor modifications.

examples/acados_matlab_octave/getting_started/simulink_example.m
josipkh: The example works also without this modification, but I added it so it can be run from the testing script I added as well.

examples/acados_matlab_octave/pendulum_on_cart_model/example_ocp.m
josipkh: I understand, sorry for the mistake.

examples/acados_matlab_octave/getting_started/extensive_example_ocp.m
josipkh: Agreed, reverted :)

josipkh:
> One more comment: did you change anything in the example simulink blocks? I would prefer not changing them. If there is a reason to change them, could you make sure that you save them by exporting them to an older Simulink version? Otherwise people on older Simulink versions can not run the examples.

I just had to add the blocks for setting the bounds on the initial input constraints. I also saved the models in the oldest version I could (I think R2015a).

josipkh:
> About the remaining failing simulink test. I guess the change in this value is just be because of the minor change in cost. You can either update the data file or change the tolerance here: https://github.com/acados/acados/blob/master/examples/acados_matlab_octave/test/simulink_test.m

I understand, updating the reference solution seemed more reasonable. I also added a line to the test to prevent a warning about unconnected inputs in Simulink.

examples/acados_matlab_octave/test/test_all_examples.m
FreyJo: Great!
Switching the simulink/Matlab CI to use your script to your test would be nice, but let's do this in another PR.

examples/acados_matlab_octave/getting_started/minimal_example_closed_loop.m
josipkh: What do you think about leaving the `y_0` etc. in the example as it is for now, and possibly removing them when the documentation is updated? I think that it would help clarify the general separation of the initial stage a bit more.

FreyJo: alright, let's do that. I will merge as is then :)

PR #1054, opened 2024-03-18
[WIP] Generalize zoRO custom update function

FreyJo:


interfaces/acados_template/acados_template/zoro_description.py
```diff
@@ -70,6 +70,21 @@ class ZoroDescription:
     idx_uh_t: list = field(default_factory=list)
     idx_lh_e_t: list = field(default_factory=list)
     idx_uh_e_t: list = field(default_factory=list)
+    # Inputs:
+    input_P0_diag: bool = False
+    """Determines if diag(P0) is an input to the custom update function"""
+    input_P0: bool = True
+    """Determines if diag(P0) is an input to the custom update function"""
```
naefjo, line 77: Personally, I think this should be designed such that the user cannot set an impossible state (i.e. both as true). This could be done by only providing one attribute where the user specifies the input P0 matrix traits using a string or an enum.
```suggestion
    input_P0: bool = None
    """Determines if P0 is an input to the function. Allowed options are [None | "diag" | "full_matrix"]"""
```

and in the custom update
```cpp
{%- if zoro_description.input_P0 == "diag" %}
     offset += nx;  // P0_diag
{%- elif zoro_description.input_P0 == "full_matrix" %}
     ...
```
Or is this not doable with the code generator?

FreyJo: Hi @naefjo,
thanks for the quick feedback!
I thought of this combination by adding a sanity check here
https://github.com/acados/acados/pull/1054/files#diff-d6902a8bd481c9f7d9cabcb0656e1fdfe1ba9ab2d764bacf19f065dd6aba5333R108-R109
I think it is nice to have the `input_*` fields to be all of the same type and thus would stick to the Boolean, especially because it is already implemented like this.

naefjo: Alright, makes sense :)

examples/acados_python/zoRO_example/diff_drive/main.py
```diff
@@ -147,6 +147,49 @@ def run_closed_loop_simulation(use_custom_update: bool, n_executions: int = 1):
...
+                               [8.0, 8.0]])
+    spline_seg_length = np.array([0.0, np.pi, 2.0, 2.0, 2.0, np.pi, 2.0, 2.0, 2.0])
+    track_spline = TrackSpline(spline_control_points, spline_seg_length)
+    cfg_path = PathTrackingParam()
+    path_tracking_solver = NominalPathTrackingSolver(track_spline, cfg_path=cfg_path, cfg_traj=cfg_zo)
+    x_init = np.array([0.0, cfg_path._v_s_0])
+    x_e = np.array([1.0, cfg_path._v_s_e])
+    path_tracking_solver.solve(x_init=x_init, x_e=x_e)
+
+    # closed loop mpc
```
sandmaennchen, line 175: that comment seems a bit misleading

interfaces/acados_template/acados_template/custom_update_templates/custom_update_function_zoro_template.in.c
```diff
@@ -692,6 +814,8 @@ static void uncertainty_propagate_and_update(ocp_nlp_solver *solver, ocp_nlp_in
         {%- for it in zoro_description.idx_uh_t %}
         custom_mem->d_uh_tightened[{{it}}] = custom_mem->d_uh[{{it}}]
                         - backoff_scaling_gamma * sqrt(blasfeo_dgeex1(&custom_mem->temp_beta_mat, {{it}}, {{it}}));
+
+        // TODO: make sure uh > lh?
```
sandmaennchen, line 818: maybe add to PR description? seems quite relevant :grimacing:

interfaces/acados_template/acados_template/zoro_description.py
```diff
@@ -70,6 +70,25 @@ class ZoroDescription:
...
```
sandmaennchen, line 77: I guess in the second doc string it should be simply `P0`?

interfaces/acados_template/acados_template/custom_update_templates/custom_update_function_zoro_template.in.c
FreyJo: This is no new requirement. I removed the comment.
It is of course a problem, if tightened constraints lead to an empty feasible set.
In this case, the next OCP solver call should return an infeasibility status, which I guess is the expected behavior.

sandmaennchen (approved):

PR #1071, opened 2024-04-12
DDP and Feasibility-Projected DDP implementation

david0oo:
This PR adds Differential Dynamic Programming (DDP) as a new `nlp_solver_type` to acados.
Many of the implemented features are based on the paper "Fast Generation of Feasible Trajectories in Direct Optimal Control" by David Kiessling, Katrin Baumgaertner, Jonathan Frey, Wilm Decré, Jan Swevers, and Moritz Diehl: https://arxiv.org/abs/2403.10115
- The method introduced in the paper solves feasibility problem OCPs, which are generated from standard OCPs. In particular the dynamics are kept as constraints and the constraint violation of the other constraints is used as l2-penalties in the objective function. The objective of the original OCP is discarded.
- The method is globally convergent with a line search using only the objective function since DDP keeps all iterates dynamically feasible
- An adaptive Levenberg-Marquardt regularization + GGN Hessian is used in to guarantee positive-definiteness of the Hessian matrix. The regularization allows for local quadratic convergence

The following features were implemented:
- The DDP method can be used to solve unconstrained OCPs which are not necessarily feasibility problems
- A method was added that transforms the OCP into an unconstrained OCP feasibility problem as in the paper
- The adaptive Levenberg-Marquardt strategy was added and it can be called as a new option. The strategy is suitable for (nonlinear) least-squares problems, where local quadratic convergence can be achieved for zero-residual problems. At the moment the adaptive Levenberg-Marquardt feature is only available for DDP. For details in usage please see the examples added with this pull request
- The new DDP implemenation is fully interfaced in Python. The Python interface checks that the OCP is unconstrained in case DDP is used.
- In order to obtain an unconstrained OCP formulation, the convenience function `AcadosOcp.translate_to_feasibility_problem()` is added.
- Four test problems were added: One problems solves a QP for which SQP and DDP should only take 1 iteration, the other test problems are testing the feasibility problem feature.
- added input `contraint_type` to  `formulate_constraint_as_L2_penalty()` which supports `"initial", "path", "terminal"`.

Limitations:
- So far DDP is only tested for explicit integrators. For implicit integrators additional precautions need to be taken in the line search.
- Adding to the previous point: problems with algebraic variables should be tested.
- In the QP tolerances setter: The accuracy should be set to the minimum of the current QP accuracy and the one specified.


acados/ocp_nlp/ocp_nlp_ddp.c
```diff
@@ -0,0 +1,1514 @@
...
+            else if (nlp_opts->globalization == MERIT_BACKTRACKING)
+            {
+                // do backtracking line search on objective function
+                linesearch_success = ocp_nlp_ddp_backtracking_line_search(config, dims, nlp_in, nlp_out, mem, work, opts);
+                evaluate_cost = false; // since the cost was already evaluated in the line search
+            }
+
+            mem->stat[mem->stat_n*(ddp_iter+1)+6] = mem->alpha;
+            // Copy new iterate to nlp_out
+            if (linesearch_success == 0)
```
FreyJo, line 1037: Here, I think the status, or more precisely the variable `linesearch_success` is a bit confusing. I would have expected a boolean meaning, i.e. 1 means success.
Of course, for main functions in C, 0 means success, but since `ocp_nlp_ddp_backtracking_line_search` is not a main function, I would prefer to change it to a boolean meaning.

david0oo: Sure :)

FreyJo (approved):
Looks good to me now!
Great job! :100:

PR #1074, opened 2024-04-15
Matlab Interface: Add ztraj to sfunction output

leonardocecchin:
Added an output port in the acados_ocp S-function block for the algebraic states' trajectory, `ztraj`, equivalent to `xtraj`.


FreyJo (approved):
Looks good! Thanks!

Issue #1025, opened 2024-01-18
Error when including acados within CMAKE project


hisermann:
Hi everyone, 
I installed acados using the following commands:
```
mkdir build
cd build
cmake -DACADOS_WITH_QPOASES=ON -DACADOS_WITH_OSQP=ON -DACADOS_INSTALL_DIR=/opt/acados ..
sudo make install -j4
```
Also, I have added `/opt/acados/lib` to the library path.
Now I am trying to include acados within my own CMAKE project, using the following lines:
(I am using ROS 2 with colcon build)
```
find_package(acados REQUIRED)
ament_target_dependencies(mitcontroller acados)
```
Unfortunately, I now get the following errors:
```
CMake Warning at /opt/acados/cmake/acadosConfig.cmake:73 (find_package):
  By not providing "FindOpenBLAS.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "OpenBLAS",
  but CMake did not find one.
```
```
CMake Warning at /opt/acados/cmake/acadosConfig.cmake:79 (find_package):
  By not providing "FindFortranLibs.cmake" in CMAKE_MODULE_PATH this project
  has asked CMake to find a package configuration file provided by
  "FortranLibs", but CMake did not find one.
```
```
CMake Error at CMakeLists.txt:25 (find_package):
  Found package configuration file:

    /opt/acados/cmake/acadosConfig.cmake

  but it set acados_FOUND to FALSE so package "acados" is considered to be
  NOT FOUND.  Reason given by package:

  The following imported targets are referenced, but are missing: osqp::osqp
```

We figured out that this could be solved by adding the following to `acados/CMakeLists.txt`:
```
install(DIRECTORY ${PROJECT_SOURCE_DIR}/cmake/
        DESTINATION cmake
        FILES_MATCHING PATTERN "Find*.cmake")
```
and the following to `cmake/acadosConfig.cmake.in`:
```
if(ACADOS_WITH_OSQP)
    find_package(osqp)
endif()
```

Further linker and include errors could be resolved by manually adding the include dirs and lib dirs to our CMAKELists.txt:

```
target_link_libraries(mitcontroller /opt/acados/lib/libblasfeo.so)
target_link_libraries(mitcontroller /opt/acados/lib/libhpipm.so)
target_link_libraries(mitcontroller /opt/acados/lib/libacados.so)

target_include_directories(
        mitcontroller
[...]

PR #1081, opened 2024-04-23
Fix acados cmake config

FranekStark:
Although most of the issues with using acados in another cmake project (https://github.com/acados/acados/issues/1025) have been fixed with https://github.com/acados/acados/pull/1065, there are still a few lines missing in acadosConfig.cmake.in.


FreyJo (approved):
Looks good, thanks!

PR #1085, opened 2024-04-24
Modularize CI workflow

AkashJohnSubash:
- Seperate acados `core_build` from interface installation and tests for improved feedback on what fails
- `core_build` takes < 2 min
- Parallelize Python, MATLAB, Octave, Simulink jobs. (Dependent only on `core_build` job) < 5 min
- ctest parallelization using `-j, -parallel` flags is constrained by forcing serial execution for specific tests in `\interface\CmakeLists.txt `


.github/workflows/full_build.yml
```diff
@@ -13,7 +13,6 @@ on:
 env:
   # Customize the CMake build type here (Release, Debug, RelWithDebInfo, etc.)
   BUILD_TYPE: Release
-  ACADOS_UNIT_TESTS: ON
```
FreyJo, line 16: somehow this is still used in the `cmake` commands. I suggest to remove there.

.github/workflows/full_build.yml
```diff
@@ -36,118 +35,244 @@ jobs:
...
+    - name: Store shared libs (/lib)
+      uses: actions/upload-artifact@v4
+      with:
+        name: lib
+        if-no-files-found: error
+        path: ${{runner.workspace}}/acados/lib/
+        compression-level: 0 #TODO test time difference
+        overwrite: true
+
+    - name: Store build scripts (/build)
```
FreyJo, line 69: Is storing the `build` folder needed?

.github/workflows/full_build.yml
```diff
@@ -36,118 +35,244 @@ jobs:
...
+
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        submodules: 'recursive'
+
+    - uses: actions/download-artifact@v4
+      with:
+        path: ${{runner.workspace}}/acados
+
```
FreyJo, line 133: I see that until here, all interface jobs are the same.
Is this duplication what you said could be avoided with the composite action?
If it is just these lines, I am not sure if it is worth it.

.github/workflows/full_build.yml
```diff
@@ -36,118 +35,244 @@ jobs:
...
+      # See https://cmake.org/cmake/help/latest/manual/ctest.1.html for more detail
+      run: ${{runner.workspace}}/acados/.github/linux/run_tests.sh
 
-    - name: Prepare Octave
+  MATLAB_interface:
+    needs: core_build
+    runs-on: ubuntu-20.04
+
+    steps:
+    - uses: actions/checkout@v2
```
FreyJo, line 126: Maybe update to v4, if it works out of the box.

.github/workflows/full_build.yml
```diff
@@ -36,118 +35,244 @@ jobs:
...
+      with:
+        submodules: 'recursive'
```
FreyJo, line 128: I think this should not be needed for Simulink, Octave and Matlab.
For the Python test, there is a submodule with `test_data` which is some results are compared against.

.github/workflows/full_build.yml
```diff
@@ -36,118 +35,244 @@ jobs:
...
       run: |
-        sudo apt-get update
-        sudo apt-get install liboctave-dev -y --fix-missing
-        octave --version
-        ./../.github/linux/install_casadi_octave.sh
-        echo "OCTAVE_PATH=$OCTAVE_PATH:$(pwd)/casadi-octave" >> $GITHUB_ENV
+        sudo chmod +x ${{runner.workspace}}/acados/.github/linux/install_casadi_matlab.sh
+        ${{runner.workspace}}/acados/.github/linux/install_casadi_matlab.sh
+        echo "MATLABPATH=$MATLABPATH:${{runner.workspace}}/acados/interfaces/acados_matlab_octave:${{runner.workspace}}/acados/interfaces/acados_matlab_octave/acados_template_mex:${{runner.workspace}}/acados/external/casadi-matlab" >> $GITHUB_ENV
+        echo "ACADOS_INSTALL_DIR=${{runner.workspace}}/acados" >> $GITHUB_ENV
```
FreyJo, line 141: The duplication of setting up the environment variables is not so nice.
And the fact that this is part of some other step.
Maybe we could have one script `export_paths.sh` that is executed in each job in one step.
If this sets some paths that are not needed for the particular job, like `MATLABPATH`, it would be fine too.
Just not sure how to get the `{{runner.workspace}}` in there.

interfaces/CMakeLists.txt
```diff
@@ -206,7 +211,7 @@ add_test(NAME python_multiphase_nonlinear_constraints
         COMMAND "${CMAKE_COMMAND}" -E chdir ${PROJECT_SOURCE_DIR}/examples/acados_python/multiphase_nonlinear_constraints
         python main.py)
 
-# qpDUNES example
+# qpDUNES example TODO couple with pendulum examples
```
FreyJo, line 214: why not just add it in the list below?

FreyJo (commented):
I added some comments and pushed some small changes.

.github/workflows/full_build.yml
AkashJohnSubash: yes, the ctest cmake file is defined there

.github/workflows/full_build.yml
AkashJohnSubash: I just pass a {{runner.workspace}} as an argument to the bash script.

.github/workflows/full_build.yml
AkashJohnSubash: true, but the idea was to combine checkout, download-artifacts, and export-paths into a compose action

interfaces/CMakeLists.txt
AkashJohnSubash: Will cause the ctest to report a failure, if the module is not found when built with ACADOS_WITH_QPDUNES=OFF

.github/workflows/full_build.yml
AkashJohnSubash: Checking out specific submodules seems to be not yet supported by the _checkout@v4_ action

FreyJo (approved):
Polished it a bit more, merging now.
Good job! :+1:

PR #1087, opened 2024-04-29
Batch integrator and OCP solver for parallel solves

sandmaennchen:
This PR implements a parallelizable batch solve for acados integrators and OCP solvers. The batch solve is interfaced from python with `AcadosOcpSolverBatch` and `AcadosSimSolverBatch`. Parallelization via openmp can be switched on/off via the option `with_parallel_batch_solve` which has been added to the solver options.

These classes hold `N_batch` identical solvers and provide a parallelizable solve function. All other operations have to be called individually for each of the solvers via their standard interface.

Limitations:
* number of openmp threads cannot be set at the moment.

### DIFF
--- a/examples/acados_python/getting_started/minimal_example_sim.py
+++ b/examples/acados_python/getting_started/minimal_example_sim.py
@@ -36,18 +36,13 @@ import numpy as np
 
 
 def main():
+
     sim = AcadosSim()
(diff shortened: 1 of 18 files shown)


FreyJo (approved):
Great! :zap:

PR #1089, opened 2024-04-30
Follow up to #1087

sandmaennchen:
Apparently, shared libraries with and without openmp cannot easily be combined.  If the acados batch solvers introduced in #1087  should be used, it is thus recommended to compile the acados shared library with openmp and set the number of threads to 1.
This problem seems to be related to:
* https://github.com/JuliaLang/julia/issues/10938
* https://github.com/casadi/casadi/issues/2447

This PR implements corresponding warnings in `AcadosOcpBatchSolver` and `AcadosSimBatchSolver` and removes the examples from the tests (as we do not want to compile acados with openmp for all tests).


FreyJo (approved):
Great! :1st_place_medal:

PR #1118, opened 2024-05-21
Add Anderson acceleration to SQP

FreyJo:


david0oo:
@FreyJo this PR can be deleted

PR #1126, opened 2024-06-03
MATLAB interface - control rates example

josipkh:
This PR contributes an example with constraints and cost on the control input rates.
It is a slightly modified version of the code originally posted [here](https://discourse.acados.org/t/penalties-and-constraints-for-control-input-rates-via-augmented-states/1070/5).
Also, the script which can be used to test all examples automatically was updated with the recent fixes in `acados`.


FreyJo (approved):
Really nice reference! Thanks :)

PR #1133, opened 2024-06-13
Setter for updating `max_iter` in Python

zanellia:
Allow to set `max_iter` in Python.
Check whether the value requested is higher than the one used to allocate memory and raise an Exception in this case.

Minor changes:
- Make `solver_options` of `AcadosOcpSolver` private
- `np.NaN` -> `np.nan`

### DIFF
--- a/interfaces/acados_template/acados_template/acados_ocp_solver.py
+++ b/interfaces/acados_template/acados_template/acados_ocp_solver.py
@@ -258,6 +258,11 @@ class AcadosOcpSolver:
         self.time_solution_sens_solve = 0.0
         self.time_solution_sens_lin = 0.0
 
+        # store nlp_solver_max_iter value
+        # (needed to raise a warning whenever max_iter is set to a higher value,
+        # since that could cause memory violations when writing to stats)
+        self.__nlp_solver_max_iter = self.solver_options['nlp_solver_max_iter']
+
         # gettable fields
         self.__qp_dynamics_fields = ['A', 'B', 'b']
         self.__qp_cost_fields = ['Q', 'R', 'S', 'q', 'r']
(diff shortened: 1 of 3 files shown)


interfaces/acados_template/acados_template/acados_ocp_solver.py
```diff
@@ -258,6 +258,11 @@ def __init__(self, acados_ocp: Union[AcadosOcp, AcadosMultiphaseOcp], json_file=
         self.time_solution_sens_solve = 0.0
         self.time_solution_sens_lin = 0.0
 
+        # store nlp_solver_max_iter value
+        # (needed to raise a warning whenever max_iter is set to a higher value,
+        # since that could cause memory violations when writing to stats)
+        self.__nlp_solver_max_iter = self.solver_options['nlp_solver_max_iter']
```
FreyJo, line 264: I would just compare to `self.solver_options['nlp_solver_max_iter']` directly below.
It would make sense to make `self.solver_options` private directly to not allow changing them from outside.

FreyJo: > since that could cause memory violations when writing to stats

I think there will be no memory violations, it will just not store stats of iterations that exceed the initial number.
However, I don't think we should expose this behavior in any way.
So on a second thought, I would prefer raising an Exception below.

FreyJo (commented):
Happy to see you contributing to acados again :)
I added some comments.

interfaces/acados_template/acados_template/acados_ocp_solver.py
zanellia: I have made `solver_options` private 4cc423a7fb3 and now the code raises an Exception rather than printing a warning 5be544f80d14.

However, the tests are failing due to an error which doesn't seem to be related to the present MR:

`File "/home/runner/work/acados/acados/examples/acados_python/tests/reset_test.py", line 136, in main
    ocp_solver.set(i, 'x', np.NaN * np.ones((nx,)))
  File "/home/runner/work/acados/acados/acadosenv/lib/python3.10/site-packages/numpy/__init__.py", line 397, in __getattr__
    raise AttributeError(
AttributeError: `np.NaN` was removed in the NumPy 2.0 release. Use `np.nan` instead.. Did you mean: 'nan'?`

FreyJo (approved):
I reran the CI on the latest commit on `master` and it also failed with the same message, https://github.com/acados/acados/actions/runs/9520749546
I pushed a fix to your branch and will merge once it passed :)

zanellia:
awesome, thanks a lot @FreyJo! I had seen that the tests were failing due to the NaN-related issue with mismatching NumPy versions and commented, but I am on a train with a bad connection and somehow the comment got lost. Anyway, thanks for fixing it!

Issue #1130, opened 2024-06-07
cmakelists does not respect CMAKE_INSTALL_PREFIX


asherikov:
One of CMAKE_INSTALL_PREFIX use cases is to provide it via command line or a toolchain, which is broken by https://github.com/acados/acados/blob/master/CMakeLists.txt#L84C5-L84C25. Using a custom parameter ACADOS_INSTALL_DIR to override the default one requires package specific workarounds downstream.

One option is to use https://cmake.org/cmake/help/latest/variable/CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT.html instead, it would require bumping required cmake version to 3.7.1, which seems reasonable. If there are no objections I can do an MR.

FreyJo:
I think requiring CMake >= 3.7.1 is fine, apparently it was released in 2016.

asherikov:
Opened a PR -> https://github.com/acados/acados/pull/1138, closing this.

PR #1138, opened 2024-06-18
CMakeLists: do not override CMAKE_INSTALL_PREFIX

asherikov:
Described in https://github.com/acados/acados/issues/1130

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,7 +29,7 @@
 #
 
 
-cmake_minimum_required(VERSION 3.1)
+cmake_minimum_required(VERSION 3.7.1)
 
 if(CMAKE_MAKE_PROGRAM)
     file(TO_CMAKE_PATH ${CMAKE_MAKE_PROGRAM} CMAKE_MAKE_PROGRAM)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -44,7 +44,18 @@ set(CMAKE_POSITION_INDEPENDENT_CODE TRUE)
...
+if(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
+  if(DEFINED ACADOS_INSTALL_DIR)
+    message(WARNING "ACADOS_INSTALL_DIR variable is deprecated, use CMAKE_INSTALL_PREFIX instead.")
+    set_property(CACHE CMAKE_INSTALL_PREFIX PROPERTY VALUE "${ACADOS_INSTALL_DIR}")
+  else()
+    set_property(CACHE CMAKE_INSTALL_PREFIX PROPERTY VALUE "${PROJECT_SOURCE_DIR}")
+  endif()
+else()
+  if(DEFINED ACADOS_INSTALL_DIR)
+    message(WARNING "Deprecated ACADOS_INSTALL_DIR variable has been overriden by CMAKE_INSTALL_PREFIX: ${CMAKE_INSTALL_PREFIX}")
```
FreyJo, line 56: Can we make `ACADOS_INSTALL_DIR` have priority over `CMAKE_INSTALL_PREFIX` in a first step?
If I pull your branch and do my usual installation, I get:

```
CMake Warning at CMakeLists.txt:56 (message):
  Deprecated ACADOS_INSTALL_DIR variable has been overriden by
  CMAKE_INSTALL_PREFIX: /usr/local
```
This will result in an error for all users that use the default installation process.

I would rather require people who use `CMAKE_INSTALL_PREFIX` directly to clean their cmake cache, since this should be less acados users.
Does this make sense?

Also, there is a typo in `overriden`.

asherikov: ok, new version should behave the way you want.

FreyJo (approved):
I pushed minor modifications and would merge as is now.
Do you agree with them?

asherikov:
looks good, just a small indentation issue at line 58

asherikov:
thanks

Issue #1140, opened 2024-06-19
Update Problem Formulation


sandmaennchen:
The problem formulation 

https://github.com/acados/acados/blob/master/docs/problem_formulation/problem_formulation_ocp_mex.pdf

is a bit outdated:

* nonlinear constraints may depend on algebraic variables
* convex-over-nonlinear cost is missing

Issue #565, opened 2020-03-23
Matlab: Avoid `source env.sh`


FreyJo:
It was suggested in the forum to set the environment variables in the MATLAB/Octave rc file, such that 
`source env.sh` can be avoided.
https://discourse.acados.org/t/matlab-mex-more-elegant-way-to-setup-env-sh/62/4

I found there is a similar file in Octave
https://octave.org/doc/v4.2.1/Startup-Files.html

Should be tested across multiple OS, Matlab, Octave versions.

giaf:
I would leave both options.

Personally I like the env.sh because I like to have everything local to the project folder, for development it makes easier and safer when I have multiple versions of acados/hpipm/blasfeo on the system.

But I agree that for an end user, having to set this once forever in some config file may be more convenient. Instructions on alternatives to env.sh should definitely be added in the READMEs.

sandmaennchen:
The documentation now includes a remark on how to avoid `source env.sh`.

https://github.com/acados/acados/blob/master/docs/matlab_octave_interface/index.md#export-environment-variables

PR #1141, opened 2024-06-24
Documentation update

josipkh:
This PR addresses some of the points listed in #1140 (and others) and updates the documentation in several places.


josipkh:
Hi Jonathan,
thanks for checking!
Yes, I would like to finish updating the problem formulation pdf with a few more things related to the initial state...
I was hoping to do it by now, but some other things came up so I plan to finish it by the end of the week.
I will mark the PR as ready then, hope that's ok :)

josipkh:
Hello,
I made a few more changes and marked the PR as ready.
There is two more things I wanted to ask about:
- are there any changes you would suggest for the [MATLAB examples readme](https://github.com/acados/acados/blob/master/examples/acados_matlab_octave/README.md)? maybe even removing it altogether (Python interface doesn't have one)
- perhaps adding a note about the environment setup as suggested in #565 would be nice, but I am not sure where is the best place to do so (I mainly use Windows, where I edit MATLAB's `startup.m` script)

FreyJo:
Hi Josip,
thanks for raising those points.
I addressed them in the last two commits here.
If you agree, I would merge this now.

josipkh:
Thanks for the quick update!
Feel free to merge from my side :)

PR #1148, opened 2024-07-09
Remove `t` slacks from `ocp_nlp_out`

FreyJo:
The slacks `t` that were used to slack the inequality constraint values are actually not needed in an NLP solver and are thus removed in this PR.
Instead of using `t`, we now directly evaluate inequality residual using the function evaluations.
This results in consistent inequality residuals and will help future work on globalization.
Related interface functionality such as getters and setters for `t` and `initialize_t_slacks` have been removed.

### DIFF
--- a/interfaces/acados_template/acados_template/acados_ocp_solver.py
+++ b/interfaces/acados_template/acados_template/acados_ocp_solver.py
@@ -1392,8 +1392,7 @@ class AcadosOcpSolver:
 
         Note:
         - additional supported fields are ['P', 'K', 'Lr'], which can be extracted form QP solver PARTIAL_CONDENSING_HPIPM.
-        - for PARTIAL_CONDENSING_* QP solvers, the following additional fields are available:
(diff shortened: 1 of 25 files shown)


interfaces/acados_template/acados_template/acados_ocp_solver.py
```diff
@@ -717,7 +717,7 @@ def get(self, stage_: int, field_: str):
         Get the last solution of the solver:
 
             :param stage: integer corresponding to shooting node
-            :param field: string in ['x', 'u', 'z', 'pi', 'lam', 't', 'sl', 'su', 'sens_u', 'sens_x']
+            :param field: string in ['x', 'u', 'z', 'pi', 'lam', 'sl', 'su', 'sens_u', 'sens_x']
 
             .. note:: regarding lam, t: \n
```
sandmaennchen, line 722: here you missed one

sandmaennchen (approved):

PR #1149, opened 2024-07-10
Remove main mex template

FreyJo:
These files are outdated, there were used to test the template based solver in Maltab without really interacting with it, before the MEX interface was completely ported to the template backend in https://github.com/acados/acados/commit/bb9f1b299d5d62cd0aaf515544ebdc520cbaba2b


sandmaennchen (approved):
:broom: :heart:

PR #1152, opened 2024-07-12
Fix small typo in Makefile

josipkh:
Hi, I noticed two small typos in the Makefile.
Also, I was wondering if something similar to [this section](https://github.com/acados/acados/blob/master/Makefile#L71-L73) should be added for [DAQP](https://github.com/acados/acados/blob/master/acados/dense_qp/dense_qp_daqp.c)?

EDIT: I see now that DAQP is not included in the Makefile, not sure if it's relevant in that case.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -137,13 +137,13 @@ endif
 ifeq ($(ACADOS_WITH_QORE), 1)
 STATIC_DEPS += qore_static
 CLEAN_DEPS += qore_clean
-LINK_FLAG_QPDUNES = -lqore
+LINK_FLAG_QORE = -lqore
 endif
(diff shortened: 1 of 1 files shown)


FreyJo (approved):
Thanks for fixing the typos.
Indeed DAQP is not included in Make yet.
Also, the make build system is much less tested and we recommend using CMake, see e.g. https://discourse.acados.org/t/a-compile-error-when-run-make-examples-c/1640/2?u=freyjo

Shall I merge?

josipkh:
I understand, thanks for the clarification.
Feel free to merge as far as I'm concerned :)

PR #1155, opened 2024-07-15
Fix QP status issue

FreyJo:
This PR addresses an issue which caused the SQP solver to always return status 4, if it was returned once.
This was introduced in https://github.com/acados/acados/pull/1057
Added a test to check that the OCP solver can recover from such a situation.


sandmaennchen (approved):
nice!

PR #1160, opened 2024-07-17
Align python and MATLAB classes for problem formulation

sandmaennchen:
This PR is a first step towards aligning the MATLAB interface with the python interface.

python changes:

- new attributes `dyn_impl_dae_fun_jac`, `dyn_impl_dae_jac`, `dyn_impl_dae_fun` in the `AcadosModel` for generic external implicit dynamics
- new attribute `nlp_solver_warm_start_first_qp` in `AcadosOptions`

MATLAB changes:
- Renaming of internal classes `acados_<...>_json` to match corresponding python class `Acados...`
- lots of option defaults added which are already available in python
- remove `rti_phase` in formulation object, as it only makes sense to change this online


FreyJo (approved):
Great to have this :+1:

PR #1167, opened 2024-07-24
Merit backtracking cleanup

FreyJo:
- split into functions
- separate function for preliminary line search before SOC
- reset merit function weights when doing SOC. This is the only change affecting the behavior of the algorithm in this PR.
- adjusted test to account for point above.

### DIFF
--- a/acados/ocp_nlp/ocp_nlp_sqp.c
+++ b/acados/ocp_nlp/ocp_nlp_sqp.c
@@ -537,6 +537,105 @@ static void ocp_nlp_sqp_dump_qp_out_to_file(ocp_qp_out *qp_out, int sqp_iter, in
 #endif
 
 
+static double ocp_nlp_get_violation_inf_norm(ocp_nlp_config *config, ocp_nlp_dims *dims,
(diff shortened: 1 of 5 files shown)


acados/ocp_nlp/ocp_nlp_sqp.c
```diff
@@ -537,6 +537,148 @@ static void ocp_nlp_sqp_dump_qp_out_to_file(ocp_qp_out *qp_out, int sqp_iter, in
...
+    double violation_step = ocp_nlp_get_violation_inf_norm(config, dims, in, out, opts, mem, work);
+    if (opts->print_level > 0)
+    {
+        printf("\npreliminary line_search: merit0 %e, merit1 %e; viol_current %e, viol_step %e\n", merit_fun0, merit_fun1, violation_current, violation_step);
+    }
+
+    if (isnan(merit_fun1) || isinf(merit_fun1))
+    {
+        // do nothing and continue with normal line search, i.e. step reduction
+        return ACADOS_NAN_DETECTED;
```
sandmaennchen, line 661: I think in this case we also want to reset the merit function weights

sandmaennchen (approved):

PR #1170, opened 2024-07-26
MATLAB: Code generation of CasADi functions based on `AcadosOcp`/`AcadosSim`

sandmaennchen:
This PR changes the code generation of CasADi functions in MATLAB such that it is solely based on the OCP description used of the template interface, i.e. `AcadosOcp`. Similar for integrators where only `AcadosSim` is used for code generation.

### DIFF
--- a/interfaces/acados_matlab_octave/create_consistent_empty_fields.m
+++ b/interfaces/acados_matlab_octave/create_consistent_empty_fields.m
@@ -29,31 +29,29 @@
 
 %
 
-function model = create_consistent_empty_fields(model, opts)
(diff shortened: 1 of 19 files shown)


interfaces/acados_matlab_octave/create_consistent_empty_fields.m
```diff
@@ -29,31 +29,29 @@
 
 %
 
-function model = create_consistent_empty_fields(model, opts)
+function model = create_consistent_empty_fields(model)
```
FreyJo, line 32: Moving this kind of functionality to the template classes is the next step I guess.

PR #1172, opened 2024-07-26
Matlab QP conditioning

FreyJo:
- SQP & DDP: move regularize before check_termination, such that we can get the QP at the stationary point that is actually solved, if we exit with success.
- MEX: add functions `dump_last_qp_to_json`, `qp_diagnostics`

### DIFF
--- a/acados/ocp_nlp/ocp_nlp_ddp.c
+++ b/acados/ocp_nlp/ocp_nlp_ddp.c
@@ -807,6 +807,7 @@ int ocp_nlp_ddp(void *config_, void *dims_, void *nlp_in_, void *nlp_out_,
         reg_param_memory = nlp_opts->levenberg_marquardt;
 
         // regularize Hessian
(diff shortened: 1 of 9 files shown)


acados/ocp_nlp/ocp_nlp_ddp.c
```diff
@@ -806,6 +806,12 @@ int ocp_nlp_ddp(void *config_, void *dims_, void *nlp_in_, void *nlp_out_,
         }
         reg_param_memory = nlp_opts->levenberg_marquardt;
 
+        // regularize Hessian
```
sandmaennchen, line 809: maybe add note from SQP here as well

sandmaennchen (approved):

acados/ocp_nlp/ocp_nlp_ddp.c
FreyJo: done

Issue #650, opened 2020-10-15
Handling One-Sided Constraints


FreyJo:
The `acados` problem description is based on two sided constraints.
Adding one-sided constraints in this framework is nontrivial and can lead to ill conditioning, especially in the QP solver.
The most user-friendly way would be to allow the user to set them to `Inf` and convert them internally.

- for HPIPM, recently constraint `mask` was added to deal with one sided constraints.
They can be changed after creation and dont change the memory needed.
https://github.com/giaf/hpipm/blob/b86f426a2de7bd8fdf35aa63c9f6b59146d8eae5/Changelog.txt#L67
https://github.com/giaf/hpipm/blob/b86f426a2de7bd8fdf35aa63c9f6b59146d8eae5/include/hpipm_d_ocp_qp.h#L116
- for other QP solvers, the plan would be to replace the `Inf` with some `ACADOS_INF` value that has to be defined and should be tested on some example

That is just a rough plan of what should be done, any comments welcome!

For now, we recommend to use not too large numbers for unbounded sides of the constraint to avoid ill-conditioning.

FreyJo:
Preliminary support for this is added in https://github.com/acados/acados/pull/1174
Follow-ups include:
- more testing, feedback from the community
- Support in QP solvers other than HPIPM
   - [x] DAQP
   - [ ] qpOASES
   - [ ] qpDUNES
   - [ ] OSQP
   - [ ] Clarabel

sandmaennchen:
Further point to consider:

- [ ] Multiplier associated with masked bounds should always be zero. In particular, this needs to be ensured whenever multipliers are initialized and whenever the bounds are updated (potentially masking previously active constraints)

sandmaennchen:
I added a note in the updated problem description, https://github.com/acados/acados/pull/1221, which should be updated when one-sided constraints are fully supported.

PR #1174, opened 2024-07-29
One sided constraint support for HPIPM solvers

FreyJo:
- add `ACADOS_INFTY` in Python and C, `get_acados_infty` in Matlab.
- If any constraints have bounds `ACADOS_INFTY`, the constraint is marked as not relevant in the QP representation object `ocp_qp`, by setting the corresponding `d_mask` values to `0`.
- The underlying QP solver interface should make use of this mask.
- The HPIPM solvers do make use of this natively.
- For now the value is `ACADOS_INFTY=1e10` which might can be changed later after some more extensive testing.
- Support for that in QP solvers other than HPIPM is not included in this PR and tracked in https://github.com/acados/acados/issues/650

### DIFF
--- a/acados/ocp_nlp/ocp_nlp_constraints_bgh.c
+++ b/acados/ocp_nlp/ocp_nlp_constraints_bgh.c
@@ -1650,6 +1650,24 @@ void ocp_nlp_constraints_bgh_update_qp_vectors(void *config_, void *dims_, void 
     // fun[2*ni : 2*(ni+ns)] = - slack + slack_bounds
     blasfeo_daxpy(2*ns, -1.0, memory->ux, nu+nx, &model->d, 2*nb+2*ng+2*nh, &memory->fun, 2*nb+2*ng+2*nh);
 
+    // Set dmask for QP: 0 means unconstrained.
+    for (int i = 0; i < nb+ng+nh; i++)
+    {
+        if (BLASFEO_DVECEL(&model->d, i) == -ACADOS_INFTY)
+        {
+            // printf("found upper infinity bound\n");
+            BLASFEO_DVECEL(memory->dmask, i) = 0;
+        }
+    }
+    for (int i = nb+ng+nh; i < 2*nb+ng+nh; i++)
+    {
+        if (BLASFEO_DVECEL(&model->d, i) == ACADOS_INFTY)
+        {
+            // printf("found upper infinity bound\n");
+            BLASFEO_DVECEL(memory->dmask, i) = 0;
+        }
+    }
+
     return;
 }
 
--- a/acados/ocp_nlp/ocp_nlp_constraints_bgh.c
+++ b/acados/ocp_nlp/ocp_nlp_constraints_bgh.c
@@ -1612,6 +1612,9 @@ void ocp_nlp_constraints_bgh_compute_fun(void *config_, void *dims_, void *model
     // fun[2*ni : 2*(ni+ns)] = - slack + slack_bounds
     blasfeo_daxpy(2*ns, -1.0, ux, nu+nx, &model->d, 2*nb+2*ng+2*nh, &memory->fun, 2*nb+2*ng+2*nh);
 
+    // fun = fun * mask
+    blasfeo_dvecmul(2*(nb+ng+nh), memory->dmask, 0, &memory->fun, 0, &memory->fun, 0);
+
     return;
 }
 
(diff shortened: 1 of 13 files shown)


acados/ocp_nlp/ocp_nlp_constraints_bgh.c
```diff
@@ -1644,6 +1650,24 @@ void ocp_nlp_constraints_bgh_update_qp_vectors(void *config_, void *dims_, void
     // fun[2*ni : 2*(ni+ns)] = - slack + slack_bounds
     blasfeo_daxpy(2*ns, -1.0, memory->ux, nu+nx, &model->d, 2*nb+2*ng+2*nh, &memory->fun, 2*nb+2*ng+2*nh);
 
+    // Set dmask for QP: 0 means unconstrained.
+    for (int i = 0; i < nb+ng+nh; i++)
+    {
+        if (BLASFEO_DVECEL(&model->d, i) == -ACADOS_INFTY)
```
sandmaennchen, line 1656: `<=`

acados/ocp_nlp/ocp_nlp_constraints_bgh.c
```diff
@@ -1644,6 +1650,24 @@ void ocp_nlp_constraints_bgh_update_qp_vectors(void *config_, void *dims_, void
...
+        {
+            // printf("found upper infinity bound\n");
+            BLASFEO_DVECEL(memory->dmask, i) = 0;
+        }
+    }
+    for (int i = nb+ng+nh; i < 2*nb+ng+nh; i++)
+    {
+        if (BLASFEO_DVECEL(&model->d, i) == ACADOS_INFTY)
```
sandmaennchen, line 1664: `>=`

sandmaennchen (approved):

acados/ocp_nlp/ocp_nlp_constraints_bgh.c
```diff
@@ -1644,6 +1653,27 @@ void ocp_nlp_constraints_bgh_update_qp_vectors(void *config_, void *dims_, void
     // fun[2*ni : 2*(ni+ns)] = - slack + slack_bounds
     blasfeo_daxpy(2*ns, -1.0, memory->ux, nu+nx, &model->d, 2*nb+2*ng+2*nh, &memory->fun, 2*nb+2*ng+2*nh);
 
+    // Set dmask for QP: 0 means unconstrained.
+    for (int i = 0; i < nb+ng+nh; i++)
+    {
+        if (BLASFEO_DVECEL(&model->d, i) <= -ACADOS_INFTY)
```
giaf, line 1659: I would go for exactly == , never know if a user wants to solve a badly scaled problem on purpose.
At least exactly == decreases likelihood of removing some desired constraint by chance.

giaf: Unless there are reasons why it should not hold exactly == with a constraint that was set to ACADOS_INFTY.

giaf:
`d_mask` has the same size as `d` (namely `2*nb+2*ng+2*ns`) in HPIPM, and also `model->d` has the same (linearized) size in constraints bgh and bgp.
So also the term `+2*ns` should be considered.

giaf:
HPIPM is implemented such that soft constraints slacks in the primal variables, Lagange multipliers and relative slacks are guaranteed to be exactly 0 for disregarded constraints, see e.g. https://github.com/giaf/hpipm/blob/master/ocp_qp/x_ocp_qp_ipm.c#L1821.
This means that at the NLP level you just need to mask out final components in the `fun` evaluation, as you are doing.

However, in the case of another QP solver, it has to be guaranteed in the QP solver interface that such quantities of disregarded constraints are and remain exactly at 0.
Otherwise, also input arguments to e.g. Hessian evaluation or fun evaluation or the like should be masked out properly.

FreyJo:
Thanks for checking @giaf !
For now, this is only tested for HPIPM as mentioned in the PR descriptions.
In case of other QP solvers, we surely need to do more work and testing, also with soft constrained problems.
I would try to keep track of the current state on this and the potential issues you mentioned in https://github.com/acados/acados/issues/650

Could you elaborate on what you mean with "relative slacks" here:

> HPIPM is implemented such that [...] relative slacks are guaranteed to be exactly 0 for disregarded constraints  [...] 

Regarding the comparison with `ACADOS_INFTY`, I discussed more with @sandmaennchen and we decided to stick with the `>=` comparison.
If a user wants to solve a badly scaled problem on purpose, they still need to be aware that `ACADOS_INFTY` exists and that it is set to `1e10`.
They might have a constraint function bounded by `1e10` which would be ignored, and one with `5e10` which is included in the acados NLP, which is kind of weird behavior.
I also believe that this is not really a breaking change for anyone, as I doubt there are successful acados applications around with constraint bounds `> 1e10`.

PR #1175, opened 2024-07-29
Use `d_mask` in DAQP interface to handle one-sided constraints

FreyJo:
See https://github.com/acados/acados/pull/1174 and https://github.com/acados/acados/issues/650


FreyJo:
Thanks for checking and clarifying @darnstrom !
Brief follow up: Is it safe to assume that the returned multipliers of constraints with `DAQP_INF` as bounds are zero?

darnstrom:
> Brief follow up: Is it safe to assume that the returned multipliers of constraints with `DAQP_INF` as bounds are zero?

Yes! It should be impossible for constraints with DAQP_INF to become active, which yields multipliers that are zero.

PR #1177, opened 2024-07-31
Simulink sparse parameter update and testing

FreyJo:
- Simulink: add possible output `parameter_traj`, for testing
- test setting and getting parameters in Simulink in `simulink_param_test`
- add function `add_sparse_param_port_simulink(simulink_opts, idx_p, port_name, stage_idx_0, stage_idx_e)` in Matlab, which allows to specify information for an input port of the simulink block corresponding to an acados OCP solver. Inputs are
    - `idx_p` is a 0-based vector of parameter indices to be updated by the port.
    - `port_name` is used to identify the port and print information.
    - `stage_idx_0` is the first stage for which the parameters should be updated by the port (in `[0, N]`).
    - `stage_idx_e` is the last stage for which the parameters should be updated by the port (in `[stage_idx_0, N]`).
- The resulting port gets as inputs:
   - first one:  value which indicates if update should be performed (0 means no update).
   - afterwards: new numerical values of parameters to be updated at stages specified by `stage_idx_0`, `stage_idx_e`.
   - size of input port: 1 + (`stage_idx_e` - `stage_idx_0` + 1) * length(`idx_p`)

- test sparse parameter update ports in Simulink


examples/acados_matlab_octave/test/simulink_sparse_param_test.m
```diff
@@ -0,0 +1,124 @@
...
+simulink_opts.inputs.u_init = 0;
+simulink_opts.inputs.pi_init = 0;
+simulink_opts.inputs.ignore_inits = 0;
+simulink_opts.outputs.pi_all = 0;
+simulink_opts.outputs.sqp_iter = 0;
+
+% parameter ports
+simulink_opts.inputs.parameter_traj = 0;
+simulink_opts.outputs.parameter_traj = 1;
+simulink_opts = add_sparse_param_port_simulink(simulink_opts, 0:7, 'first_8', 0, N);
```
sandmaennchen, line 63: maybe add a comment that this adds multiple sparse parameter ports?

Issue #1125, opened 2024-05-31
Error compiling Acados with GCC 14.1.1: `implicit declaration of function ‘posix_memalign’`


hshose:
Compiling acados with up-to-date `gcc 14.1.1 20240522`, I get the following error:

```
acados/external/blasfeo/auxiliary/blasfeo_stdlib.c: In function ‘blasfeo_malloc_align’:
acados/external/blasfeo/auxiliary/blasfeo_stdlib.c:72:19: error: implicit declaration of function ‘posix_memalign’ [-Wimplicit-function-declaration]
   72 |         int err = posix_memalign( ptr, CACHE_LINE_SIZE, size );
      |                   ^~~~~~~~~~~~~~
make[2]: *** [external/blasfeo/CMakeFiles/blasfeo.dir/build.make:93: external/blasfeo/CMakeFiles/blasfeo.dir/auxiliary/blasfeo_stdlib.c.o] Error 1
make[2]: *** Waiting for unfinished jobs....
```

Downgrading to `gcc 13.2.1-6 20240417` resolves the issue for now.

hshose:
Proposed fix in [blasfeo PR](https://github.com/giaf/blasfeo/pull/175)

tmmsartor:
Related [commend](https://github.com/giaf/blasfeo/pull/175#issuecomment-2184079588)
Changing compilation standard for GNU/Clang case from `-std=c99` to `-std=gnu99` can also fix the issue.
If `acados` is not using gnu specific C non-standard extensions using `-std=gnu99` should not results in any differences, but that must be checked.
Apparently one difference is that some warnings are considered errors, namely `incompatible-pointer-types` and `int-conversion` but this can be overridden with `-Wno-incompatible-pointer-types` and `-Wno-int-conversion`.

NB: to see the different C-macro defined in the two cases one can run something like:
```bash
touch empty.c
gcc -std=gnu99 -dM -E empty.c > gnu99.c
gcc -std=c99 -dM -E empty.c > c99.c
vimdiff gnu99.c c99.c
```

tmmsartor:
Another way to fix this is to set to pass that flag from `acados` i.e. in the way of #1194 .

Also worth to mention that when trying to compile blasfeo without adding additional C_FLAGS like `c99` (which also diasable  the posix_memalign  definition) unfolds to ramification which ultimately make same test to fails, even if it does not seem a catastrophic failure (`Exception: acados solver took 23 iterations, expected range(17, 23).`)

[...]

PR #1194, opened 2024-08-13
1125: explicitly enable POSIX symbols used by blasfeo

tmmsartor:
See #1125

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -128,6 +128,7 @@ endif()
 
 if(CMAKE_CXX_COMPILER_ID MATCHES "GNU" OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
     set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=c99 -Wall -fdiagnostics-show-option")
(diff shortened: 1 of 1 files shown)


FreyJo:
Would it be possible to make a similar change directly in the BLASFEO CMake files?
It seems this flag is only required in BLASFEO. After reading through your comments on https://github.com/acados/acados/issues/1195, it seems that implementing the fix within BLASFEO would be more beneficial, as it would also apply when using a BLASFEO installation that wasn't compiled within acados. This approach would therefore be preferable.

tmmsartor:
The point is that BLASFEO is compiling and passing its test successfully without that flag, because the problem arise only when adding the `-std=c99` which among other things disable the definition of 'posix_memalign` (a POSIX specific method) I think that adding the macro `_POSIX_C_SOURCE` in BLASFEO to counter the effect of `-std=c99` when  `-std=c99` is not used in BLASFEO (at least not as default) should be further motivated.

Anyway there are still some related questions that I will like/try to answer:
* Would be actually good for BLASFEO to use `-std=c99` by default also when compiling with gnu compilers (Clang, gcc)?
* Why some `acados` tests are failing if BLASFEO is not compiled with `-std=c99`?
I will reference any findings in this PR.

FreyJo:
Closing this, as https://github.com/acados/acados/pull/1342 was merged.

PR #1217, opened 2024-08-22
Correction: Torsion calculation in quadrotor_nav example

JohnTGZ:
Torsion should be a product of the normal and the derivative of the binormal with respect to arc-length.
Reference: [Minimum-time trajectory generation for quadrotors in constrained environments](https://arxiv.org/abs/1706.06478)


AkashJohnSubash (approved):
Good catch @JohnTGZ, this was an experimental definition I forgot to revert from.

FreyJo (approved):

PR #1221, opened 2024-08-27
Update problem formulation

sandmaennchen:
This PR updates the problem formulation which is now valid for both the MATLAB and python interfaces :partying_face: 
* Improved some docstrings along the way.
* Added note on the two Matlab interfaces and examples in the docs.

The section on solver options has been updated, but there is many more options which are not part of this document yet.


docs/problem_formulation/problem_formulation_ocp_mex.tex
```diff
@@ -487,123 +579,86 @@ \subsection{Terminal Constraints}\label{sec:constraints:terminal}
...
-Tables \ref{tab:solver_options}, \ref{tab:solver_options:qp_solver_option} and \ref{tab:solver_options:regularize_method} show (almost) all available options.
-These options are set in \matlab{} via \code{ocp\_opts.set(<stringid>, <value>)}.
-% Some options are not (yet) implemented in the \code{.set(*)}-interface and can be accessed via a struct by using \code{ocp\_opts.opts\_struct.(<stringid>) = <value>}.
-Furthermore, the struct \code{ocp\_opts.opts\_struct} and \href{https://github.com/acados/acados/blob/master/interfaces/acados_matlab_octave/acados_ocp_opts.m}{\code{acados\_ocp\_opts.m}} can be used as a reference for what other fields are available.
-\\
-Note that some options of the solver can be modified after creation using the routine: \code{set(<stringid>, <value>)}.
+
+
+Tables \ref{tab:solver_options}, \ref{tab:solver_options:qp_solver_option} and \ref{tab:solver_options:regularize_method} show some of the available options.
+A full list of the available options is given on the \href{https://docs.acados.org/python_interface/index.html#acados_template.acados_ocp_options.AcadosOcpOptions}{docs page generated from the docstrings in~\python}.
```
FreyJo, line 616: Maybe even highlight, that we recommend to look up the options there instead of in the PDF.
Mention MATLAB specific options separately: I think it is just `compile_interface` right now.
Mention options class `AcadosOcpOptions`

docs/problem_formulation/problem_formulation_ocp_mex.tex
```diff
@@ -631,47 +686,54 @@ \section{Solver \& Options}\label{sec:solver}
...
+        % \multirow{2}{2cm}{\centering OOQP} & partial & \code{PARTIAL\_CONDENSING\_ooqp} \\
          % & full & \code{full\_condensing\_ooqp} \\[1ex]
-        OSQP & partial & \code{partial\_condensing\_osqp} \\
+        OSQP & partial & \code{PARTIAL\_CONDENSING\_OSQP} & \cite{Stellato2020} \\
         % QORE & full & \code{full\_condensing\_qore} \\
-        qpDUNES & partial & \code{partial\_condensing\_qpdunes} \\
-        qpOASES & full & \code{full\_condensing\_qpoases} \\
-        DAQP & full & \code{full\_condensing\_daqp} \\
+        qpDUNES & partial & \code{PARTIAL\_CONDENSING\_QPDUNES} & \cite{Frasch2015}\\
+        qpOASES & full & \code{full\_condensing\_QPOASES} & \cite{Ferreau2014}\\
```
FreyJo, line 706: Capitalization is off, but we could remove the Table and just link to the property:
https://docs.acados.org/python_interface/index.html#acados_template.acados_ocp_options.AcadosOcpOptions.qp_solver

docs/problem_formulation/problem_formulation_ocp_mex.tex
```diff
@@ -156,13 +168,12 @@ \section{Problem Formulation}\label{sec:problem}
 
 %
 \paragraph{Document Purpose.}
-This document describes the \matlab{} interface of \acados.
-Here, the focus is to give a mathematical overview of the problem formulation and possible options to model it within \acados.
-The problem formulation and the possibilities of \acados{} are similar in the \python{} interface, however, some of the string identifiers are different.
-The documentation is not exhaustive and does not contain a full description for the \matlab{} interface.
+This document describes the class of optimal control problems that can be formulated and solved in \acados{} via the class \texttt{AcadosOcp}, available in the \matlab{} and \python{} interfaces.
+%Here, the focus is to give a mathematical overview of the problem formulation and possible options to model it within \acados.
+This documentation is (most likely) not exhaustive and does not contain a full description of all options provided by the \python{} and \matlab{} interfaces.
```
FreyJo, line 173: Maybe add one sentence on `AcadosMultiphaseOcp`?

docs/problem_formulation/problem_formulation_ocp_mex.tex
```diff
@@ -156,13 +168,12 @@ \section{Problem Formulation}\label{sec:problem}
...
 
-You can find examples in the directory \code{<acados>/examples/acados\_matlab\_octave}.
-The source code of the \acados{} \matlab{} interface is found in: \code{<acados>/interfaces/acados\_matlab\_octave} and should serve as a more extensive, complete and up-to-date documentation about the possibilities.
+%You can find examples in the directory \code{<acados>/examples/acados\_matlab\_octave}.
+%The source code of the \acados{} \matlab{} interface is found in: \code{<acados>/interfaces/acados\_matlab\_octave} and should serve as a more extensive, complete and up-to-date documentation about the possibilities.
```
FreyJo, line 176: The following paragraph "Abbreviations" seems to be not properly used anymore.
This paragraph is mainly to explain `Z*` penalties and `J*` matrices.

The `Z*` slack penalties are after the overhaul given as vectors, before diagonal matrices.
There are two things that should be clarified in this PDF: the restriction (not any matrix) and the way they are to be set.

I suggest adding a remark for `J*`/ `Js*` matrices: explaining restrictions on what they can be and how they relate to `idx*`. Then link to this remark in section 1 and the constraint section.

I would suggest to handle the `Z*` similarly.
Does that make sense?

FreyJo (approved):
I pushed minor modifications and added some comments, which I hope you can address before merging :)

docs/problem_formulation/problem_formulation_ocp_mex.tex
sandmaennchen: done

docs/problem_formulation/problem_formulation_ocp_mex.tex
sandmaennchen: capitalization is fixed! I would keep the table to also give the references for the QP solvers

docs/problem_formulation/problem_formulation_ocp_mex.tex
sandmaennchen: I shortened this section a bit and provide an example. Maybe check this one again :see_no_evil:

FreyJo:
Found one last typo. Please merge if you like :)

PR #1234, opened 2024-09-02
Add `p_global` to problem formulation in Python

FreyJo:
This allows users to define a CasADi variable `AcadosModel.p_global` containing global parameters, i.e. parameters that can not change over the horizon.
The code generation of external functions has been refactored using the new `GenerateContext` class.
This class extracts all expressions that only depend on `p_global` from the CasADi graph of the output expressions and replaces them with "double memory pools".
This allows for an efficient precomputation of potentially expensive terms which only depend on these parameters, e.g. spline coefficients, when p_global are underlying data points.
However, since the double pools are part of the compilation unit of the external functions, this implementation is inherently limited to global parameters.
Implementing different parameters over the horizon for which expressions are precomputed can be done by using an `AcadosMultiphaseOcp` formulation and defining `p_global` as the concatenation of the corresponding parameters of the phases.

Only supported for OCP solvers.
Updating these parameters can be done using `AcadosOcpSolver.set_p_global(values)`
NOTE: this is only supported with CasADi beta release https://github.com/casadi/casadi/releases/tag/nightly-se


examples/acados_python/p_global_example/example_p_global.py
```diff
@@ -0,0 +1,299 @@
...
+    Fmax = 80
+    ocp.constraints.lbu = np.array([-Fmax])
+    ocp.constraints.ubu = np.array([+Fmax])
+    ocp.constraints.idxbu = np.array([0])
+
+    ocp.constraints.x0 = np.array([0.0, np.pi, 0.0, 0.0])
+
+    ocp.parameter_values = np.array([9.81])
+
+    if not use_p_global:
```
sandmaennchen, line 170: maybe move up to where `p_global` is set initially

examples/acados_python/p_global_example/utils.py
```diff
@@ -0,0 +1,98 @@
...
+        X_true: arrray with shape (N_sim, nx)
+        X_est: arrray with shape (N_sim-N_mhe, nx)
+        Y_measured: array with shape (N_sim, ny)
+        latexify: latex style plots
+    """
+
+    if latexify:
+        latexify_plot()
+
+    WITH_ESTIMATION = X_est is not None and Y_measured is not None
```
sandmaennchen, line 53: remove estimation part?

interfaces/acados_template/acados_template/acados_dims.py
```diff
@@ -364,6 +365,11 @@ def ng_e(self):
         Type: int; default: 0"""
         return self.__ng_e
 
+    @property
+    def np_global(self):
+        """length of p_global; default: 0"""
```
sandmaennchen, line 370: length sounds weird here, simply _number of global parameters_ as above?

interfaces/acados_template/acados_template/acados_model.py
```diff
@@ -71,6 +71,14 @@ def __init__(self):
         The time dependency can be used within cost formulations and is relevant when cost integration is used.
         Start times of shooting intervals can be added using parameters.
         """
+        self.p_global = None
+        """
+        CasADi variable containing parameters that change rarely;
```
sandmaennchen, line 76: maybe first mention that these parameters are global? it might also be computationally beneficial to use them if they change often as they don't have to be set for each node individually

sandmaennchen: and aren't the computations performed in `set_p_global`? instead of custom update?

interfaces/acados_template/acados_template/casadi_function_generation.py
```diff
@@ -571,6 +560,7 @@ def generate_c_code_conl_cost(model: AcadosModel, stage_type: str, opts):
     fun_name_cost_fun_jac_hess = model.name + suffix_name_fun_jac_hess
 
     # set up functions to be exported
+    # TODO: fix these for p_global
```
sandmaennchen, line 563: what is missing here? maybe add to PR description?

sandmaennchen (approved):

examples/acados_python/p_global_example/example_p_global.py
FreyJo: `p_global_values` are not available at this point. This split is done in the example to allow for testing an equivalent formulation without `p_global` and corresponding version with multi-phase.

interfaces/acados_template/acados_template/acados_model.py
FreyJo: Oh yes, sorry, this was actually already in a commit that I forgot to push. Thanks!

interfaces/acados_template/acados_template/casadi_function_generation.py
FreyJo: I believe that this commit https://github.com/acados/acados/pull/1234/commits/3e2be45062243d44c6baeef8c82957591fc7adec should fix the code generation also if the outer loss depends on `p_global`.
Before that commit I would expect that CasADi complains about free variables.

PR #1235, opened 2024-09-02
Fix compilation with `EXT_DEP=OFF` and test it on Github actions

FreyJo:
- update BLASFEO and HPIPM
- create C example that does not use the BLASFEO print routines
- create a Github actions workflow that tests compilation with `EXT_DEP=OFF`.
- addresses the issue reported in https://discourse.acados.org/t/compiling-error-for-microautobox-iii/1757


giaf:
Both BLASFEO and HPIPM now have an `EXT_DEP` compilation option, that can be used to remove all dependencies on `malloc`, `free` and `printf`. There remain dependencies on `sqrt` and compiler-added routines like e.g. to check stack size.

In BLASFEO, removing the possibility to perform dynamic memory allocation comes with some consequences for some routines and matrix size combinations: possibly the routine would exit for too large matrix sizes.
However, for the panel-major matrix format that we generally use, drawbacks are rather limited.
As far as I can tell, the only one with consequences in the current acados implementation is the use of `dgemm_tn` in here https://github.com/acados/acados/blob/master/acados/sim/sim_irk_integrator.c#L1370 and other occurrences in the same file.
For large matrices, it could call the slow reference routines or exit, depending on the exact compilation flags.
I would suggest to avoid the use of `dgemm_tn` whenever possible (also because it has slower performance in any case compared to the other `dgemm` variants), and in this specific case it can be easily achieved by saving the transposed of `tmp_nux_ny` (that is computed anyway) in a new matrix other than `J_y_tilde` like e.g. `tmp_ny_nux`, that is passed to `dgemm_nn` instead of `dgemm_tn`.
This would both improve performance in general and avoid the possible issues described above.

sandmaennchen (approved):

PR #1241, opened 2024-09-05
MATLAB: add p_global and GenerateContext

sandmaennchen:
This PR
* introduces `p_global` to `AcadosModel`
* introduces `GenerateContext` for code generation.

Compare also https://github.com/acados/acados/pull/1234.

And a minor cleanup of code generation in python.


interfaces/acados_template/acados_template/c_templates_tera/matlab_templates/acados_mex_set_p_global.in.c
```diff
@@ -0,0 +1,120 @@
...
+    {% elif problem_class == 'MOCP' and phases_dims[0].np_global > 0 %}
+        int np_global = {{ phases_dims[0].np_global }};
+    {%- endif %}
+        if (matlab_size != np_global)
+        {
+            sprintf(buffer, "acados_mex_set_p_global: np_global = %d should match data_len = %d. Exiting.\n", np_global, matlab_size);
+            mexErrMsgTxt(buffer);
+        }
+
+    {% set n_pools = casadi_pool_names | length %}
```
FreyJo, line 98: I am quite confused about this.
It should not be needed and we should just call the function:
`int {{ name }}_acados_set_p_global({{ name }}_solver_capsule* capsule, double* data, int data_len)`
provided in `acados_solver.in.c` and with the same signature in the multi-phase template.
Or am I missing something?

Finally, we could completely avoid the file, if we implement it within `ocp_set`.
Which I think it is a good idea to try to not add too many templated MEX files.

interfaces/acados_template/acados_template/c_templates_tera/matlab_templates/acados_mex_set_p_global.in.c
```diff
@@ -0,0 +1,120 @@
...
+        fun->casadi_fun((const double **) fun->args, fun->res, fun->iw, fun->w, NULL);
+
+    {%- else %}
+        sprintf(buffer, "acados_mex_set_p_global: p_global is not defined, acados_mex_set_p_global does nothing.\n");
+        mexWarnMsgTxt(buffer);
+    {%- endif %}
+    }
+    else
+    {
+        sprintf(buffer, "acados_mex_set_p_global: wrong nrhs: %d\n", nrhs);
```
FreyJo, line 114: This is caught only called in MATLAB interface and not by the user, so we don't need to check too much.

interfaces/acados_template/acados_template/c_templates_tera/matlab_templates/mex_solver.in.m
```diff
@@ -115,6 +115,23 @@ function set_params_sparse(varargin)
...
+
+        function set_p_global(varargin)
+            % usage:
+            % ocp.set_p_global(val)
+            % Sets p_global to val and precomputes all parts of the CasADi graphs of all other functions that only depend on p_global.
+            obj = varargin{1};
+            val = varargin{2};
+
+            if nargin == 2
+                acados_mex_set_p_global_{{ name }}(obj.C_ocp, val);
```
FreyJo, line 127: Maybe better to move these checks up to `AcadosOcpSolver`. Should make it easier to debug for the user.
At some point (before the overhaul), I thought we just use the templated solver class. But by now I think we want to keep `AcadosOcpSolver`

docs/problem_formulation/problem_formulation_ocp_mex.tex
```diff
@@ -117,36 +118,36 @@ \section{Problem Formulation}\label{sec:problem}
     &&&\mathComment{Initial values, see section \ref{sec:constraints:initial}}\nonumber\\
     &\,\,\,\quad \text{s.t.}    &&\underline{x}_0 \leq J\ind{bx}\initial \, x(0) \leq \overline{x}_0 , && \label{eq:constraints:initial}\\[1ex]
     &&&\mathComment{Nonlinear constraints on the initial shooting node} \nonumber\\
-    &&&\underline{h}\initial \leq h\initial(x(0), u(0), z(0), p) + J_{\textrm{sh}}\initial \, s\lowerh\initial, &&\label{eq:constraints:path:lower_bounds:start}\\
-    &&& h\initial(x(0), u(0),  z(0), p) - J\ind{sh}\initial\, s\ind{u,h}\initial \leq \overline{h}\initial, &&\label{eq:constraints:path:upper_bounds:start}\\
+    &&&\underline{h}\initial \leq h\initial(x(0), u(0), z(0), p(0), p\glob) + J_{\textrm{sh}}\initial \, s\lowerh\initial, &&\label{eq:constraints:path:lower_bounds:start}\\
```
FreyJo, line 121: I think the PDF is not added (properly).
But changes here look fine :+1:

interfaces/acados_matlab_octave/AcadosMultiphaseOcp.m
```diff
@@ -106,6 +108,9 @@
             acados_folder = getenv('ACADOS_INSTALL_DIR');
             obj.acados_include_path = [acados_folder, '/include'];
             obj.acados_lib_path = [acados_folder, '/lib'];
+
+            obj.casadi_pool_names = [];
```
FreyJo, line 112: I think we dont have this property in the corresponding python class. Is it needed here? why?

FreyJo: Oh I see, we do set it, but we just didn't add this property and just write it to the class anyway.
Maybe we should make it a private property in python then.

FreyJo (commented):
I left some comments, but did not check the example carefully.
Overall looks very good!

interfaces/acados_template/acados_template/acados_ocp.py
```diff
@@ -1013,17 +1013,17 @@ def generate_external_functions(self, context: Optional[GenerateContext] = None)
         check_casadi_version()
         if self.model.dyn_ext_fun_type == 'casadi':
             if self.solver_options.integrator_type == 'ERK':
-                generate_c_code_explicit_ode(context, model)
+                generate_c_code_explicit_ode(context, model, model_dir)
```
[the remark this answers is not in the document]
FreyJo: Same for the Matlab functions.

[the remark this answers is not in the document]
sandmaennchen: but this way `model_dir` is defined only once and not five times.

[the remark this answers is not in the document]
sandmaennchen: one could move it to context though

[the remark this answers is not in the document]
FreyJo: I think we cant, because for multiphase the models vary but there is only one context.
I would prefer setting it up in place still but not too strongly.

FreyJo (approved):
:racehorse:

Issue #1207, opened 2024-08-19
Use CasADi common subexpression elimination


FreyJo:
Added in [CasADi 3.6.6](https://github.com/casadi/casadi/releases/tag/3.6.6).
Should be used whenever generating CasADi functions .

PR #1246, opened 2024-09-11
Add common subexpression elimination to CasADi function creation

sandmaennchen:
If the CasADi version supports common subexpression elimination, the corresponding option is set when creating CasADi functions.

Closes https://github.com/acados/acados/issues/1207


FreyJo (approved):

FreyJo:
Would be nice to make a quick commit comparison experiment to evaluate this.

PR #1247, opened 2024-09-11
MATLAB: Simple DAE example uses new interface

sandmaennchen:
This PR transcribes the simple DAE example to the new interface. The old example is moved to `legacy_interface` for testing purposes.

In the transcribed example a Gauss-Newton Hessian is used in order to also be able to run the NLS variant.


FreyJo (approved):
Looks good!

PR #1252, opened 2024-09-13
Update matlab install script to install CasADi v3.6.6

sandmaennchen:


interfaces/acados_matlab_octave/check_acados_requirements.m
```diff
@@ -36,7 +36,9 @@ function check_acados_requirements(varargin)
             error('Please set up CasADi yourself and try again.');
         end
         % download CasADi
-        CasADi_version = '3.5.5'; % TODO: update automatic download to later CasADi version.
+        CasADi_version = '3.6.6'; % TODO: update automatic download to later CasADi version.
```
FreyJo, line 39: remove TODO?

interfaces/acados_matlab_octave/check_acados_requirements.m
```diff
@@ -36,7 +36,9 @@ function check_acados_requirements(varargin)
...
+        CasADi_version_numeric = str2num(replace(CasADi_version, '.', ''));
```
FreyJo, line 40: This is not really save, meaning that if double digit minor versions or patches are present this could be a nonunique mapping. But good enough for this, I don't have something to propose.

However, this comparison will not work if we want to link to a pre-release of casadi without a real version number.
Maybe add a boolean `later_than_36` which we then set manually?

interfaces/acados_matlab_octave/check_acados_requirements.m
```diff
@@ -51,27 +53,46 @@ function check_acados_requirements(varargin)
...
 
         if ismac
-            if ~verLessThan('matlab', '8.5')
-                filename = strcat('casadi-osx-matlabR2015a-v', CasADi_version, '.tar.gz');
-            elseif ~verLessThan('matlab', '8.4')
-                filename = strcat('casadi-osx-matlabR2014b-v', CasADi_version, '.tar.gz');
-            elseif ~verLessThan('matlab', '8.3')
-                filename = strcat('casadi-osx-matlabR2014a-v', CasADi_version, '.tar.gz');
+        CasADi_version = '3.5.5'; % TODO: update automatic download to later CasADi version.
+            if CasADi_version_numeric < 366
```
FreyJo, line 57: Seems like this change was with 3.6.0 in CasADi, so we should use 360 and something like the flag I described above?
Or we just remove the CasADi < 3.6 case altogether here. It will anyway not be an automatic download if you change the version in the script manually.

interfaces/acados_matlab_octave/check_acados_requirements.m
```diff
@@ -51,27 +53,46 @@ function check_acados_requirements(varargin)
         end
 
         if ismac
-            if ~verLessThan('matlab', '8.5')
-                filename = strcat('casadi-osx-matlabR2015a-v', CasADi_version, '.tar.gz');
-            elseif ~verLessThan('matlab', '8.4')
-                filename = strcat('casadi-osx-matlabR2014b-v', CasADi_version, '.tar.gz');
-            elseif ~verLessThan('matlab', '8.3')
-                filename = strcat('casadi-osx-matlabR2014a-v', CasADi_version, '.tar.gz');
+        CasADi_version = '3.5.5'; % TODO: update automatic download to later CasADi version.
```
FreyJo, line 56: It seems to be not implemented for Mac yet?
Here `CasADi_version` and `CasADi_version_numeric` are not consistent.

interfaces/acados_matlab_octave/check_acados_requirements.m
```diff
@@ -51,27 +53,46 @@ function check_acados_requirements(varargin)
...
+                elseif ~verLessThan('matlab', '8.4')
+                    filename = strcat('casadi-windows-matlabR2014b-v', CasADi_version,'.zip');
+                elseif ~verLessThan('matlab', '8.3')
+                    filename = strcat('casadi-windows-matlabR2014a-v', CasADi_version,'.zip');
+                end
+            else
+                if ~verLessThan('matlab', '9.5')
+                    filename = strcat('casadi-', CasADi_version, '-windows64-matlab2018b.zip');
+                end
             end
```
FreyJo, line 96: Maybe before the `try` download thing, check if `filename` is specified. And if not raise that we cannot find compatible CasADi version for the system?

FreyJo (commented):
I left some comments.
I think I would be in favor of removing the  CasADi < 3.6 case.

FreyJo (approved):
Looks good now :)

PR #1257, opened 2024-09-24
Code generation: more checks whether exact Hessian contributions are needed

sandmaennchen:
With this PR the CasADi functions associated with exact Hessian contributions from NLS cost and nonlinear constraints are only generated if an exact Hessian should be used.


interfaces/acados_template/acados_template/c_templates_tera/multi_CMakeLists.in.txt
```diff
@@ -126,20 +126,20 @@ set(MODEL_SRC
     {{ model[jj].name }}_model/{{ model[jj].name }}_expl_ode_fun.c
     {{ model[jj].name }}_model/{{ model[jj].name }}_expl_vde_forw.c
     {{ model[jj].name }}_model/{{ model[jj].name }}_expl_vde_adj.c
-	{%- if hessian_approx == "EXACT" %}
+	{%- if solver_options.hessian_approx == "EXACT" %}
```
FreyJo, line 129: There is some logic at the top of this file setting the option `hessian_approx`, which is an artifact from the single phase CMake file, which is also used when generating integrators.
This logic should be removed in this file and the multi Makefile.

FreyJo (commented):
The currently failing tests should pass after adapting the Simulink related Makefile.

FreyJo (approved):

FreyJo (approved):
This should be addressed in a follow-up: https://github.com/acados/acados/issues/1259

PR #1260, opened 2024-09-25
Remove `rti_phase` from DDP and SQP

### LINKED ISSUE #1255: Remove `rti_phase` from SQP and DDP
I think we should remove `rti_phase` from SQP and DDP. But in a separate PR.

_Originally posted by @FreyJo in https://github.com/acados/acados/pull/1254#discussion_r1768179065_

FreyJo:
closes #1255 

BREAKING: setting option `rti_phase` for NLP solvers other than `SQP_RTI` is not supported anymore.


sandmaennchen (approved):
:+1:

PR #1275, opened 2024-09-30
Simulink: add optional input port `p_global`, test blazing splines in Simulink

FreyJo:
- The port `p_global` expects 1+np_global numbers.
The first value indicates if the global parameters are to be updated, 0 means no update.

- Set `ext_fun_compile_flags` correctly in `make_sfun` such that blazing spline example compiles without warnings.

Added a corresponding test in the p_global Matlab example.


sandmaennchen (approved):
:+1:

PR #1278, opened 2024-10-02
Initialization of `p_global`

sandmaennchen:
Add field `p_global_values` to OCP object in MATLAB and python interface and initialize `p_global` with the provided values at solver creation.

### DIFF
--- a/interfaces/acados_template/acados_template/acados_multiphase_ocp.py
+++ b/interfaces/acados_template/acados_template/acados_multiphase_ocp.py
@@ -198,9 +198,9 @@ class AcadosMultiphaseOcp:
 
     @property
     def p_global_values(self):
-        """:math:`p` - list of initial values for p_global vector.
-        Type: `list` of `numpy.ndarray` of shape `(np_global_i, )`.
-        - can be updated stagewise."""
+        """:math:`p` - initial values for p_global vector.
+        Type: `numpy.ndarray` of shape `(np_global_i, )`.
+        """
         return self.__p_global_values
 
     @p_global_values.setter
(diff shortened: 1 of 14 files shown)


interfaces/acados_template/acados_template/acados_multiphase_ocp.py
```diff
@@ -194,6 +195,23 @@ def parameter_values(self, parameter_values):
...
+        - can be updated stagewise."""
+        return self.__p_global_values
+
+    @p_global_values.setter
+    def p_global_values(self, p_global_values):
+        if not isinstance(p_global_values, list):
+            raise Exception('p_global_values must be a list of numpy.ndarrays.')
+        elif len(p_global_values) != self.n_phases:
+            raise Exception('p_global_values must be a list of length n_phases.')
+        self.__p_global_values = p_global_values
```
FreyJo, line 212: `p_global` should be a single numpy array also for MOCP.
I remember that we opted for an implementation where `p_global` has to be the same vector for all stages (including MOCP formulation) and that if one indeed wants different parameters for the phases they all need to be in `p_global`.
Entries of `p_global` can be used across multiple phases.
This is actually already checked here:
https://github.com/acados/acados/blob/master/interfaces/acados_template/acados_template/acados_multiphase_ocp.py#L246-L252

interfaces/acados_template/acados_template/c_templates_tera/acados_multi_solver.in.c
```diff
@@ -58,6 +58,12 @@
 {%- endfor %}
 {%- set np_max = np_values | sort | last %}
 
+{%- set np_global_values = [] -%}
+{%- for jj in range(end=n_phases) %}
+    {%- set_global np_global_values = np_global_values | concat(with=(phases_dims[jj].np_global)) %}
+{%- endfor %}
+{%- set np_global_max = np_global_values | sort | last %}
```
FreyJo, line 65: this should be rremoved, just use `dims_0.np_global`

FreyJo (approved):

PR #1279, opened 2024-10-02
WIP: SQP with feasible QP

FreyJo:


david0oo:
Preliminary PR description:
This PR introduces a novel `nlp_solver_type` named `SQP_WITH_FEASIBLE_QP`. This means that in every iteration an l1-relaxed QP is solved that ensures that all iterations are well-defined. 

The following option was added:
- `globalization_funnel_use_merit_fun_only`: This turns off the funnel method and only checks for a penalty function. In general the funnel allows for more nonmonotonicity and converges often faster
- `initial_objective_multiplier`

Note that the novel method is still in a preliminary state and its development is still ongoing. The following problems can still happen:
- HPIPM fails: This in general happens if the problem is badly scaled. Therefore try to rescale your objective function. We are working on internal scaling. Set the qpsolver `mu0` to 1e4. Also Partial condensing might be a problem. Therefore, avoiding any condensing might be a possible 
- The solver converges to an infeasible stationary point. In this case, please change the initial `objective_multiplier` to something smaller than 1.

If you encounter problems, please report them in the acados forum.

This method is inspired by the following papers:
- Byrd, Nocedal, Curtis
- SQuID
- Gould, Robinson, Loh
- Inexact SQuID

PR #1289, opened 2024-10-08
Update qpOASES with cmake update

FreyJo:
Currently our CI is failing because the upload artifacts does not upload the link files somehow.
This would be one fix.

Related issue that caused the breaking change:
https://github.com/actions/upload-artifact/issues/590


sandmaennchen (approved):

FreyJo:
I think we really don't need the symlink and it might cause trouble in different places.
I will just merge this instead of hoping that Github actions does fix their behavior and not roll changes back and forth within a major version https://github.com/actions/upload-artifact/issues/631

PR #1292, opened 2024-10-10
Include integrator residual in termination criterion

sandmaennchen:
This PR adds the integrator residual into the termination criterion.
* add integrator residuals to `ocp_nlp_res`


david0oo:
whats the difference between res_integrator and res_eq?

sandmaennchen:
Hi David, this PR is very much work in progress and isn't working yet. The bug that we want to fix is reported here: https://discourse.acados.org/t/does-acados-support-state-times-input-dynamics/1793/11

The main problem is that for IRK integrators one needs to additionally check that the integrator residuals are satisfied at the current iterate. If the $z$ variables do not appear in the KKT conditions, but only in the integrator residuals, you might end up with $z$ values that do not satisfy your dynamics even though the solver is reporting status 0.

PR #1309, opened 2024-10-22
MATLAB: generic NLP example

josipkh:
This PR suggests a new example on solving a non-OCP NLP using the MATLAB interface.
The original code was taken from [this post](https://discourse.acados.org/t/solving-simple-nlp-problem-exploit-blasfeo-performance/271).


josipkh:
Thanks for the check :)
I saw the python example, but didn't look into the `test` folder for MATLAB...

I was mainly following the forum discussion so I only used N = 1, but I vaguely remember N = 0 failing in a different context.

FreyJo (approved):

PR #1313, opened 2024-10-24
Rework `p_global` to use `global_data` instead of memory pools

FreyJo:
Previously, as of #1241, `p_global` was only supported for `MX`, since the memory pool workflow is limited to this.
The `p_global_precompute_fun` now computes the same stuff but has one output vector called `global_data` (a concatenation of all the expressions previously computed by the function).
This `global_data` has been added to `ocp_nlp_in` and is added as a last input to all other external CasADi functions by the `GenerateContext`.

Detailed changes below:


sandmaennchen (approved):
looks great! :heart_eyes: amazing journey :rocket:

PR #1314, opened 2024-10-24
Github Actions: use Release in full_build, and Debug in ext_dep_off

FreyJo:
In this way the artifacts uploaded are compiled with Release which is nice when using them in other projects.


dirkpr:
Extremely useful PR, thanks! @FreyJo @sandmaennchen

sandmaennchen (approved):

PR #1316, opened 2024-10-25
Migrate sensitivities of solution and value function of parameters to `p_global`

FreyJo:
Solution sensitivities with respect to parameters and derivatives of the value function with respect to parameters are now computed with respect to `p_global` instead of `p`. This is anyway what was implemented conceptually. However, if different parameter values for stages were used, these sensitivities were not clearly defined.
The precomputations and the fact that the parameter vector is only defined once makes this implementation more efficient.

- added support for `p_global` in Cython

Breaking:
- concerns options `with_solution_sens_wrt_params`, `with_solution_sens_wrt_params`
- `eval_solution_sensitivity` with argument `with_respect_to = "params_global"`
- `eval_and_get_optimal_value_gradient` with argument `with_respect_to = "params_global"`
- rename option `with_respect_to='params_global'` to `'p_global'` with deprecation warning

### DIFF
--- a/acados/ocp_nlp/ocp_nlp_common.c
+++ b/acados/ocp_nlp/ocp_nlp_common.c
@@ -354,12 +354,29 @@ ocp_nlp_dims *ocp_nlp_dims_assign(void *config_, void *raw_memory)
 
 void ocp_nlp_dims_set_global(void *config_, void *dims_, const char *field, int value_field)
 {
-    // ocp_nlp_config *config = config_;
+    ocp_nlp_config *config = config_;
     ocp_nlp_dims *dims = dims_;
+    int N = dims->N;
 
     if (!strcmp(field, "np_global"))
     {
         dims->np_global = value_field;
+        // cost
+        for (int i = 0; i <= N; i++)
+        {
+            config->cost[i]->dims_set(config->cost[i], dims->cost[i], "np_global", &value_field);
+        }
+        // dynamics
+        for (int i = 0; i < N; i++)
+        {
+            config->dynamics[i]->dims_set(config->dynamics[i], dims->dynamics[i], "np_global", &value_field);
+        }
+        // TODO: implement for constraints
+        // // constraints
+        // for (int i = 0; i <= N; i++)
+        // {
+        //     config->constraints[i]->dims_set(config->constraints[i], dims->constraints[i], "np_global", &int_array[i]);
+        // }
     }
     else if (!strcmp(field, "n_global_data"))
     {
@@ -1720,6 +1737,8 @@ acados_size_t ocp_nlp_workspace_calculate_size(ocp_nlp_config *config, ocp_nlp_d
     ocp_nlp_constraints_config **constraints = config->constraints;
 
     int N = dims->N;
+    int np_global = dims->np_global;
+
     int *nx = dims->nx;
     // int *nu = dims->nu;
     int *ni = dims->ni;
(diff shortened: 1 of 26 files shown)


acados/ocp_nlp/ocp_nlp_common.c
```diff
@@ -3219,29 +3239,29 @@ void ocp_nlp_common_eval_lagr_grad_p(ocp_nlp_config *config, ocp_nlp_dims *dims,
     else if (!strcmp("params_global", field))
```
sandmaennchen, line 3239: rename to `p_global`?

acados/ocp_nlp/ocp_nlp_common.c
```diff
@@ -3209,7 +3229,7 @@ void ocp_nlp_common_eval_lagr_grad_p(ocp_nlp_config *config, ocp_nlp_dims *dims,
     int N = dims->N;
     // int *nu = dims->nu;
     // int *nx = dims->nx;
-    int *np = dims->np;
+    int np_global = dims->np_global;
 
     if (!strcmp("params_stage", field))
```
sandmaennchen, line 3234: rename to `p`?

interfaces/acados_template/acados_template/acados_ocp.py
```diff
@@ -845,6 +847,8 @@ def make_consistent(self, is_mocp_phase=False) -> None:
                     raise Exception(f'with_solution_sens_wrt_params is only implemented if constraints depend not on parameters. Got parameter dependency for {horizon_type} constraint.')
```
sandmaennchen, line 847: "don't depend on"

interfaces/acados_template/acados_template/acados_ocp_solver.py
```diff
@@ -510,7 +510,7 @@ def eval_and_get_optimal_value_gradient(self, with_respect_to: str = "initial_st
 
         Notes:
         - for field `initial_state`, the gradient is the Lagrange multiplier of the initial state constraint.
-        The gradient computation consist of adding the Lagrange multipliers correspondin to the upper and lower bound of the initial state.
+        The gradient computation consist of adding the Lagrange multipliers corresponding to the upper and lower bound of the initial state.
```
sandmaennchen, line 513: "consists"

interfaces/acados_template/acados_template/acados_ocp_solver_pyx.pyx
```diff
@@ -255,7 +255,7 @@ cdef class AcadosOcpSolverCython:
 
         Notes:
         - for field `initial_state`, the gradient is the Lagrange multiplier of the initial state constraint.
-        The gradient computation consist of adding the Lagrange multipliers correspondin to the upper and lower bound of the initial state.
+        The gradient computation consist of adding the Lagrange multipliers corresponding to the upper and lower bound of the initial state.
```
sandmaennchen, line 258: "consists"

sandmaennchen (approved):
nice! :heart_hands:

kisisjrlly:
Very impressive work. But I still have the question, why not implement sensitivity calculations for different stage parameter values p? Although the fact that the precomputation and parameter vectors are defined only once makes this implementation more efficient, this approach cannot satisfy all scenarios. For example, for some time-varying optimal control problems or scenarios where the parameter values of different stages are determined by other modules (such as the output of neural networks)

FreyJo:
Hi @kisisjrlly 
It is possible to use the stage wise parameters to select different parts of `p_global` and basically implement sensitivities wrt stage wise varying parameters. We do not have an example with this yet, but already thought about it and do expect such a formulation to be very efficient, such that no additional implementation in C is needed.
Probably, such a formulation will become a feature of [leap-c](https://github.com/leap-c/leap-c) soon.

PR #1320, opened 2024-10-28
Implement solver timeout functionality for SQP

sandmaennchen:
This PR implements an optional solver timeout within SQP.

New options
- `timeout_max_time`: set the maximum time before solver timeout. Default: 0 (= no timeout)
- `timeout_heuristic`: heuristic to use for computing `predicted_per_iter_time`, i.e. the computation time of the next SQP iteration; possible values are `'ZERO'` (default), `'LAST'`, `'MAX'`, `'AVERAGE'`

A solver timeout is triggered whenever `timeout_max_time <= time_tot + predicted_per_iter_time`.
The `predicted_per_iter_time` is estimated according to the used heuristic:
- `'ZERO'`: `predicted_per_iter_time` is always zero
-  `'LAST'`: time of last iteration is used as estimate for `predicted_per_iter_time`
- `'MAX_CALL'`: Use the maximum time per iteration for the current solver call as estimate.
- `'MAX_OVERALL'`: Use the maximum time per iteration over all solver calls as estimate.
-  `'AVERAGE'`: exponential moving average per iteration time is used as estimate for `predicted_per_iter_time` (weight is hardcoded to 0.5) 

Also:
- fix multiphase solver templates with DDP


FreyJo:
Great feature! :tada: 
I can produce the following plot in the Python example with these changes:
computation_time_LAST
computation_time_MAX
computation_time_NO_TIMEOUT
computation_time_ZERO

FreyJo (approved):
Great work!

FreyJo (approved):

mjt123:
This configuration works fine when I use the Python interface, but when I use the generated C++ code, there is no such setting. What should I do

PR #1327, opened 2024-10-28
Minor cleanup

josipkh:
This PR fixes a few small typos in the MATLAB examples and the website.

### DIFF
--- a/examples/acados_matlab_octave/test/test_all_examples.m
+++ b/examples/acados_matlab_octave/test/test_all_examples.m
@@ -35,7 +35,7 @@ clearvars; clc; close all;
 % list the examples you would like to test
 
 targets = {
-	'../control_rates/main.m';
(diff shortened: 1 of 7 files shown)


FreyJo (approved):
Thank you :)

examples/acados_matlab_octave/test/test_all_examples.m
```diff
@@ -35,6 +35,7 @@
 % list the examples you would like to test
 
 targets = {
+    '../control_rates/main.m';
```
FreyJo, line 38: It seems like this is failing.
But the fix might be contained in https://github.com/acados/acados/pull/1320/files

josipkh: Yes, I think I will wait until it's merged :)

PR #1330, opened 2024-10-29
Python: Add translate functionality from LS/NLS/CONL cost to external cost

sandmaennchen:
This PR add the function `translate_cost_to_external_cost` to the class `AcadosOcp`. It allows to reformulate costs of type `LINEAR_LS`, `NONLINEAR_LS` and `CONVEX_OVER_NONLINEAR` as costs of type `EXTERNAL`.


FreyJo (approved):
tip 🔝

PR #1331, opened 2024-10-29
Adjoint solution sensitivities with respect to global parameters

FreyJo:
- can be used from Python via `AcadosOcpSolver.eval_adjoint_solution_sensitivity()`
```
Evaluate the adjoint sensitivity of the solution with respect to the parameters.
    :param seed_x : np.ndarray or list of np.ndarrays - seed for the states at stage `stages`
    :param seed_u : np.ndarray or list of np.ndarrays - seed for the controls at stage `stages`
    :param stages_seed_x : int or list of int - stages corresponding to the seeds_x
    :param stages_seed_u : int or list of int - stages corresponding to the seeds_u
    :param with_respect_to : string in ["p_global"]
    :param sanity_checks : bool - whether to perform sanity checks, turn off for minimal overhead, default: True
 ```
- chain-mass example `solution_sensitivity_example.py`: fix wrt global parameters, extend to also evaluate an adjoint sensitivity and compare timings of forward vs adjoint solution sensitivities.
- Python: gather solution sensitivity specific tests in function `sanity_check_parametric_sensitivities()`
- add test / example: `examples/acados_python/pendulum_on_cart/solution_sensitivities/forw_vs_adj_param_sens.py`

### DIFF
--- a/examples/acados_python/pendulum_on_cart/solution_sensitivities/sensitivity_utils.py
+++ b/examples/acados_python/pendulum_on_cart/solution_sensitivities/sensitivity_utils.py
@@ -101,7 +101,7 @@ def export_parametric_ocp(
         # add nonlinear dependency in cost
         cost_scale_factor = ca.exp(cost_scale_param)
     else:
-        cost_scale_factor = ca.exp(ocp.model.p_global)
+        cost_scale_factor = 1.0
 
     # NOTE here we make the cost parametric
     ocp.model.cost_expr_ext_cost = cost_scale_factor * ocp.model.x.T @ Q_mat @ ocp.model.x + ocp.model.u.T @ R_mat @ ocp.model.u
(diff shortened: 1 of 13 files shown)

### CODE REVIEW

File: examples/acados_python/pendulum_on_cart/solution_sensitivities/sensitivity_utils.py [hunk 1]
```diff
@@ -83,25 +84,35 @@ def export_parametric_ocp(
...
 
+    if cost_scale_as_param:
+        # add parameter to model
+        cost_scale_param = ca.SX.sym('cost_scale_param')
+        ocp.model.p_global = ca.vertcat(ocp.model.p_global, cost_scale_param)
+        ocp.p_global_values = np.concatenate((ocp.p_global_values, np.ones((1,))))
+        # add nonlinear dependency in cost
+        cost_scale_factor = ca.exp(cost_scale_param)
+    else:
+        cost_scale_factor = ca.exp(ocp.model.p_global)
```
sandmaennchen, line 104: don't quite get what you want to do/test here? would expect the the cost_scale_factor to be constant in the else part?

sandmaennchen (approved):
love it! :heart_eyes:

examples/acados_python/pendulum_on_cart/solution_sensitivities/sensitivity_utils.py
FreyJo: I think it is mainly to test with more than 1 parameters. Additionally, it allows to make the cost (non)linear in the parameter which is good for testing.
But yes, in `else` it should be a fixed value. Just pushed a fix.

PR #1333, opened 2024-10-30
Consistent mask and multipliers

sandmaennchen:
This PR addresses the issue that masked constraints and multipliers might be inconsistent: If a constraint is masked, the corresponding multiplier should always be zero.

- The mask is updated whenever the constraint bounds are updated.
- lambdas are multiplied with the mask whenever they or the bounds are updated

`dmask` is now stored in `ocp_nlp_in` as the mask ist now required in `ocp_nlp_out_set` and `ocp_nlp_constraints_model_set`.

BREAKING CHANGES in C interface:
- `ocp_nlp_out_set` now requires `nlp_in` as additional argument
- `ocp_nlp_constraints_model_set` now requires `nlp_out` as additional argument
- `ocp_nlp_out` needs to be created before `ocp_nlp_in`


interfaces/acados_template/acados_template/c_templates_tera/matlab_templates/acados_mex_set.in.c
```diff
@@ -143,7 +143,47 @@ void mexFunction(int nlhs, mxArray *plhs[], int nrhs, const mxArray *prhs[])
...
+        }
+    }
+    else if (!strcmp(field, "constr_ubu"))
+    {
+        for (int ii=s0; ii<se; ii++)
+        {
+            acados_size = ocp_nlp_dims_get_from_attr(config, dims, out, ii, "ubu");
+            MEX_DIM_CHECK_VEC(fun_name, field, matlab_size, acados_size);
+            if (matlab_size != 0)
+                ocp_nlp_constraints_model_set(config, dims, in, out, ii, "ubu", value);
```
FreyJo, line 186: I don't think there should be lines added here.. Let's hope this fixes the failing test.

FreyJo (approved):
It does look very good now. I would still highly appreciate adding a test before merging :upside_down_face:

PR #1342, opened 2024-11-06
Update BLASFEO to remove POSIX alignment dependency

FreyJo:
Should fix https://github.com/acados/acados/issues/1125 see https://github.com/acados/acados/issues/1125#issuecomment-2458171848


giaf:
This https://github.com/giaf/blasfeo/commit/9c2a8345f9de37fbb4b3cd84810b0250810de622 should fix the failing test for reference.

PR #1343, opened 2024-11-06
Solution sensitivities refactoring

FreyJo:
refactor functionality in `ocp_nlp_params_jac_compute`.
New description of this function:
```
this function sets up: jac_lag_stat_p_global, jac_ineq_p_global, jac_dyn_p_global
    - jac_dyn_p_global is computed in dynamics module
    - jac_ineq_p_global is computed in constraints module (TODO!)
    - jac_lag_stat_p_global: first dynamics writes its contribution, then cost and constraints modules add their contribution.
```
Detailed:
- split `tmp_nvninx_np_global` into matrices `jac_lag_stat_p_global`, `jac_ineq_p_global`, `jac_dyn_p_global`
- fix shape of inequality part `jac_ineq_p_global`, should be 2*ni
- directly set up those matrices in the cost and dynamics modules instead of taking out columns in `ocp_nlp_common`
- avoid copying

This resulted in significant speed-up in the chain benchmark example:
Minimum time for `eval rhs` on my machine
Before the changes:  775 ms
After the changes: 527 ms


sandmaennchen (approved):
:+1:

Issue #1326, opened 2024-10-28
Fix residuals with `SQP_RTI`


sandmaennchen:

sandmaennchen:
Closed in https://github.com/acados/acados/pull/1346

PR #1346, opened 2024-11-08
Fix `get_residuals`

FreyJo:
- Python: implement convenience function `get_initial_residuals ` which are the residuals that always come for free in RTI, in the sense that they don't require additional external function evaluations
- C: call `eval_kkt_residual` in `eval_residuals`
- Python: when solver is RTI or `recompute` is True, `get_residuals` computes the correct residuals when updating parameters or other problem data.
- implement `eval_kkt_residual` for all NLP solvers.
- test correct behavior

- Simulink output port `KKT_residuals` now also computes the residuals correctly in case of RTI. This can significantly increase the turnaround time of the Simulink block when the output port is activated!


FreyJo:
Closes https://github.com/acados/acados/issues/1326

sandmaennchen (approved):

PR #1347, opened 2024-11-08
MOCP: fixes and test with parameters in MATLAB on CI

FreyJo:
- MATLAB: fix zero default for `parameter_values` of MOCP
- fix `cost_0` in `multi_CMakeLists.in.txt `
- improve dimension check when setting parameters
- test MOCP with parameters in MATLAB


sandmaennchen (approved):
great! :1st_place_medal:

PR #1352, opened 2024-11-13
Fix value function sensitivities, add example varying parameters in certain direction on linear OCP

FreyJo:
- fixed external function generation: output should be column vector
- external functions: use `exit` instead of assert to also check when compiling with Release
- `ocp_nlp_common_eval_lagr_grad_p`: pass correct option pointers to submodules
- fix adj_p in constraint module with nh==0
- add new example and improved plot


sandmaennchen (approved):
:+1:

PR #1354, opened 2024-11-13
Fix convex-over-nonlinear function generation, add convex OCP example

FreyJo:
witnessed quadratic convergence


sandmaennchen (approved):
nice!

PR #1357, opened 2024-11-15
Add option `cost_scaling` in template based interfaces

FreyJo:
This allows to have `AcadosOcp.shooting_nodes` be "correct" when incorporating transition stages with costs.
Simplified transition examples accordingly.
```
cost_scaling: 
Vector with cost scaling factors of length N_horizon + 1.
If None set automatically to [`time_steps`, 1.0].
Default: None
```
Improved docstrings for `time_steps` and `shooting_nodes`


sandmaennchen (approved):
:hearts:

PR #1383, opened 2024-11-29
Python/Matlab: More efficient code generation for MOCP models

sandmaennchen:
This PR ignores non-required stages in code generation for multi-phase OCPs. This way the code generation overhead is reduced. If global parameters are used, this might also avoid the computation of parametric expressions which are not needed.

### DIFF
--- a/interfaces/acados_matlab_octave/AcadosOcp.m
+++ b/interfaces/acados_matlab_octave/AcadosOcp.m
@@ -855,14 +855,14 @@ classdef AcadosOcp < handle
             else
                 code_gen_opts = context.opts;
             end
-            context = setup_code_generation_context(ocp, context);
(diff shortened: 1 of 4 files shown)


interfaces/acados_matlab_octave/AcadosOcp.m
```diff
@@ -899,14 +899,26 @@ function make_consistent(self, is_mocp_phase)
...
             end
 
+            if ignore_initial && ignore_terminal
+                idxs = [2];
+            elseif ignore_terminal
+                idxs = [1, 2];
+            elseif ignore_initial
+                idxs = [2, 3];
+            else
+                idxs = [1, 2, 3]
```
FreyJo, line 909: Maybe, rename this, as `idxs` is used for the slack indices.

FreyJo (approved):

PR #1390, opened 2024-12-02
MATLAB: check values in `idxbx[_0,_e]` and `idxbu`

josipkh:
This PR adds a check for the indices related to simple bounds.


FreyJo (approved):
Thanks for the PR!
Indeed the initial state constraint for the last phase was not defined correctly in the example. I just pushed a fix.
If all tests past now, we can merge :)

josipkh:
Seems to work fine now, thanks for the quick update :)

PR #1393, opened 2024-12-04
Fix detect constraint for python interface

aghezz1:
After using in on more complicated examples, I realized there were few bugs that I fixed here.

- Fixing a bug in raising warning for GN Hessian


sandmaennchen (approved):
Thanks Andrea!

PR #1394, opened 2024-12-11
MATLAB: fix getter for QP values at terminal stage

sandmaennchen:
- Fix getter for QP matrices and vectors at the terminal stage.
- Fix `dump_last_qp_to_json` to include constraint-related terms at terminal stage.

### DIFF
--- a/interfaces/acados_matlab_octave/AcadosOcpSolver.m
+++ b/interfaces/acados_matlab_octave/AcadosOcpSolver.m
@@ -120,14 +120,14 @@ classdef AcadosOcpSolver < handle
                     return;
                 else
                     value = cell(obj.ocp.solver_options.N_horizon, 1);
-                    for n=0:obj.ocp.solver_options.N_horizon
+                    for n=0:(obj.ocp.solver_options.N_horizon-1)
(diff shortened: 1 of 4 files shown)


interfaces/acados_matlab_octave/AcadosOcpSolver.m
```diff
@@ -112,22 +112,28 @@ function set(obj, field, value, varargin)
...
 
-                    value = [R, S; S', Q];
+                    if n < obj.ocp.solver_options.N_horizon
+                        Q = obj.get('qp_Q', n);
+                        R = obj.get('qp_R', n);
+                        S = obj.get('qp_S', n);
+
+                        value = [R, S; S', Q];
+                    else
+                        value = obj.get('qp_Q', n);
```
david0oo, line 123: What happens, if someone requests `'qp_R'` on the last stage?

sandmaennchen: this is what I changed with the additional if here: `if n < obj.ocp.solver_options.N_horizon`

david0oo (approved):

PR #1395, opened 2024-12-11
C interface/Python: Getter for numerical values

sandmaennchen:
This PR adds getters for numerical values in cost and constraint modules and interfaces them from Python.

Adds the following functions to `AcadosOcpSolver`:
- `cost_get(stage, field)`
- `constraints_get(stage, field)`


examples/acados_python/max_iter_test/max_iter_test.py
```diff
@@ -106,6 +106,13 @@ def create_acados_solver_and_solve_problem(globalization='FIXED_STEP', solver_ty
 
     status = ocp_solver.solve()
 
+    # test getter
+    Vu_ = ocp_solver.cost_get(1, "Vu")
+    Vx_ = ocp_solver.cost_get(1, "Vx")
```
david0oo, line 111: Can we test the other cost getters that were newly added as well?

david0oo: I think we need an example that has an external cost and an example with NLS.

examples/acados_python/pendulum_on_cart/ocp/example_sqp_rti_loop.py
```diff
@@ -145,6 +145,9 @@ def main():
     cost = ocp_solver.get_cost()
     print("cost function value of solution = ", cost)
 
+    # test getter
```
david0oo, line 148: Same for constraints_get? Can we add more tests for all the new getters?

david0oo (changes requested):

examples/acados_python/pendulum_on_cart/ocp/example_sqp_rti_loop.py
sandmaennchen: will be tested by evaluator :smile:, cf https://github.com/acados/acados/pull/1402

PR #1401, opened 2024-12-13
C templates/MATLAB/Python: OCPs with `p_global` and no `global_data` 

sandmaennchen:
This PR allows OCPs for which global parameters are defined, but no CasADi function depends on them, i.e. `global_data` is empty.
In addition return status 0 in `precompute_dependencies`


FreyJo (approved):
Only works with new CasADi as p_global in general. So, all good :+1:

PR #1415, opened 2025-01-14
New option ext_fun_expand

jgillis:
The expand option transforms MX to SX without need for remodeling.

### DIFF
--- a/interfaces/acados_matlab_octave/GenerateContext.m
+++ b/interfaces/acados_matlab_octave/GenerateContext.m
@@ -40,10 +40,11 @@ classdef GenerateContext < handle
         casadi_fun_opts
         global_data_sym
         global_data_expr
+        solver_options
     end
 
     methods
-        function obj = GenerateContext(p_global, problem_name, opts)
+        function obj = GenerateContext(p_global, problem_name, opts, solver_options)
             if nargin < 3
                 opts = [];
             end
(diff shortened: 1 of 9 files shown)


interfaces/acados_matlab_octave/GenerateContext.m
```diff
@@ -229,6 +231,12 @@
                     rethrow(e);
                 end
 
+                if ~contains(name,'p_global_precompute')
```
sandmaennchen, line 234: Why not expanding the precompute function? And can all functions be expanded, i.e. should we add a try-catch block?

jgillis: I must admit this fix was overly specific to the project I was working on.
In that project, having expand for cost, constraints, ... was desirable while that for p_global_precompute was not.
Better to offer a more fine-grained control.

Prior to casadi 3.7, Functions could throw an error when expanding. From 3.7, it will always succeed.
This does not imply that `expand` is now invariably a good idea. Just as before, it may be a great idea or terrible idea, depending on the model.

FreyJo: Thanks for the clarification!

I guess if it is beneficial to expand a function, it is likely also the case for its derivative, right?
In this case, I would suggest to add options to set whether the casadi functions corresponding to cost, constraints, dynamics and the precompute functions should be expanded.
They could be called `ext_fun_casadi_expand_cost`, `ext_fun_casadi_expand_dyn`, ...

In addition we should add the try-catch block and refer to casadi 3.7, if this fails.

PR #1428, opened 2025-02-04
QP warm-start in first NLP solver iteration

FreyJo:
Implement option `nlp_solver_warm_start_first_qp_from_nlp`, if True first QP will be initialized using values from NLP iterate, otherwise from previous QP solution, only relevant if `nlp_solver_warm_start_first_qp` is True. Default: False.

The new initialization (with `nlp_solver_warm_start_first_qp = True`) is likely beneficial.
We might change the default behavior after more testing.

Fixed option `warm_start_first_qp` for SQP-RTI.

### DIFF
--- a/acados/ocp_nlp/ocp_nlp_sqp.c
+++ b/acados/ocp_nlp/ocp_nlp_sqp.c
@@ -724,11 +724,19 @@ int ocp_nlp_sqp(void *config_, void *dims_, void *nlp_in_, void *nlp_out_,
 
 
         /* solve QP */
-        // (typically) no warm start at first iteration
-        if (sqp_iter == 0 && !opts->warm_start_first_qp)
+        // warm start of first QP
+        if (sqp_iter == 0)
         {
-            int tmp_int = 0;
-            qp_solver->opts_set(qp_solver, nlp_opts->qp_solver_opts, "warm_start", &tmp_int);
+            // (typically) no warm start at first iteration
+            if (!opts->warm_start_first_qp)
+            {
+                int tmp_int = 0;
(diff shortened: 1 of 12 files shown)


acados/ocp_nlp/ocp_nlp_sqp.c
```diff
@@ -724,11 +724,19 @@ int ocp_nlp_sqp(void *config_, void *dims_, void *nlp_in_, void *nlp_out_,
...
-            qp_solver->opts_set(qp_solver, nlp_opts->qp_solver_opts, "warm_start", &tmp_int);
+            // (typically) no warm start at first iteration
+            if (!opts->warm_start_first_qp)
+            {
+                int tmp_int = 0;
+                qp_solver->opts_set(qp_solver, nlp_opts->qp_solver_opts, "warm_start", &tmp_int);
+            }
+            else
+            {
+                ocp_nlp_initialize_qp_from_nlp(config, dims, qp_in, nlp_out, qp_out);
```
giaf, line 738: If I get it correctly, with these changes if it is set `warm_start_first_qp`, the QP would be warm started with the NLP, while before it was so with the last QP solution. They are kind of equivalent only for at SQP convergence. I would leave the old behavior by default and put the `ocp_nlp_initialize_qp_from_nlp` under the control of a new flag `if(opts->warm_start_first_qp_from_nlp)` or something like that. So by default nothing is changed compared to the current behavior, and the new feature is enabled only where needed.

giaf: So `ocp_nlp_initialize_qp_from_nlp` would be executed only if both `opts->warm_start_first_qp` and `opts->warm_start_first_qp_from_nlp` are true.

giaf:
Looks good!

Issue #912, opened 2023-02-09
Solution Sensitivity incorrect if solver is initialized at the solution


andretag:
I am wondering if there is a bug in the output of `eval_param_sens`, because when I evaluate the sensitivity of the solution found at the origin ([0., 0,], considering an MPC for a double integrator), I obtain a value that does not match the slope of the solution found solving the full mpc problem in the neighborhood of that point. 

A plot highlighting the issue is attached, showing the optimal `u` from MPC ("full mpc solution") as a function of one of the two components of the state (`x_1`, while `x_2=0`) and the slope obtained from the sensitivity at the origin of the state space.

<img width="394" alt="Screen Shot 2023-02-09 at 10 48 25" src="https://user-images.githubusercontent.com/16794027/217863854-6b32c657-e302-4638-bb9b-86c1347ec848.png">

The issue does not appear at other points: 
<img width="386" alt="Screen Shot 2023-02-09 at 10 54 39" src="https://user-images.githubusercontent.com/16794027/217865519-ee0ad903-e5cb-4af4-aaa8-2412bf4c17b2.png">
 
Attached you can find the Jupiter notebook used to generate the plots (Updated: fix missing include)
[acados_sensitivity_test.ipynb.zip](https://github.com/acados/acados/files/10699143/acados_sensitivity_test.ipynb.zip)

Happy to open a PR to fix, if pointed in the right direction.

giaf:
The sensitivity is based on the latest internal linearization of the KKT conditions performed in HPIPM.
If the last QP did not converge to high accuracy, the sensitivity may be inaccurate.

In this case, the system to be controlled is a double integrator with reference [0,0].
When the MPC problem is solved with x0=[0,0], what is it happening? Are some QPs still solved? To what accuracy?

FreyJo:
I had a chat with @giaf and we found what the issue is. I'll try to summarize:

The problem in your example is, that the initial guess is already the solution of the OCP in the zero position.
Thus, in SQP the residuals fulfill the tolerance before a QP is solved.
However, `eval_param_sens` function assumes:
1. that the problem is solved for the current data and
2. that HPIPM has a factorized KKT system at the solution in memory.
In your example, the latter is not the case.
You can verify this, by just removing the `load_iterate` in your example, in this case you don't initialize at the solution and the sensitivities are correct.
Note, that this is a very special case in practice.

[...]

PR #1432, opened 2025-02-07
Work on sensitivity solver: add `setup_qp_matrices_and_factorize` and more

FreyJo:
### Added `setup_qp_matrices_and_factorize()`:
```
This function sets up the QP and factorizes Hessian matrix.

The current NLP iterate is used to setup the QP iterate, which includes t-slacks.
The t and lambda values are clipped according to `solution_sens_qp_t_lam_min` to avoid ill-conditioning.
Then the Hessian matrix is factorized.

If a two-solver approach is used to obtain solution sensitivities from acados, this function should be called before calling `eval_solution_sensitivity`, or `eval_adjoint_solution_sensitivity()`.

This is only implemented for HPIPM QP solver without condensing.
```

### Key differences of using `setup_qp_matrices_and_factorize()` vs. `solve()` in the two-solver approach:
- The QP solver does not need to perform multiple IP iterations to converge and only a single Riccati factorization is performed at the current iterate.
- In the chain example this reduces the "Median time to factorize exact Hessian QP:" 450.000 ms to 117.000 ms -> roughly 4x faster :rocket: 
- In the `solve()`-approach, the Riccati factorization at the second last QP iterate is used, as the system is not factorized at the final QP iterate. However, for the final iterate the system of equations becomes typically ill-conditioned, which is why the final iterate is modified, by clipping the inequality multipliers and t slacks, see option `solution_sens_qp_t_lam_min`.
- Less options that have to be set. The options that were needed to use the `solve` approach are now commented:
```
        # Old settings needed, when calling solve() instead of setup_qp_matrices_and_factorize()
        # ocp.solver_options.globalization_fixed_step_length = 0.0 # to not perfrom an SQP step update
        # ocp.solver_options.nlp_solver_max_iter = 1
        # ocp.solver_options.qp_solver_iter_max = 200
        # ocp.solver_options.tol = 1e-10  # to "force" a QP solve
```

### How it works
The option  `update_fact_exit` was added in HPIPM. This option being active, plus the warm start, plus limiting the QP solver iterations to 0, results in the correct behavior.

### Other things that changed in this PR
- fixed warm-starting QP from NLP iterate, was not correctly implemented in #1428, as the QP iterate needs to be condensed and eliminated degrees of freedom need to be taken into account. Only for N==N2 condensing right now, to be extended https://github.com/acados/acados/issues/1433


sandmaennchen:
Maybe add to the description what the option `update_fact_exit` does?

examples/acados_python/pendulum_on_cart/solution_sensitivities/forw_vs_adj_param_sens.py
```diff
@@ -192,5 +238,9 @@ def main(qp_solver_ric_alg: int, use_cython=False, generate_solvers=True, plot_t
         plot_pendulum(ocp.solver_options.shooting_nodes, Fmax, simU, simX, latexify=True, time_label=ocp.model.t_label, x_labels=ocp.model.x_labels, u_labels=ocp.model.u_labels)
 
 
+def test_failure_message(msg):
+    print(f"ERROR: {msg}")
+    # raise Exception(msg)
+
```
sandmaennchen, line 244: Change back to exception?

interfaces/acados_template/acados_template/c_templates_tera/acados_solver.in.c
```diff
@@ -2771,6 +2774,17 @@ int {{ model.name }}_acados_solve({{ model.name }}_solver_capsule* capsule)
 }
 
 
+
+int {{ model.name }}_acados_setup_qp_matrices_and_factorize({{ model.name }}_solver_capsule* capsule)
+{
+    // solve NLP
```
sandmaennchen, line 2780: Remove comment

sandmaennchen (approved):

acados/ocp_nlp/ocp_nlp_common.c
```diff
@@ -3229,6 +3240,76 @@ void ocp_nlp_cost_compute(ocp_nlp_config *config, ocp_nlp_dims *dims, ocp_nlp_in
...
+    config->qp_solver->opts_set(config->qp_solver, nlp_opts->qp_solver_opts, "t_min", &tmp_double);
+    config->qp_solver->opts_set(config->qp_solver, nlp_opts->qp_solver_opts, "lam_min", &tmp_double);
+
+    // QP solve
+    qp_status = ocp_nlp_solve_qp_and_correct_dual(config, dims, nlp_opts, nlp_mem, nlp_work, false, NULL, NULL, NULL);
+
+    // reset QP solver settings
+    qp_solver->opts_set(qp_solver, nlp_opts->qp_solver_opts, "warm_start", &nlp_opts->qp_warm_start);
+    qp_solver->opts_set(qp_solver, nlp_opts->qp_solver_opts, "iter_max", &nlp_opts->qp_iter_max);
+    tmp_double = 1e-16;
```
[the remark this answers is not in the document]
FreyJo: True. However
- a lot of boilerplate code is missing to get options to back them up.
- the default values for `t_min` and `lam_min` are consistent across all HPIPM modes and unchanged > 5 years.
- these values can not be changed by the user through high level interfaces
Thus, I find the extra work here to do it differently not worth right now.
I added a comment.

FreyJo:
> Maybe add to the description what the option `update_fact_exit` does?

It is mentioned under "How it works" in the PR description.
I addressed all other comments. Thanks for the quick review :pray:

giaf:
I see the point of having a separate function for the second solve, mainly to hide details of the second solver setup to the user.
But I'm not too convinced about the name `setup_qp_matrices_and_factorize`.
What about something like `hotstart_and_factorize_kkt`, IMO it tells more about the key point: this second solver is hotstarted at the provided solution guess  (no iterations are performed), and the KKT system is factorized at that point, so ready to be used for e.g. sensitivity (or also Riccati feedback matrices) computations.

giaf:
In HPIPM I added `lam0_min` and `t0_min` (with setters and getters) for use only in variables initialization https://github.com/giaf/hpipm/commit/f5ca1caa7b2af0b4adab430ead66b76c4508a2da .
So you should now use these ones instead of `lam_min` and `t_min`.
You can use the getters to have local copies of these values before you overwrite them.
Getters for additional opts can be easily added.

FreyJo:
> I see the point of having a separate function for the second solve, mainly to hide details of the second solver setup to the user.

Another reason that `solve()` might exit without a factorization, see https://github.com/acados/acados/issues/912

I much prefer the name `setup_qp_matrices_and_factorize` to `hotstart_and_factorize_kkt`, as creating the QP with an exact Hessian is the more computationally expensive step compared to initializing the QP solver.
In the end, not all details of the function are in the function name, but the docstring describes also the hotstart with the modified iterate.
Also, `hotstart` does more suggest that some iterations are happening, which they are not in this routine as you wrote.

Regarding the option getters. Since you both mentioned it, and @giaf  implemented in HPIPM, I will follow up in this PR accordingly.

PR #1439, opened 2025-02-13
Solve relaxed KKT conditions and compute corresponding sensitivities

FreyJo:
This PR allows one to solve relaxed KKT conditions and compute corresponding sensitivities.

## Example
The newly added example regards the example
minimize $(x-\theta^2)^2$ subject to $-1 \leq x\leq 1$, where $\theta$ is modelled as `p_global`
Results are shown in picture below.
basic_example

## Detailed Changes
- Add option `tau_min`:
```
Minimum value for the barrier parameter tau.
Relevant if an interior point method is used as a (sub)solver, right now this is only HPIPM.
If no interior point method is used, this is set to 0.
Default: 0
```
- Add option `qp_solver_t0_init`:
```
For HPIPM QP solver: Initialization scheme of lambda and t slacks within HPIPM.
0: initialize with sqrt(mu0)
1: initialize with 1.0
2: heuristic for primal feasibility
When using larger value for tau_min, it is beneficial to not use 2, as the initialization of (t, lambda) might be too far off from the central path and prevent convergence.
Type: int > 0
Default: 2
```
- The complementarity residual `res_comp` is now computed as `inf_norm(lam_i * ineq_fun_i - tau_min * ones)`. This is backwards compatible as `tau_min` was zero before.
- Update HPIPM
- Generalize `condense_qp_out` to handle `nbxe[0] = 0`


FreyJo:
The option is not just specific to the QP solver, but is also used in the residual computation of the NLP solvers.
In addition, this option might be used when implementing an interior point NLP solver in acados.

sandmaennchen (approved):

PR #1440, opened 2025-02-14
Update `print_iterations`

david0oo:
The printing of iterations was made more versatile such that it can be adjusted more easily for different solvers, different globalization strategies, or other future developments.

- the printing of # it, rest_stat, res_eq, res_ineq, res_comp was moved to `ocp_nlp_common.c`
- solver specific stuff like step_norm, lm_reg., qp_status, and qp_iter was moved to `ocp_nlp_sqp` and `ocp_nlp_ddp`
- globalization specific stuff was moved to the globalization modules

The print_iteration function is now called within the solver file, and then internally the different print functions of the different print modules are called sequentially. Also the iteration log has now a unified appearance.


FreyJo (approved):
Great! Modular printing :ok_hand:

PR #1447, opened 2025-02-26
Improve docstrings for warm start options

Confectio:
@FreyJo Would this be ok?


FreyJo (approved):
Thanks for the PR!
I revised the docstings a bit. If you agree with the changes, I'll merge :)

Confectio:
Yes, I agree :)

PR #1457, opened 2025-03-05
Condensing: call `d_cond_qp_expand_sol` with options as `d_cond_qp_expand_primal_sol` was removed in HPIPM

FreyJo:
- Plus update HPIPM
- See https://github.com/giaf/hpipm/commit/f49620429d79146a3a891f618ff4816f04a9481b


giaf:
To what I could see looking at the code, this should make the job with how default opts are set. I haven't tested though.

PR #1460, opened 2025-03-06
Modify `AcadosOcpBatchSolver` to use flexible batch size

FreyJo:


interfaces/acados_template/acados_template/acados_ocp_batch_solver.py
```diff
@@ -350,3 +364,13 @@ def load_iterate_from_flat_obj(self, iterate: AcadosOcpFlattenedBatchIterate) ->
         self.set_flat("su", iterate.su)
         self.set_flat("pi", iterate.pi)
         self.set_flat("lam", iterate.lam)
+
+    def __check_n_batch(self, n_batch: Optional[int], optional=True) -> int:
```
sandmaennchen, line 368: Is the `optional` argument really needed? I think you only need it in the sensitivity functions where `n_batch` cannot become `None`

sandmaennchen (approved):

PR #1463, opened 2025-03-14
add QP scaling for cost and constraints

FreyJo:
This PR adds the option to scale the cost and constraints of QPs.
In particular, the following options have been added:
- `qpscaling_scale_constraints`:  String in ["NO_OBJECTIVE_SCALING", "OBJECTIVE_GERSHGORIN"]
        Default: "NO_OBJECTIVE_SCALING".
        - `NO_OBJECTIVE_SCALING`: no scaling of the objective
        - `OBJECTIVE_GERSHGORIN`: estimate max. abs. eigenvalue using Gershgorin circles as `max_abs_eig`, then sets the objective scaling factor as `obj_factor = min(1.0, qpscaling_ub_max_abs_eig/max_abs_eig)`

- `qpscaling_scale_constraints`
        String in ["NO_CONSTRAINT_SCALING", "INF_NORM"]
        Default: "NO_CONSTRAINT_SCALING".
        - `NO_CONSTRAINT_SCALING`: no scaling of the constraints
        - `INF_NORM`: scales each constraint except simple bounds by factor `1.0 / max(inf_norm_coeff, inf_norm_constraint_bound)`, such that the inf-norm of the constraint coefficients and bounds is <= 1.0.
        Slack penalties are adjusted accordingly to get an equivalent solution.
        First, the cost is scaled, then the constraints.

## Implementation details
Minimal computational overhead is associated with the new scaling, in case it is not used. The implementation avoids copying values from the different data structures if they are the same and aliases pointers instead.

## Examples & Tests:
- equivalence of primal and dual solutions is tested on different examples including slack variables and one-sided constraints
- some tests were added, showcasing examples which cannot be solved without scaling the QP.

## Limitations
- Real-time iterations, i.e. RTI and AS-RTI variants with split preparation and feedback phase are not compatible with QP scaling yet. Work on this has been started at https://github.com/FreyJo/acados/pull/76 but is not a priority right now.
- Computation of solution sensitivities is not compatible with QP scaling yet.
- Scaling tolerances of the QP solver is not done. Right now, the QP solver tolerances need to be set small enough, for the NLP solver to converge. A follow-up work is described in https://github.com/acados/acados/issues/1553


.gitignore
```diff
@@ -93,7 +93,7 @@ env/
 
 # Examples #
 c_generated_code
-*_generated_code
+*_generated_code*
```
sandmaennchen, line 96: :heart:

acados/utils/types.h
```diff
@@ -81,6 +81,7 @@ enum return_values
     ACADOS_READY = 5,
     ACADOS_UNBOUNDED = 6,
     ACADOS_TIMEOUT = 7,
+    ACADOS_QPSCALING_BOUNDS_NOT_SATISFIED = 8,
```
sandmaennchen, line 84: add to PR description and doc string

examples/acados_python/hock_schittkowsky/hs074_constraint_scaling.py
```diff
@@ -0,0 +1,125 @@
...
+    ocp.solver_options.N_horizon = 0
+    ocp.solver_options.qp_solver = 'PARTIAL_CONDENSING_HPIPM'
+    ocp.solver_options.qp_solver_mu0 = 1e3
+    ocp.solver_options.hessian_approx = 'EXACT'
+    ocp.solver_options.regularize_method = 'MIRROR'
+    ocp.solver_options.print_level = 1
+    ocp.solver_options.nlp_solver_max_iter = 1000
+    ocp.solver_options.qp_solver_iter_max = 1000
+
+    # Search direction
```
sandmaennchen: weird comment?

examples/acados_python/linear_mass_model/test_qpscaling_slacked.py
```diff
@@ -0,0 +1,263 @@
...
+    sol_1 = ocp_solver1.store_iterate_to_obj()
+
+    # test QP scaling
+    ocp_2, ocp_solver2 = create_solver("2", soften_obstacle, soften_terminal, soften_controls, nlp_solver_type="SQP", allow_switching_modes=False, use_qp_scaling=True)
+    sol_2 = call_solver(ocp_2, ocp_solver2, soften_obstacle, soften_terminal, soften_controls, plot=False)
+    check_qp_scaling(ocp_solver2)
+    sol_2 = ocp_solver2.store_iterate_to_obj()
+    ocp_solver2.get_from_qp_in(1, "idxs_rev")
+
+    breakpoint()
```
sandmaennchen: remove?

examples/acados_python/linear_mass_model/test_qpscaling_slacked.py
```diff
@@ -0,0 +1,263 @@
...
+        for i in range(len(v1)):
+            if not np.allclose(v1[i], v2[i], atol=1e-6):
+                print(f"Field {field} differs at index {i}: max diff = {np.max(np.abs(v1[i] - v2[i]))}")
+                print(f"got difference {v1[i] - v2[i]}")
+            else:
+                pass
+                # print(f"Field {field} is the same at index {i}.")
+
+
+    if sol_1.allclose(sol_2):
```
sandmaennchen: with this check, why do you additionally need the checks above?

interfaces/acados_matlab_octave/AcadosOcpOptions.m
```diff
@@ -187,6 +193,10 @@
             obj.print_level = 0;
             obj.cost_discretization = 'EULER';
             obj.regularize_method = 'NO_REGULARIZE';
+            obj.qpscaling_ub_max_abs_eig = 1e5;
+            obj.qpscaling_lb_norm_inf_grad_obj = 1e-4;
+            obj.qpscaling_scale_objective = false;
```
sandmaennchen: I would keep this consistent with python

sandmaennchen (approved):

examples/acados_python/linear_mass_model/test_qpscaling_slacked.py
FreyJo: Not needed indeed. Added a comment. But for testing more output was useful, so kept the additional check for now.

PR #1468, opened 2025-03-17
Implement more flexible code reuse in Matlab & Python

FreyJo:
## MATLAB: implement code reuse
- add `solver_creation_opts` argument to `AcadosOcpSolver`
- with
```
default_solver_creation_opts = struct('json_file', '', ...
                    'build', true, ...
                    'generate', true, ...
                    'compile_mex_wrapper', true, ...
                    'compile_interface', [], ...
                    'output_dir', fullfile(pwd, 'build'));
```
- `solver_creation_opts` can be used to specify some of those options. For the others, the default is used.

## MATLAB and Python: implement code reuse without specifying OCP formulation
- allow creating `AcadosOcpSolver` with `acados_ocp=None` and just provide json file instead
- in this case generated code must also be available
- load information from json file that is needed within the solver; NOTE: we don't recover the full OCP description as class with subclasses, but only load what is needed in form of dicts / structs. This is less restrictive and should make it easier to use solvers generated with Python in Matlab and vice versa.


interfaces/acados_matlab_octave/AcadosOcpSolver.m
```diff
@@ -31,51 +31,116 @@
...
+        problem_class
+        name
+        has_x0
+        nsbu_0
+        nbxe_0
+    end
     methods
 
-        function obj = AcadosOcpSolver(ocp, output_dir)
+        function obj = AcadosOcpSolver(ocp, varargin)
```
sandmaennchen, line 54: maybe add docstring here, arguments are different from python

interfaces/acados_matlab_octave/AcadosOcpSolver.m
```diff
@@ -31,51 +31,116 @@
...
-        qp_gettable_fields = {'qp_Q', 'qp_R', 'qp_S', 'qp_q', 'qp_r', 'qp_A', 'qp_B', 'qp_b', 'qp_C', 'qp_D', 'qp_lg', 'qp_ug', 'qp_lbx', 'qp_ubx', 'qp_lbu', 'qp_ubu', 'qp_zl', 'qp_zu', 'qp_Zl', 'qp_Zu'}
     end % properties
 
+    properties (Access = private)
+        fields = {'x', 'u', 'z', 'sl', 'su', 'lam', 'pi'};
+        qp_gettable_fields = {'qp_Q', 'qp_R', 'qp_S', 'qp_q', 'qp_r', 'qp_A', 'qp_B', 'qp_b', 'qp_C', 'qp_D', 'qp_lg', 'qp_ug', 'qp_lbx', 'qp_ubx', 'qp_lbu', 'qp_ubu', 'qp_zl', 'qp_zu', 'qp_Zl', 'qp_Zu'}
+        t_ocp % templated solver
 
+        % required info loaded from json
+        N
```
sandmaennchen, line 44: N_horizon ?

interfaces/acados_template/acados_template/acados_ocp_solver.py
```diff
@@ -195,7 +195,15 @@ def create_cython_solver(cls, json_file):
     def save_p_global(self) -> bool:
         return self.__save_p_global
 
-    def __init__(self, acados_ocp: Union[AcadosOcp, AcadosMultiphaseOcp], json_file=None, simulink_opts=None, build=True, generate=True, cmake_builder: CMakeBuilder = None, verbose=True, save_p_global=False):
+    @property
+    def N(self) -> int:
```
sandmaennchen, line 199: N_horizon?

sandmaennchen (approved):

interfaces/acados_template/acados_template/acados_ocp_solver.py
FreyJo: this field was already there in Python, just not as a property.
It is also used in an example.
We can rename it, but it is breaking at this point.

interfaces/acados_matlab_octave/AcadosOcpSolver.m
FreyJo: addressed in https://github.com/acados/acados/pull/1472/commits/b15ef106fc9110fce6b4a29a78636f5fe4067db9

interfaces/acados_matlab_octave/AcadosOcpSolver.m
FreyJo: Addressed in https://github.com/acados/acados/pull/1472/commits/8945621d1018207ad08657491b59423e8662e96a

FreyJo:
> not sure if I like the duplication of the options in the solver 🙈

I see the point.
However, it is easier to do it like this for now. Otherwise we need to implement casts in the setters, or don't use the setters. Any type incompatibility would break the possibility of loading the solver.

Closing this one, as I would directly merge https://github.com/acados/acados/pull/1472 instead, which contains only minor changes on top of this regarding the interfaces. However, there are many more tests which ensure interoperability between MATLAB and Python :)

PR #1462, opened 2025-03-12
Implement adaptive epsilon for regularizations: `MIRROR` and `PROJECT`

FreyJo:
New options:
- `reg_max_cond_block`: Maximum condition number per block when using regularize_method in ['PROJECT', 'MIRROR'] with `reg_adaptive_eps` = True, Default: 1e-7
- `reg_adaptive_eps`: Determines if epsilon is chosen adaptively in regularization used if regularize_method in ['PROJECT', 'MIRROR']
If true, epsilon is chosen block-wise based on `reg_max_cond_block`.
Otherwise, epsilon is chosen globally based on `reg_epsilon`.
Default: False

Inspired by "A nonmonotone filter SQP method: local convergence and numerical results" by NICHOLAS I. M. GOULD, YUELING LOH,  DANIEL P. ROBINSON, Sec. 5.1

PR #1471, opened 2025-03-19
Follow up to #1462

FreyJo:
- fix test, run test on CI
- add option: `reg_min_epsilon`: Minimum value for epsilon if regularize_method in ['PROJECT', 'MIRROR'] is used with reg_adaptive_eps. Default: 1e-8.
- fix `reg_max_cond_block` default: 1e7 instead of 1e-7. Add check that it must be >= 1.0.


david0oo (approved):
Looks good. Keep up the good work!

Issue #1433, opened 2025-02-10
Implement `condense_qp_out` for real condensing.


FreyJo:
Currently only implemented for N=N2.
Introduced in https://github.com/acados/acados/pull/1432

For full condensing, it should be implemented here:
https://github.com/acados/acados/blob/7de086d1807b18af3d4b3e627ba90e524454a780/acados/ocp_qp/ocp_qp_full_condensing.c#L487-L492

For partial condensing, here: https://github.com/acados/acados/blob/7de086d1807b18af3d4b3e627ba90e524454a780/acados/ocp_qp/ocp_qp_partial_condensing.c#L542

PR #1474, opened 2025-03-21
Condense qp out

FreyJo:
Closes https://github.com/acados/acados/issues/1433

Also:
- fix setting warm_start_first_qp_from_nlp in RTI
- Python: allow updating warm start options on the fly


giaf:
Looking at the C code it seems all good.
Did you run some tests using the new features?

sandmaennchen (approved):

PR #1476, opened 2025-03-24
Python: Implement reformulation as external cost with Gauss-Newton for LS/NLS

FreyJo:
This is in particularly useful when computing parametric solution sensitivities with a two-solver approach.
It allows to use the same parameterization of the problem in both solvers, but to use a Gauss-Newton Hessian in the "nominal" solver.

Note: for linear LS cost, there is nothing to do, as the exact Hessian and its GN approximation coincide.


dirkpr (approved):
Good stuff!

PR #1477, opened 2025-03-24
Python: checks on external checks

FreyJo:
Breaking: `cost_expr_ext_cost_custom_hess*` can not be float or int anymore in Python, but should be a CasADi MX/SX/DM


sandmaennchen (approved):
At some point, we should implement all of the model attributes as properties and move the type checks to the setters ..

PR #1479, opened 2025-03-25
Symbolic cost parameters

dirkpr:
## Enhancement: Symbolic Expressions for Cost Matrices in acados Python Interface
This PR extends the Python interface to acados by enabling symbolic expressions for the following cost matrices and reference vectors:
- `W_0` (initial stage cost weight matrix)
- `W` (stage cost weight matrix)
- `W_e` (terminal cost weight matrix) 
- `yref_0` (initial stage reference vector)
- `yref` (stage reference vector)
- `yref_e` (terminal reference vector)

### Benefits
This enhancement allows for parameterized least-squares cost functions. The least-squares cost is translated to an external cost, and parameters can be dynamically modified at runtime.

### Use Case
This is particularly valuable for advanced two-solver implementations where:
1. A forward solver leverages a Gauss-Newton Hessian approximation
2. A backward solver employs an exact Hessian for exact sensitivity computation

### DIFF
--- a/examples/acados_python/pendulum_on_cart/ocp/ocp_example_cost_formulations.py
+++ b/examples/acados_python/pendulum_on_cart/ocp/ocp_example_cost_formulations.py
@@ -204,7 +204,7 @@ def formulate_ocp(cost_version: str) -> AcadosOcp:
         ocp.cost.cost_type = 'NONLINEAR_LS'
         ocp.cost.cost_type_e = 'NONLINEAR_LS'
 
-        p_global =struct_symSX([
+        p_global = struct_symSX([
             entry('W', shape=(ny, ny)),
             entry('yref', shape=(ny, )),
             entry('W_e', shape=(ny_e, ny_e)),
--- a/examples/acados_python/pendulum_on_cart/ocp/ocp_example_cost_formulations.py
+++ b/examples/acados_python/pendulum_on_cart/ocp/ocp_example_cost_formulations.py
@@ -229,14 +229,6 @@ def formulate_ocp(cost_version: str) -> AcadosOcp:
         p_global_values['yref_e'] = np.zeros((ny_e, ))
 
         ocp.p_global_values = p_global_values.cat.full().flatten()
-
-        ocp.cost.W_0 = np.zeros((0,0))
-        ocp.cost.W   = np.zeros((0,0))
-        ocp.cost.W_e = np.zeros((0,0))
-
-        ocp.cost.yref_0 = np.array([])
-        ocp.cost.yref   = np.array([])
-        ocp.cost.yref_e = np.array([])
     else:
         raise Exception('Unknown cost_version.')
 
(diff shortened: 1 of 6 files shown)


examples/acados_python/pendulum_on_cart/ocp/ocp_example_cost_formulations.py
```diff
@@ -199,6 +200,43 @@ def formulate_ocp(cost_version: str) -> AcadosOcp:
...
+
+        ocp.p_global_values = p_global_values.cat.full().flatten()
+
+        ocp.cost.W_0 = np.zeros((0,0))
+        ocp.cost.W   = np.zeros((0,0))
+        ocp.cost.W_e = np.zeros((0,0))
+
+        ocp.cost.yref_0 = np.array([])
+        ocp.cost.yref   = np.array([])
+        ocp.cost.yref_e = np.array([])
```
FreyJo, line 239: I would suggest removing those symbolic fields in `translate_cost_to_external_cost()`.
They can also be removed if they are not symbolic as `W`, `yref` etc are not part of the `EXTERNAL` cost formulation.

interfaces/acados_template/acados_template/utils.py
```diff
@@ -419,6 +429,18 @@ def check_if_2d_nparray(val, name) -> None:
...
     return
 
+def check_if_2d_nparray_or_symbolic(val, name) -> None:
+    if not isinstance(val, (np.ndarray, SX, MX)):
+        raise Exception(f"{name} must be a array of type np.ndarray, casadi.SX, or casadi.MX, got {type(val)}")
+
+    if isinstance(val, (np.ndarray)):
+        if val.ndim != 2:
+            raise Exception(f"{name} must be a 2D array of type np.ndarray, casadi.SX, or casadi.MX, got shape {val.shape}")
+    else:
```
FreyJo, line 439: remove else, as casadi is always 2D.
I would suggest to refactor

```
def check_if_2d_nparray_or_casadi_symbolic(val, name) -> None:
    if isinstance(val, (SX, MX, DM)):
        return
    if not isinstance(val, np.ndarray):
        raise Exception(f"{name} must be a numpy array or casadi symbolic, got {type(val)}")
    if val.ndim != 2:
        raise Exception(f"{name} must be a 2D numpy array or casadi symbolic, got shape {val.shape}")
```

interfaces/acados_template/acados_template/acados_ocp_cost.py
```diff
@@ -169,12 +169,12 @@ def cost_ext_fun_type_0(self):
 
     @yref_0.setter
     def yref_0(self, yref_0):
-        yref_0 = check_if_nparray_and_flatten(yref_0, "yref_0")
+        yref_0 = check_if_nparray_or_symbolic_and_flatten(yref_0, "yref_0")
```
FreyJo, line 172: These loosened checks should be readded in `AcadosOcp.make_consistent`.
Something like:
```
if isinstance(yref, (ca.SX, ca.MX)):
    raise Exception("yref should be numpy array, symbolics are only supported before solver creation, to allow reformulating costs, e.g. using translate_cost_to_external_cost().")
```

examples/acados_python/pendulum_on_cart/ocp/ocp_example_cost_formulations.py
```diff
@@ -199,6 +200,43 @@ def formulate_ocp(cost_version: str) -> AcadosOcp:
         ocp.model.cost_y_expr = ca.vertcat(x, u)
         ocp.model.cost_y_expr_e = x
         ocp.translate_cost_to_external_cost(cost_hessian='GAUSS_NEWTON')
+    elif cost_version == 'NLS_TO_EXTERNAL_P_GLOBAL':
+        ocp.cost.cost_type = 'NONLINEAR_LS'
+        ocp.cost.cost_type_e = 'NONLINEAR_LS'
+
+        p_global =struct_symSX([
```
FreyJo, line 207: Not sure if using `struct_symSX` makes this example easier here. But no strong objection.

FreyJo (changes requested):

interfaces/acados_template/acados_template/utils.py
dirkpr: Added with f96afce7b9c0f674ebf2ee18d1efb2bd2a96206b

interfaces/acados_template/acados_template/acados_ocp_cost.py
dirkpr: Added with  9e12dd07a2f0c90acee2036f17d5eea7ddf7273e

FreyJo (approved):

PR #1483, opened 2025-03-27
Dynamic num threads in batch solvers

Confectio:
Previously, the number of threads to be used in the `AcadosOcpBatchSolver` or `AcadosSimBatchSolver` batched methods had to be set in the respective `AcadosOcpOptions` or `AcadosSimOptions` and could not be changed without rebuilding the solver.
Now, it is a property that can be set dynamically, even after building the solver.

The property `num_threads_in_batch_solve` in both `AcadosOcpOptions` and `AcadosSimOptions` is now deprecated,
and warnings will be printed if it is used.
Instead, the `AcadosOcpBatchSolver` and `AcadosSimBatchSolver` have a new (settable) property `num_threads_in_batch_solve` and take the initial value in their constructor via a new `num_threads_in_batch_solve` argument. 
The value of the new attribute is now used to determine how many threads the batched methods should use.

The property `num_threads_in_batch_solve` in `AcadosOcpOptions` or `AcadosSimOptions` was previously used for determining whether the solvers should be build with openmp and how the C code of the batched methods should be templated (if `num_threads_in_batch_solve > 1`).
Instead a new property `with_batch_functionality` is used as a flag to determine whether the solvers should be build with openmp and whether the C code of the batched methods should be created at all.
This new property is not required to be set by the user himself, but is being set by the BatchSolvers themselves by default (a warning is printed though, to notify the user that the code is still build with the openmp flag).

The matlab templates and the batch solver examples have been adjusted accordingly. Further, the minimal batch solver examples include new simple tests for the new `num_threads_in_batch_solve` property.

### DIFF
--- a/interfaces/acados_template/acados_template/acados_ocp_batch_solver.py
+++ b/interfaces/acados_template/acados_template/acados_ocp_batch_solver.py
@@ -129,6 +129,10 @@ class AcadosOcpBatchSolver():
     def num_threads_in_batch_solve(self):
         """Number of threads used for parallelizing the batch methods."""
         return self.__num_threads_in_batch_solve
+    
+    @num_threads_in_batch_solve.setter
+    def num_threads_in_batch_solve(self, num_threads_in_batch_solve):
+        self.__num_threads_in_batch_solve = num_threads_in_batch_solve
 
 
     def solve(self):
--- a/interfaces/acados_template/acados_template/acados_ocp_batch_solver.py
+++ b/interfaces/acados_template/acados_template/acados_ocp_batch_solver.py
@@ -76,19 +76,19 @@ class AcadosOcpBatchSolver():
         self.__status = np.zeros((self.N_batch,), dtype=np.intc, order="C")
         self.__status_p = cast(self.__status.ctypes.data, POINTER(c_int))
 
-        getattr(self.__shared_lib, f"{self.__name}_acados_batch_solve").argtypes = [POINTER(c_void_p), POINTER(c_int), c_int]
+        getattr(self.__shared_lib, f"{self.__name}_acados_batch_solve").argtypes = [POINTER(c_void_p), POINTER(c_int), c_int, c_int]
         getattr(self.__shared_lib, f"{self.__name}_acados_batch_solve").restype = c_void_p
 
-        getattr(self.__shared_lib, f"{self.__name}_acados_batch_eval_params_jac").argtypes = [POINTER(c_void_p), c_int]
+        getattr(self.__shared_lib, f"{self.__name}_acados_batch_eval_params_jac").argtypes = [POINTER(c_void_p), c_int, c_int]
         getattr(self.__shared_lib, f"{self.__name}_acados_batch_eval_params_jac").restype = c_void_p
 
-        getattr(self.__shared_lib, f"{self.__name}_acados_batch_eval_solution_sens_adj_p").argtypes = [POINTER(c_void_p), c_char_p, c_int, POINTER(c_double), c_int, c_int]
+        getattr(self.__shared_lib, f"{self.__name}_acados_batch_eval_solution_sens_adj_p").argtypes = [POINTER(c_void_p), c_char_p, c_int, POINTER(c_double), c_int, c_int, c_int]
         getattr(self.__shared_lib, f"{self.__name}_acados_batch_eval_solution_sens_adj_p").restype = c_void_p
 
-        getattr(self.__shared_lib, f"{self.__name}_acados_batch_set_flat").argtypes = [POINTER(c_void_p), c_char_p, POINTER(c_double), c_int, c_int]
+        getattr(self.__shared_lib, f"{self.__name}_acados_batch_set_flat").argtypes = [POINTER(c_void_p), c_char_p, POINTER(c_double), c_int, c_int, c_int]
         getattr(self.__shared_lib, f"{self.__name}_acados_batch_set_flat").restype = c_void_p
 
-        getattr(self.__shared_lib, f"{self.__name}_acados_batch_get_flat").argtypes = [POINTER(c_void_p), c_char_p, POINTER(c_double), c_int, c_int]
+        getattr(self.__shared_lib, f"{self.__name}_acados_batch_get_flat").argtypes = [POINTER(c_void_p), c_char_p, POINTER(c_double), c_int, c_int, c_int]
         getattr(self.__shared_lib, f"{self.__name}_acados_batch_get_flat").restype = c_void_p
 
         if self.ocp_solvers[0].acados_lib_uses_omp:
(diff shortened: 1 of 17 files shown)


interfaces/acados_template/acados_template/acados_ocp_batch_solver.py
```diff
@@ -111,14 +124,19 @@ def ocp_solvers(self):
     def N_batch(self):
         """Batch size."""
         return self.__N_batch
+    
+    @property
+    def num_threads_in_batch_solve(self):
+        """Number of threads used for parallelizing the batch methods."""
+        return self.__num_threads_in_batch_solve
```
FreyJo, line 131: It seems like the setter is missing to actually make it dynamic.
Can you add it and also test it in the example?

interfaces/acados_template/acados_template/c_templates_tera/acados_sim_solver.in.h
```diff
@@ -80,7 +80,9 @@ typedef struct {{ model.name }}_sim_solver_capsule
 
 ACADOS_SYMBOL_EXPORT int {{ model.name }}_acados_sim_create({{ model.name }}_sim_solver_capsule *capsule);
 ACADOS_SYMBOL_EXPORT int {{ model.name }}_acados_sim_solve({{ model.name }}_sim_solver_capsule *capsule);
-ACADOS_SYMBOL_EXPORT void {{ model.name }}_acados_sim_batch_solve({{ model.name }}_sim_solver_capsule **capsules, int N_batch);
+{% if solver_options.with_batch_functionality %}
+    ACADOS_SYMBOL_EXPORT void {{ model.name }}_acados_sim_batch_solve({{ model.name }}_sim_solver_capsule **capsules, int N_batch, int num_threads_in_batch_solve);
```
FreyJo, line 84: In this templated logic, we don't want to indent, such that the rendered code is aligned nicely.

interfaces/acados_template/acados_template/c_templates_tera/acados_solver.in.c
```diff
@@ -2817,143 +2815,161 @@ int {{ model.name }}_acados_setup_qp_matrices_and_factorize({{ model.name }}_sol
...
-void {{ model.name }}_acados_batch_solve({{ model.name }}_solver_capsule ** capsules, int * status_out, int N_batch)
-{
-{% if solver_options.num_threads_in_batch_solve > 1 %}
-    int num_threads_bkp = omp_get_num_threads();
-    omp_set_num_threads({{ solver_options.num_threads_in_batch_solve }});
-
-    #pragma omp parallel for
-{%- endif %}
-    for (int i = 0; i < N_batch; i++)
+{% if solver_options.with_batch_functionality %}
```
FreyJo, line 2818: Also here don't indent based on template logic.

interfaces/acados_template/acados_template/c_templates_tera/acados_solver.in.h
```diff
@@ -295,14 +295,15 @@ ACADOS_SYMBOL_EXPORT int {{ name }}_acados_set_p_global_and_precompute_dependenc
...
-
-ACADOS_SYMBOL_EXPORT void {{ model.name }}_acados_batch_eval_solution_sens_adj_p({{ model.name }}_solver_capsule ** capsules, const char *field, int stage, double *out, int offset, int N_batch);
-ACADOS_SYMBOL_EXPORT void {{ model.name }}_acados_batch_eval_params_jac({{ model.name }}_solver_capsule ** capsules, int N_batch);
+    ACADOS_SYMBOL_EXPORT void {{ model.name }}_acados_batch_set_flat({{ model.name }}_solver_capsule ** capsules, const char *field, double *data, int N_data, int N_batch, int num_threads_in_batch_solve);
+    ACADOS_SYMBOL_EXPORT void {{ model.name }}_acados_batch_get_flat({{ model.name }}_solver_capsule ** capsules, const char *field, double *data, int N_data, int N_batch, int num_threads_in_batch_solve);
 
+    ACADOS_SYMBOL_EXPORT void {{ model.name }}_acados_batch_eval_solution_sens_adj_p({{ model.name }}_solver_capsule ** capsules, const char *field, int stage, double *out, int offset, int N_batch, int num_threads_in_batch_solve);
+    ACADOS_SYMBOL_EXPORT void {{ model.name }}_acados_batch_eval_params_jac({{ model.name }}_solver_capsule ** capsules, int N_batch, int num_threads_in_batch_solve);
+{% endif %}
```
FreyJo, line 306: Here too

interfaces/acados_template/acados_template/acados_ocp_batch_solver.py
```diff
@@ -51,10 +52,22 @@ class AcadosOcpBatchSolver():
...
 
-    def __init__(self, ocp: AcadosOcp, N_batch: int, json_file: str = 'acados_ocp.json',  build: bool = True, generate: bool = True, verbose: bool=True):
+    def __init__(self, ocp: AcadosOcp, N_batch: int, num_threads_in_batch_solve: Union[int, None] = None, json_file: str = 'acados_ocp.json',  build: bool = True, generate: bool = True, verbose: bool=True):
 
         if not isinstance(N_batch, int) or N_batch <= 0:
             raise Exception("AcadosOcpBatchSolver: argument N_batch should be a positive integer.")
+        if num_threads_in_batch_solve is None:
+            num_threads_in_batch_solve = ocp.solver_options.num_threads_in_batch_solve
+            print(f"Warning: num_threads_in_batch_solve is None. Using value {num_threads_in_batch_solve} set in ocp.solver_options instead.")
+            print("In the future, it should be passed explicitly in the AcadosOcpBatchSolver constructor.")
```
FreyJo, line 62: It would be good to also put the deprecation warning in the setter of `num_threads_in_batch_solve`.
Also, since `num_threads_in_batch_solve` is not used in the template logic anymore, and as it was never functional in MATLAB, we can remove the option in all .m files.

interfaces/acados_template/acados_template/c_templates_tera/acados_sim_solver.in.c
```diff
@@ -417,25 +417,27 @@ int {{ model.name }}_acados_sim_solve({{ model.name }}_sim_solver_capsule *capsu
...
-void {{ model.name }}_acados_sim_batch_solve({{ model.name }}_sim_solver_capsule ** capsules, int N_batch)
-{
-{% if solver_options.num_threads_in_batch_solve > 1 %}
-    int num_threads_bkp = omp_get_num_threads();
-    omp_set_num_threads({{ solver_options.num_threads_in_batch_solve }});
-
-    #pragma omp parallel for
-{%- endif %}
-    for (int i = 0; i < N_batch; i++)
+{% if solver_options.with_batch_functionality %}
```
FreyJo, line 420: In this templated logic, we don't want to indent, such that the rendered code is aligned nicely.

FreyJo (commented):
Good job!
I have a few comments which should be easy to address.

FreyJo (approved):

PR #1484, opened 2025-03-28
Python: Model with properties

sandmaennchen:
Finally `AcadosModel` with properties. 

Next step: move type checks from `make_consistent` to setters


FreyJo (approved):
Looks good!
If it was not for backwards compatibility, I would be in favor of having cost and constraints related expressions to `AcadosOcpConstraints`, `AcadosOcpCost` at some point.

PR #1492, opened 2025-04-08
Globalization: fix iteration number in merit backtracking

FreyJo:


david0oo (changes requested):
Looks good. Are there other locations in the acados code base where it makes sense to change `sqp_iter` to `nlp_mem->iter`?

PR #1494, opened 2025-04-09
C: move `qp_res` memory management to common fully

FreyJo:
It was already in the common workspace.


david0oo (approved):
Looks good! I like that our code gets DRYer and cleaner.

PR #1495, opened 2025-04-09
Step norm logging

FreyJo:
-  move primal_step_norm to common memory, as option is also there
-  Interfaces: fix log_primal_step_norm for SQP_WFQP and multi-phase 
- add option ` log_dual_step_norm`
- test getting step norms in MATLAB and Python

Bit unrelated:
-  SQP_WFQP: remove max_iter duplicated from common

### DIFF
--- a/acados/ocp_nlp/ocp_nlp_common.c
+++ b/acados/ocp_nlp/ocp_nlp_common.c
@@ -1631,6 +1631,11 @@ acados_size_t ocp_nlp_memory_calculate_size(ocp_nlp_config *config, ocp_nlp_dims
     size += sizeof(struct ocp_nlp_timings);
 
     size += (N+1)*sizeof(bool); // set_sim_guess
+    // primal step norm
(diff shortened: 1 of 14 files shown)


acados/ocp_nlp/ocp_nlp_common.c
```diff
@@ -3374,6 +3392,32 @@ void ocp_nlp_res_get_inf_norm(ocp_nlp_res *res, double *out)
 }
 
 
+double ocp_nlp_compute_delta_dual_norm(ocp_nlp_dims *dims, ocp_nlp_workspace *work, ocp_nlp_out *nlp_out, ocp_qp_out *qp_out)
```
david0oo, line 3395: Should we call it `dual_norm_inf`? Then, it is clear from the function signature?

david0oo (changes requested):
Only change the signature. The rest looks good.

PR #1496, opened 2025-04-10
C: avoid using local variable `sqp_iter`

FreyJo:
Follow-up of https://github.com/acados/acados/pull/1492.
The changed tests in https://github.com/acados/acados/pull/1492 have been reverted as the iteration reference was still wrong there. Now I verified by printing.

### DIFF
--- a/acados/ocp_nlp/ocp_nlp_sqp_with_feasible_qp.c
+++ b/acados/ocp_nlp/ocp_nlp_sqp_with_feasible_qp.c
@@ -959,7 +959,7 @@ Solves the QP. Either solves feasibility QP or nominal QP
 */
 static int prepare_and_solve_QP(ocp_nlp_config* config, ocp_nlp_sqp_wfqp_opts* opts, ocp_qp_in* qp_in, ocp_qp_out* qp_out,
                     ocp_nlp_dims *dims, ocp_nlp_sqp_wfqp_memory* mem, ocp_nlp_in* nlp_in, ocp_nlp_out* nlp_out,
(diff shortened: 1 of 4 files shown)


acados/ocp_nlp/ocp_nlp_sqp_with_feasible_qp.c
```diff
@@ -990,7 +990,7 @@ static int prepare_and_solve_QP(ocp_nlp_config* config, ocp_nlp_sqp_wfqp_opts* o
         if (solve_feasibility_qp && opts->use_constraint_hessian_in_feas_qp)
         {
             // LM for feasibility QP
-            ocp_nlp_add_levenberg_marquardt_term(config, dims, nlp_in, nlp_out, opts->nlp_opts, nlp_mem, nlp_work, mem->alpha, sqp_iter, qp_in);
+            ocp_nlp_add_levenberg_marquardt_term(config, dims, nlp_in, nlp_out, opts->nlp_opts, nlp_mem, nlp_work, mem->alpha, nlp_mem->iter, qp_in);
```
david0oo, line 993: Since nlp_mem is passed. Do we need nlp_mem->iter as additional argument?

david0oo (changes requested):
Looks good. If comments are addressed, and tests pass, can be merged

acados/ocp_nlp/ocp_nlp_sqp_with_feasible_qp.c
FreyJo: In `SQP_RTI`, this function is called with iter=0. I don't know the reason, but I don't want to change the behavior.

PR #1498, opened 2025-04-11
Allow `solver_options.N_horizon=0`

Confectio:
This PR modifies the Matlab and Python interfaces to allow formulating optimization problems without any shooting-structure and `solver_options.N_horizon=0` .
If `solver_options.N_horizon=0`, only the terminal costs and constraints will be regarded.
In particular, also no dynamics, other costs, and constraints have to be defined.

We have also updated some examples to use this feature instead of the workarounds.


Confectio:
Please someone look at the maratos example, it fails :)

FreyJo (approved):
Very good!
It is not clear what is going on in the maratos example after looking into it a bit.
I refactored it to easily switch between the old and new formulations and to log iterates without for looping.
The difference in behavior should be investigated a bit more closely later.

PR #1501, opened 2025-04-17
Python verbosity

FreyJo:
- add verbose to `make_consistent()` for `AcadosOcp` and `AcadosMultiphaseOcp`
- for batch solvers: only use verbose when creating the first solver

### DIFF
--- a/interfaces/acados_template/acados_template/acados_ocp_solver.py
+++ b/interfaces/acados_template/acados_template/acados_ocp_solver.py
@@ -90,7 +90,7 @@ class AcadosOcpSolver:
         return self.__shared_lib
 
     @classmethod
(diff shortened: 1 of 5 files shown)


interfaces/acados_template/acados_template/acados_ocp_solver.py
```diff
@@ -90,7 +90,7 @@ def shared_lib(self,):
         return self.__shared_lib
 
     @classmethod
-    def generate(cls, acados_ocp: Union[AcadosOcp, AcadosMultiphaseOcp], json_file: str, simulink_opts=None, cmake_builder: CMakeBuilder = None):
+    def generate(cls, acados_ocp: Union[AcadosOcp, AcadosMultiphaseOcp], json_file: str, simulink_opts=None, cmake_builder: CMakeBuilder = None, verbose=True):
```
Confectio, line 93: An entry in the docstring is missing

Confectio (approved):

PR #1506, opened 2025-04-24
Python/MATLAB: Check CasADi expressions of terminal stage

sandmaennchen:
Check whether CasADi expressions of terminal stage depend on `u` or `z` 
+ minor cleanup in `make_consistent` and code generation

### DIFF
--- a/interfaces/acados_template/acados_template/casadi_function_generation.py
+++ b/interfaces/acados_template/acados_template/casadi_function_generation.py
@@ -458,11 +458,6 @@ def generate_c_code_external_cost(context: GenerateContext, model: AcadosModel, 
         suffix_name_value_sens = "_cost_ext_cost_e_grad_p"
         ext_cost = model.cost_expr_ext_cost_e
         custom_hess = model.cost_expr_ext_cost_custom_hess_e
-        # Last stage cannot depend on u and z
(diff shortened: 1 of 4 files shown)


interfaces/acados_template/acados_template/casadi_function_generation.py
```diff
@@ -458,14 +458,10 @@ def generate_c_code_external_cost(context: GenerateContext, model: AcadosModel,
...
-        # Last stage cannot depend on u and z
-        if any(ca.which_depends(ext_cost, model.u)):
-            raise ValueError("terminal cost cannot depend on u.")
-        if any(ca.which_depends(ext_cost, model.z)):
-            raise ValueError("terminal cost cannot depend on z.")
+
         # create dummy u, z
-        u = symbol("u", 0, 0)
-        z = symbol("z", 0, 0)
+        u = symbol("u", 0, 1)
```
FreyJo, line 463: why did you change this?

FreyJo (approved):

interfaces/acados_template/acados_template/casadi_function_generation.py
sandmaennchen: to be consistent with the empty values set in `make_consistent`. I think it does not matter which one is used, but I got confused at first because they were not the same

PR #1510, opened 2025-04-28
add our application on tiltable quadrotors to "real_world_examples"

Li-Jinjie:
Hi Jonathan, 

This PR provides an example of using NMPC for overactuated tiltable quadrotors. We have some more impressive demos, and we’ll update them after we submit the paper.

Best,
Jinjie


FreyJo (approved):
Awesome, thanks you! :blush:

PR #1512, opened 2025-04-28
Updates with respect to new HPIPM sensitivity API

FreyJo:

PR #1515, opened 2025-04-29
Test forward vs. adjoint solution sensitivities on CI with quick settings

FreyJo:


joao-krauchenco-mottu (approved):

examples/acados_python/chain_mass/solution_sensitivity_example.py
```diff
@@ -259,6 +259,7 @@ def export_parametric_ocp(
     nlp_iter: int = 50,
     nlp_tol: float = 1e-5,
     random_scale: dict = {"m": 0.0, "D": 0.0, "L": 0.0, "C": 0.0},
+    ext_fun_compile_flags: Optional[str] = None,
```
sandmaennchen, line 262: why do you need this?

sandmaennchen (approved):

examples/acados_python/chain_mass/solution_sensitivity_example.py
FreyJo: The code generation and compilation is what makes the example slow. That's why I switched to Euler and turned off "-O2" which is the default for `ext_fun_compile_flags`, in order to run the test relatively quickly on actions.
It was previously not included because it is not so fast. However, now it uncovers some issues in https://github.com/acados/acados/pull/1512

PR #1520, opened 2025-05-05
Python: fix `u_labels` init

josipkh:
When not set, `u_labels` were wrongly initialized with the dimension of `x`.


sandmaennchen (approved):
Thanks!

Issue #1525, opened 2025-05-14
`AcadosCasadiOcpSolver` extensions and todos


FreyJo:
This list is roughly sorted by the priority of the tasks, and might be rearranged later.

- [ ] reorder internal variables of CasADi NLP to have $x_0, u_0, x_1, ... $, this is needed also for comparisons with fatrop, as mentioned here: https://github.com/acados/acados/pull/1523#discussion_r2084300602

## Implement support for and test equivalence of:
- [ ] slack variables
- [ ] Lagrange multipliers, note: acados formulation has them split into ones of upper bound lower bound, while CasADi has the sum of them
- [ ] implement initializations: store iterate / load iterate, to enable initializing acados solver with CasADi IPOPT solution

- [ ] implement updates:
    - [ ] stage-wise parameters
    - [ ] global parameter
    - [ ] bounds

## Additional features
- [ ] implement IRK via: casados integrators

PR #1527, opened 2025-05-19
Reorder variables in `AcadosCasadiOcpSolver`

Pandatheon:
- organized the variables along time nodes as $x_0, x_1,....,x_{N+1}$ and $u_0, u_1...., u_N$ and then vectorized them in the interleaved order as $x_0, u_0, x_1, u_1...., x_N, u_N,x_{N+1}$.
- I also adjust the `get()` function to get solution values in an interleaved manner.

### DIFF
--- a/interfaces/acados_template/acados_template/acados_casadi_ocp_solver.py
+++ b/interfaces/acados_template/acados_template/acados_casadi_ocp_solver.py
@@ -36,7 +36,6 @@ from typing import Union, Tuple
 import numpy as np
 from .acados_ocp import AcadosOcp
 from .acados_ocp_iterate import AcadosOcpIterate, AcadosOcpIterates, AcadosOcpFlattenedIterate
-from itertools import zip_longest
 
 class AcadosCasadiOcpSolver:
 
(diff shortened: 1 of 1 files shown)


interfaces/acados_template/acados_template/acados_casadi_ocp_solver.py
```diff
@@ -59,32 +59,36 @@ def create_casadi_nlp_formulation(cls, ocp: AcadosOcp) -> Tuple[dict, dict]:
         if any([dims.ns_0, dims.ns, dims.ns_e]):
             raise NotImplementedError("CasADi NLP formulation not implemented for formulations with soft constraints yet.")
 
-        # create variables
+        # create variables indexed by shooting nodes, vectorize in the end
         ca_symbol = model.get_casadi_symbol()
         xtraj = ca_symbol('x', dims.nx, solver_options.N_horizon+1)
+        xtraj_node = [xtraj[:, i] for i in range(xtraj.shape[-1])]
```
sandmaennchen, line 65: maybe use `ca.horzsplit`

interfaces/acados_template/acados_template/acados_casadi_ocp_solver.py
```diff
@@ -103,22 +107,22 @@ def create_casadi_nlp_formulation(cls, ocp: AcadosOcp) -> Tuple[dict, dict]:
         for i in range(solver_options.N_horizon):
             # add dynamics constraints
             if solver_options.integrator_type == "DISCRETE":
-                g.append(xtraj[:, i+1] - f_discr_fun(xtraj[:, i], utraj[:, i], ptraj[:, i], model.p_global))
+                g.apppend(xtraj_node[i+1] - f_discr_fun(xtraj_node[i], utraj_node[i], ptraj_node[i], model.p_global))
             elif solver_options.integrator_type == "ERK":
-                g.append(xtraj[:, i+1] - f_discr_fun(xtraj[:, i], utraj[:, i], solver_options.time_steps[i]))
+                g.append(xtraj_node[i+1] - f_discr_fun(xtraj_node[i], utraj_node[i], solver_options.time_steps[i]))
```
sandmaennchen, line 112: not related to this PR, but the ERK integrator might also depend on `p` and `p_global` I guess?

interfaces/acados_template/acados_template/acados_casadi_ocp_solver.py
```diff
@@ -244,14 +252,14 @@ def get(self, stage: int, field: str):
         if self.nlp_sol is None:
             raise ValueError('No solution available. Please call solve() first.')
         dims = self.acados_ocp.dims
+        pivot = stage*(dims.nx+dims.nu)
 
         if field == 'x':
             sol_w = self.nlp_sol['x']
```
sandmaennchen, line 258: I would unpack the solution and call `full()` once after solving the problem

sandmaennchen (approved):
Nice! :+1:  

I have some minor comments, but mostly unrelated to the reordering :wink:

interfaces/acados_template/acados_template/acados_casadi_ocp_solver.py
FreyJo: yes, tracked in https://github.com/acados/acados/issues/1525

interfaces/acados_template/acados_template/acados_casadi_ocp_solver.py
FreyJo: I actually think not creating a matrix at all is preferred, like just 1 line instead of 2:
`xtraj_node = [ca_symbol(f'x{i}', dims.nx, 1) for i in range(N_horizon+1)]`

Because
a) extensibility for varying dimension (multi-phase and slacks)
b) As the x values are only used individually, I would bet the resulting solver to be more efficient if the x values are individual symbols. (Not verified though).

Pandatheon:
- revise the creating of symbols, avoiding matrix
- unpack solution with `full()` after solving the problem
- write explicit `for` loop for interleaving variables

PR #1528, opened 2025-05-20
Fix `p_global` precomputation in case sparse data is detected

FreyJo:
By making it dense in the MATLAB & Python interfaces.


sandmaennchen (approved):
:+1:

PR #1540, opened 2025-06-03
Fix compatibility of full condensing and SQP-WFQP

FreyJo:
Make full condensing behave like partial one in terms of when dims need to be provided. Added corresponding test.
Additionally: Implemented `AcadosOcpFlattenedIterate.allclose()` for testing.


david0oo (approved):
Looks good!

PR #1541, opened 2025-06-05
[docs] Fix mistake in doc string of acados sim object for num of stages

johanneskbl:
The number of stages is per default 4 and not 1 as stated in the doc string.
Please look at the screenshot ([or code](https://github.com/acados/acados/blob/main/interfaces/acados_template/acados_template/acados_sim.py#L82)) to confirm.

To me this was a bit surprising since I tried to play around with this property.
Screenshot from 2025-06-05 14-50-26


sandmaennchen:
Good catch, thanks!

sandmaennchen (approved):

PR #1543, opened 2025-06-05
Extend `AcadosCasadiOcpSolver` with options and support fatrop

FreyJo:


interfaces/acados_template/acados_template/acados_casadi_ocp_solver.py
```diff
@@ -253,14 +262,28 @@ def create_casadi_nlp_formulation(cls, ocp: AcadosOcp) -> Tuple[dict, dict, np.n
         return nlp, bounds, w0, index_map
 
 
-    def __init__(self, acados_ocp: AcadosOcp, solver: str = "ipopt", verbose=True):
+    def __init__(self, acados_ocp: AcadosOcp, solver: str = "ipopt", verbose=True,
+                 casadi_nlp_opts: dict = None):
```
sandmaennchen: I think the name is a bit weird, I would rather call it `casadi_solver_opts`. And the type is `Optional[dict]` 🙃

PR #1548, opened 2025-06-12
OCP sensitivity check

JasperHoffmann:
This PR moves the sensitivity check to the `AcadosOcp` class:
1. `AcadosOcp` gets a `ensure_solution_sensitivities_available` method.
2. `AcadosOcpOptions` gets a private method for checking whether the solver options are correctly set.
3. `AcadosModel` gets a private method to check whether the Hessian is custom.

For `AcadosMultiPhaseOcp` the function `_ensure_solution_sensitivities_available` raises an error, as the sensitivity functionality is not interfaced yet.


FreyJo (approved):
Looks good, thanks!

PR #1550, opened 2025-06-18
MATLAB: implement `AcadosSimSolver.simulate()`

josipkh:
Analogous to the [Python version](https://github.com/acados/acados/blob/8ff3477761d5272b0e5735b97cf3fd821a1bd0b3/interfaces/acados_template/acados_template/acados_sim_solver.py#L241).
I added a comment in the `getting_started` examples, for new users.

### DIFF
--- a/examples/acados_matlab_octave/getting_started/minimal_example_closed_loop.m
+++ b/examples/acados_matlab_octave/getting_started/minimal_example_closed_loop.m
@@ -185,23 +185,9 @@ for i=1:N_sim
     u0 = ocp_solver.get('u', 0);
     status = ocp_solver.get('status'); % 0 - success
 
-    % set initial state for the simulation
-    sim_solver.set('x', x0);
-    sim_solver.set('u', u0);
-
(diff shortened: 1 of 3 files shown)


examples/acados_matlab_octave/getting_started/minimal_example_closed_loop.m
```diff
@@ -198,6 +198,10 @@
     % get simulated state
     x_sim(:,i+1) = sim_solver.get('xn');
     u_sim(:,i) = u0;
+
+    % one can also use:
+    % x_sim(:,i+1) = sim_solver.simulate(x0, u0);
+    % u_sim(:,i) = u0;
```
FreyJo: Could you replace that here instead of adding the commented version, such that the new function is tested?
The `sim_status` check can also be removed here. It is not really useful.

josipkh: done :)

PR #1551, opened 2025-06-18
`AcadosCasadiOcpSolver`: Fix partial bounds for `set()` and `get()`

Pandatheon:


interfaces/acados_template/acados_template/acados_casadi_ocp_solver.py
```diff
@@ -96,18 +100,28 @@ def __init__(self, ocp: AcadosOcp, with_hessian=False):
         lb_xtraj_node[0][constraints.idxbx_0] = constraints.lbx_0
         ub_xtraj_node[0][constraints.idxbx_0] = constraints.ubx_0
         offset = 0
+        index_map['lam_x_in_lam_w'].append(list(offset + constraints.idxbx_0))
+        offset += dims.nbx_0 +dims.nbu
         for i in range(1, N_horizon):
             lb_xtraj_node[i][constraints.idxbx] = constraints.lbx
             ub_xtraj_node[i][constraints.idxbx] = constraints.ubx
+            index_map['lam_x_in_lam_w'].append(list(offset + constraints.idxbx))
+            offset += dims.nx + dims.nbu
```
FreyJo: Here, it should probably be `dims.nu` instead of `dims.nbu`.
In order to test this, you could add an additional control variable in the test problem, which has just a quadratic cost, but no bounds and doesn't influence the state at all.

I think it is better to merge the loops where the bounds are set up, and add some `if`s inside to handle the first and last stage, such that the `offset` variable is more consistent / understandable.
This should be helpful especially, when we want to add slacks later on.
But maybe you can just do this merge when you add the slacks.

interfaces/acados_template/acados_template/acados_casadi_ocp_solver.py
```diff
@@ -55,6 +55,10 @@ def __init__(self, ocp: AcadosOcp, with_hessian=False):
             # indices of variables within w
             'x_in_w': [],
             'u_in_w': [],
+            # indices of state bounds within lam_x(lam_w) in casadi formulation
+            'lam_x_in_lam_w':[],
```
FreyJo: I suggest renaming to `lam_bx_in_lam_w`, same for u.

examples/acados_python/casadi_tests/test_casadi_get_set.py
```diff
@@ -0,0 +1,123 @@
...
+    return simX, simU
+
+def formulate_ocp(Tf: float = 1.0, N: int = 20)-> AcadosOcp:
+    # create ocp object to formulate the OCP
+    ocp = AcadosOcp()
+
+    # set model
+    model = export_pendulum_ode_model()
+    ocp.model = model
+
```
FreyJo, line 31: Does this test work?
Please add it in `interfaces/CMakeLists.txt` such that it is tested on Github actions.

interfaces/acados_template/acados_template/acados_casadi_ocp_solver.py
```diff
@@ -90,24 +94,36 @@ def __init__(self, ocp: AcadosOcp, with_hessian=False):
...
+        for i in range(0, N_horizon+1):
+            if i == 0:
+                lb_xtraj_node[0][constraints.idxbx_0] = constraints.lbx_0
+                ub_xtraj_node[0][constraints.idxbx_0] = constraints.ubx_0
+                index_map['lam_bx_in_lam_w'].append(list(offset + constraints.idxbx_0))
+                offset += dims.nx
+                lb_utraj_node[i][constraints.idxbu] = constraints.lbu
+                ub_utraj_node[i][constraints.idxbu] = constraints.ubu
+                index_map['lam_bu_in_lam_w'].append(list(offset + constraints.idxbu))
+                offset += dims.nu
```
FreyJo: I suggest to move the u part in a second `if i < N_horizon` to avoid duplicating it.

examples/acados_python/casadi_tests/test_casadi_get_set.py
```diff
@@ -0,0 +1,123 @@
...
+    simX[N_horizon,:] = ocp_solver.get(N_horizon, "x")
+
+    return simX, simU
+
+def formulate_ocp(Tf: float = 1.0, N: int = 20)-> AcadosOcp:
+    # create ocp object to formulate the OCP
+    ocp = AcadosOcp()
+
+    # set model
+    model = export_pendulum_ode_model()
```
FreyJo, line 27: Please test with `idxbu != range(nu)`. Should be as easy as adding
```
# add dummy control
model.u = ca.vertcat(model.u, ca.SX.sym('dummy_u'))
```
here.
And adjusting `R_mat` below.

FreyJo (approved):

PR #1562, opened 2025-06-25
Developer checks

FreyJo:


FreyJo:
Can't formally approve. But looks good, thanks @david0oo :100:

PR #1565, opened 2025-06-26
C: Avoid initializing Hessian explicitly

FreyJo:
- Addressed some TODOs in `SQP_WFQP`
- cost integration: Now cost scaling happens in dynamics module if cost is integrated, only slack contributions to cost, are always scaled in cost module.
    - fix if `cost_scaling != time_step`.
   - Implement separate pointers for Hessian of cost and dynamics correctly


david0oo (approved):
Looks good! And performance on NLPs/OCPs did not change.

PR #1567, opened 2025-06-27
Removed duplicate row in `AcadosCasadiOcp`

david0oo:
Remove of duplicate redundant row. Hoping that the `AcadosCasadiOcp` supports general constraints soon.


FreyJo:
Check is also remove in https://github.com/acados/acados/pull/1571 where linear constraints are added.

PR #1585, opened 2025-07-09
`AcadosCasadiOCP`: Add support for soft nonlinear constraints

Pandatheon:
- Add support for soft nonlinear constraint in `AcadosCasadiOCP`
- Preliminarily complete `get()` and `set()`for slack variables in `AcadosCasadiOcpSolver`
- Add a draft test for testing soft nonlinear constraint
- Initially write some helper functions for simplification

### DIFF
--- a/interfaces/acados_template/acados_template/acados_casadi_ocp_solver.py
+++ b/interfaces/acados_template/acados_template/acados_casadi_ocp_solver.py
@@ -82,8 +82,10 @@ class AcadosCasadiOcp:
         N_horizon = solver_options.N_horizon
 
         # check what is not supported yet
-        if any([dims.nsbx,dims.nsbx_e, dims.nsbu]):
-            raise NotImplementedError("AcadosCasadiOcpSolver does not support slack variables (s) for variables (w) yet.")
+        if any([dims.nsbx, dims.nsbx_e, dims.nsbu]):
+            raise NotImplementedError("AcadosCasadiOcpSolver does not support slack variables (s) for variables (x and u) yet.")
+        if any([dims.nsg, dims.nsg_e, dims.nsphi, dims.nsphi_e]):
+            raise NotImplementedError("AcadosCasadiOcpSolver does not support slack variables (s) for linear and compound constraints (h, phi) yet.")
         if dims.nz > 0:
             raise NotImplementedError("AcadosCasadiOcpSolver does not support algebraic variables (z) yet.")
         if ocp.solver_options.integrator_type not in ["DISCRETE", "ERK"]:
--- a/interfaces/acados_template/acados_template/acados_casadi_ocp_solver.py
+++ b/interfaces/acados_template/acados_template/acados_casadi_ocp_solver.py
@@ -329,18 +329,21 @@ class AcadosCasadiOcp:
             # initial stage
             if i == 0 and N_horizon > 0:
                 lh, uh = constraints.lh_0, constraints.uh_0
+                lg, ug = constraints.lg, constraints.ug
                 ng, nh, nphi = dims.ng, dims.nh_0, dims.nphi_0
                 nsg, nsh, nphi, idxsh = dims.nsg, dims.nsh_0, dims.nphi_0, constraints.idxsh_0
                 h_i_nlp_expr = h_0_fun(xtraj_node[i], utraj_node[i], ptraj_node[i], model.p_global)
             # intermediate stages
             elif i < N_horizon:
                 lh, uh = constraints.lh, constraints.uh
+                lg, ug = constraints.lg, constraints.ug
                 ng, nh, nphi = dims.ng, dims.nh, dims.nphi
                 nsg, nsh, nphi, idxsh = dims.nsg, dims.nsh, dims.nphi, constraints.idxsh
                 h_i_nlp_expr = h_fun(xtraj_node[i], utraj_node[i], ptraj_node[i], model.p_global)
             # terminal stage
             else:
                 lh, uh = constraints.lh_e, constraints.uh_e
+                lg, ug = constraints.lg_e, constraints.ug_e
                 ng, nh, nphi = dims.ng_e, dims.nh_e, dims.nphi_e
                 nsg, nsh, nphi, idxsh = dims.nsg_e, dims.nsh_e, dims.nphi_e, constraints.idxsh_e
                 h_i_nlp_expr = h_e_fun(xtraj_node[i], ptraj_node[i], model.p_global)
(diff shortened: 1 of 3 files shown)


interfaces/acados_template/acados_template/acados_casadi_ocp_solver.py
```diff
@@ -234,20 +328,53 @@ def __init__(self, ocp: AcadosOcp, with_hessian=False):
...
+
+                        h_0_nlp_expr = h_0_fun(xtraj_node[i], utraj_node[i], ptraj_node[i], model.p_global)
+
+                        
+                        for j in range(len(constraints.lh_0)):
+                            if j in hard_h_indices:
+                                g.append(h_0_nlp_expr[hard_h_indices])
+                                lbg.append(constraints.lh_0[hard_h_indices])
+                                ubg.append(constraints.uh_0[hard_h_indices])
+                                index_map['lam_gnl_in_lam_g'][i].append(*list(range(offset, offset + len(hard_h_indices))))
```
FreyJo: Here, it should be `[j]` instead of `hard_h_indices`.
The same mistake is in more places.
The code works for the current example only because there is `len(hard_h_indices) <= 1`.
You can maybe try to add a test variant, where the u bounds are also formulated as `h` to have more `h` constraints.

I think it would be could to avoid duplicating the logic for initial, path, and terminal constraints.
We should try to add some helper functions to the class to make everything more readable.

interfaces/acados_template/acados_template/acados_casadi_ocp_solver.py
```diff
@@ -572,14 +784,35 @@ def get(self, stage: int, field: str):
...
-            lam = np.concatenate((lbu_lam, lbx_lam, lg_lam, ubu_lam, ubx_lam, ug_lam))
+            if any([dims.ns_0, dims.ns, dims.ns_e]):
+                lw_soft_lam = self.nlp_sol_lam_x[self.index_map['sl_in_w'][stage]]
+                uw_soft_lam = self.nlp_sol_lam_x[self.index_map['su_in_w'][stage]]
+                lg_soft_lam = self.nlp_sol_lam_g[self.index_map['lam_sl_in_lam_g'][stage]]
+                ug_soft_lam = self.nlp_sol_lam_g[self.index_map['lam_su_in_lam_g'][stage]]
+                # TODO: could be improved?
+                if self.index_map['lam_gnl_in_lam_g'][stage]:
+                    if self.index_map['lam_gnl_in_lam_g'][stage] < self.index_map['lam_su_in_lam_g'][stage]: 
+                        # hard constraints before soft constraints
```
FreyJo: Please add an exception to handle any implcit requirements.

Also add a TODO in the comment.

It should be something like this:
```
lbg_lam = np.zeros(..)
lbg[offset+idxsh] = lg_soft_lam
```
I think it also makes sense to have the not slacked constraints index vectors, like idxnsh available.

interfaces/acados_template/acados_template/acados_casadi_ocp_solver.py
Pandatheon: I just deleted some repetitive code first, which might be an ad hoc method.

interfaces/acados_template/acados_template/acados_casadi_ocp_solver.py
```diff
@@ -55,40 +55,55 @@ def __init__(self, ocp: AcadosOcp, with_hessian=False):
...
         solver_options = ocp.solver_options
         N_horizon = solver_options.N_horizon
 
         # check what is not supported yet
-        if any([dims.ns_0, dims.ns, dims.ns_e]):
-            raise NotImplementedError("AcadosCasadiOcpSolver does not support soft constraints yet.")
+        if any([dims.nsbx, dims.nsbx_e, dims.nsbu]):
+            raise NotImplementedError("AcadosCasadiOcpSolver does not support slack variables (s) for variables (x and u) yet.")
+        if any([dims.nsg, dims.nsg_e, dims.nsphi, dims.nsphi_e]):
+            raise NotImplementedError("AcadosCasadiOcpSolver does not support slack variables (s) for linear and compound constraints (h, phi) yet.")
```
FreyJo: Please change to: " for general linear and convex-over-nonlinear constraints (g, phi) yet."

interfaces/acados_template/acados_template/acados_casadi_ocp_solver.py
```diff
@@ -55,40 +55,55 @@ def __init__(self, ocp: AcadosOcp, with_hessian=False):
...
-        utraj_node = [ca_symbol(f'u{i}', dims.nu, 1) for i in range(N_horizon)]
-        if dims.nz > 0:
-            raise NotImplementedError("CasADi NLP formulation not implemented for models with algebraic variables (z).")
+        xtraj_node = []
+        utraj_node = []
+        sl_node = []
+        su_node = []
+        for i in range(N_horizon+1):
+            if i == 0:
+                self._append_node(ca_symbol, xtraj_node, utraj_node, sl_node, su_node, i, dims, )
```
FreyJo: remove trailing `, `.

FreyJo: remove the ifs here, they are now handled in the function.

interfaces/acados_template/acados_template/acados_casadi_ocp_solver.py
```diff
@@ -131,40 +160,61 @@ def __init__(self, ocp: AcadosOcp, with_hessian=False):
...
-        offset += dims.nx
-        # add parameters
-        p_list.append(ocp.parameter_values)
-        index_map['p_in_p_nlp'].append(list(range(offset_p, offset_p+dims.np)))
-        offset_p += dims.np
-        p_list.append(ocp.p_global_values)
-        index_map['p_global_in_p_nlp'].append(list(range(offset_p, offset_p+dims.np_global)))
-        offset_p += dims.np_global
+        for i in range(N_horizon+1):
+            # TODO: internally: offset_w += len(sym)
```
FreyJo: Here is still a TODO.
The idea was to make `offset_w`, `index_map`  variables of the class and add to the index_map within the helper functions.

FreyJo: Also, instead of `_add_slack_variables`, `_add_primal_variable_x`, `_add_primal_variable_u`, we could have just one function called `_add_primal_variable` and call it for x, u, sl, su.

interfaces/acados_template/acados_template/acados_casadi_ocp_solver.py
```diff
@@ -98,6 +113,19 @@ def __init__(self, ocp: AcadosOcp, with_hessian=False):
...
         ub_utraj_node = [np.inf * ca.DM.ones((dims.nu, 1)) for _ in range(N_horizon)]
+        # setup slack variables
+        lb_slack_node = []
+        ub_slack_node = []
+        for i in range(N_horizon+1):
+            if i == 0 and dims.ns_0:
+                lb_slack_node.append(0 * ca.DM.ones((dims.ns_0, 1)))
+                ub_slack_node.append(np.inf * ca.DM.ones((dims.ns_0, 1)))
+            elif i < N_horizon and dims.ns:
+                lb_slack_node.append(0 * ca.DM.ones((dims.ns, 1)))
```
FreyJo: Instead of the zeros it should be:
It should be `[lsbu, lsbx, lsg, lsh, lsphi]`.
Off course those values are zero in the vast majority of cases though.

FreyJo (commented):
I added some minor comments, but it already looks pretty good!

PR #1599, opened 2025-07-21
Simulink: automatically create the S-function block with port info

josipkh:
This PR adds the option to automatically generate the Simulink block with port information when `make_sfun.m` is called.
Also, I added a missing Simulink input to the docs, please correct me if I'm wrong.

### DIFF
--- a/interfaces/acados_template/acados_template/c_templates_tera/matlab_templates/make_sfun_sim.in.m
+++ b/interfaces/acados_template/acados_template/c_templates_tera/matlab_templates/make_sfun_sim.in.m
@@ -112,13 +112,13 @@ fprintf(output_note)
 
 
 % create the Simulink block for the integrator
-modelName = '{{ name }}_sim_solver_simulink_block';
+modelName = '{{ model.name }}_sim_solver_simulink_block';
 new_system(modelName);
 open_system(modelName);
 
-blockPath = [modelName '/{{ name }}_sim_solver'];
+blockPath = [modelName '/{{ model.name }}_sim_solver'];
 add_block('simulink/User-Defined Functions/S-Function', blockPath);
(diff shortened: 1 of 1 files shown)


interfaces/acados_template/acados_template/c_templates_tera/matlab_templates/make_sfun.in.m
```diff
@@ -569,6 +569,38 @@
 
 fprintf(output_note)
 
+{%- if simulink_opts.generate_simulink_block == 1 %}
+modelName = 'acados_ocp_simulink_block';
+new_system(modelName);
+open_system(modelName);
+
+blockPath = [modelName '/acados OCP solver'];
```
FreyJo: I think it would be nicer if `modelName` and `blockPath` contain the model / solver name.
Or do you have strong a strong reason against that?

josipkh: I agree, just didn't have a good idea which wasn't a bit too long for displaying...
do you have any suggestions?

josipkh:
Also, would you like to have the same functionality in `make_sfun_sim.m`?

interfaces/acados_template/acados_template/c_templates_tera/matlab_templates/make_sfun.in.m
FreyJo: I think `{{ name }}_ocp_solver` would be good. I think you can always increase the size of the block in the GUI to have it displayed.

FreyJo:
Yes, it would also be great to have it for the integrator! Thanks!

josipkh:
Thank you for the quick review, I fixed a small bug so it can be merged now :)

PR #1605, opened 2025-07-29
`AcadosCasadiOcp`: add support for single shooting formulation

Pandatheon:
- Modify `AcadosCasadiOcp` for Single shooting support
- Add a draft test

### DIFF
--- a/interfaces/acados_template/acados_template/acados_casadi_ocp_solver.py
+++ b/interfaces/acados_template/acados_template/acados_casadi_ocp_solver.py
@@ -41,7 +41,7 @@ from .acados_ocp_iterate import AcadosOcpIterate, AcadosOcpFlattenedIterate
 
 class AcadosCasadiOcp:
 
-    def __init__(self, ocp: AcadosOcp, with_hessian=False):
+    def __init__(self, ocp: AcadosOcp, with_hessian=False, single_shooting=False):
         """
         Creates an equivalent CasADi NLP formulation of the OCP.
         Experimental, not fully implemented yet.
@@ -100,8 +100,18 @@ class AcadosCasadiOcp:
         utraj_node = []
         sl_node = []
         su_node = []
-        for i in range(N_horizon+1):
-            self._append_node(ca_symbol, xtraj_node, utraj_node, sl_node, su_node, i, dims)
+        if not single_shooting:
+            for i in range(N_horizon+1):
+                self._append_node('x', ca_symbol, xtraj_node, i, dims)
+                self._append_node('u', ca_symbol, utraj_node, i, dims)
+                self._append_node('sl', ca_symbol, sl_node, i, dims)
+                self._append_node('su', ca_symbol, su_node, i, dims)
+        else:
+            self._append_node('x', ca_symbol, xtraj_node, 0, dims)
+            for i in range(N_horizon):
+                self._append_node('u', ca_symbol, utraj_node, i, dims)
+                self._append_node('sl', ca_symbol, sl_node, i, dims)
+                self._append_node('su', ca_symbol, su_node, i, dims)
 
         # parameters
         ptraj_node = [ca_symbol(f'p{i}', dims.np, 1) for i in range(N_horizon+1)]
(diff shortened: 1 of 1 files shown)


interfaces/acados_template/acados_template/acados_casadi_ocp_solver.py
```diff
@@ -119,8 +129,13 @@ def __init__(self, ocp: AcadosOcp, with_hessian=False):
         ub_slack_node = ([np.inf * ca.DM.ones((dims.ns, 1))] if dims.ns_0 else []) + \
                         ([np.inf * ca.DM.ones((dims.ns, 1)) for _ in range(N_horizon-1)] if dims.ns else []) + \
                         ([np.inf * ca.DM.ones((dims.ns_e, 1))] if dims.ns_e else [])
-        for i in range(0, N_horizon+1):
-            self._set_bounds_indices(i, lb_xtraj_node, ub_xtraj_node, lb_utraj_node, ub_utraj_node, constraints, dims)
+        if not single_shooting:
```
FreyJo: For readability, I always prefer `if` without `not` if both cases are treated.
Could you switch that up here, and in the following new ifs?

examples/acados_python/casadi_tests/test_casadi_single_shooting.py
```diff
@@ -0,0 +1,98 @@
...
+    casadi_x = np.array([casadi_ocp_solver.get(i, "x") for i in range(N_horizon + 1)])
+    acados_u = np.array([ocp_solver.get(i, "u") for i in range(N_horizon)])
+    casadi_u = np.array([casadi_ocp_solver.get(i, "u") for i in range(N_horizon)])
+
+    diff_u = np.linalg.norm(acados_u - casadi_u)
+    diff_x = np.linalg.norm(acados_x - casadi_x)
+    print(f"Difference in states: {diff_x}")
+    print(f"Difference in control inputs: {diff_u}")
+    if diff_u > 1e-5 or diff_x > 1e-5:
+        raise Exception("result differ significantly between acados and casadi.")
```
FreyJo, line 95: Please make sure this is tested, via cmake tests

interfaces/acados_template/acados_template/acados_casadi_ocp_solver.py
```diff
@@ -100,8 +100,18 @@ def __init__(self, ocp: AcadosOcp, with_hessian=False):
         utraj_node = []
         sl_node = []
         su_node = []
```
FreyJo: I think those `*_node` variables should be renamed to `*_list`, or `_nodes`, as otherwise, I would first associate it with a variable corresponding to just one node.

interfaces/acados_template/acados_template/acados_casadi_ocp_solver.py
```diff
@@ -662,13 +709,23 @@ def solve(self) -> int:
...
         # statistics
         solver_stats = self.casadi_solver.stats()
-        # timing = solver_stats['t_proc_total']
-        self.status = solver_stats['return_status']
-        self.nlp_iter = solver_stats['iter_count']
-        self.time_total = solver_stats['t_wall_total']
+        # timing = solver_stats['t_proc_total'] 
+        self.status = solver_stats['return_status'] if 'return_status' in solver_stats else solver_stats['success']
+        self.nlp_iter = solver_stats['iter_count'] if 'iter_count' in solver_stats else 0
+        self.time_total = solver_stats['t_wall_total'] if 't_wall_total' in solver_stats else 0
```
FreyJo: is this sometimes not in the `solver_stats`? For which solver?
I would prefer to default to `None` instead of `0` here.

examples/acados_python/casadi_tests/test_casadi_single_shooting.py
FreyJo: I guess this test does not work, as I don't see how you retrieve the x trajectory.
Do you plan to do this with a forward simulation?
Otherwise the test could also just compare the solution in the control variables.

interfaces/acados_template/acados_template/acados_casadi_ocp_solver.py
Pandatheon: The solver 'Alpaqa' doesn't have `solver_stats['return_status']` and `solver_stats['iter_count']`, but solvers like ipopt and fatrop have those, so I just make an ad hoc judgement here.

PR #1616, opened 2025-08-25
`AcadosCasadiOcp`: Add support for updating `y_ref`

Pandatheon:
- modify `AcadosCasadiOcp` for supporting updating `y_ref` by adding `y_ref` into parameter `p`
- modify `AcadosCasadiOcpSolver` for setting `y_ref`
- add a simple test for testing the support


interfaces/acados_template/acados_template/acados_casadi_ocp_solver.py
```diff
@@ -489,23 +557,26 @@ def _get_cost_node(self, i, N_horizon, xtraj_node, utraj_node, ptraj_node, sl_no
                     ptraj_node[0],
                     sl_node[0],
                     su_node[0],
-                    ocp.get_initial_cost_expression(),
+                    self.get_cost_expression(ocp, ptraj_node[0][dims.np:dims.ny+dims.np], i),
```
FreyJo: could you add an optional argument to the existing functions, `get_initial_cost_expression` and the like, for `y_ref` instead of adding `get_cost_expression`, which does not really belong to the `AcadosCasadi*` class IMO.

interfaces/acados_template/acados_template/acados_casadi_ocp_solver.py
```diff
@@ -458,6 +505,27 @@ def _append_variables_and_bounds(self, _field, w_sym_list, lbw_list, ubw_list, w
...
+        else:
+            ny = ocp.dims.ny_e
+            yref = ocp.cost.yref_e
+
+        p_list.append(ocp.parameter_values)
+        self._index_map['p_in_p_nlp'].append(list(range(self.offset_p, self.offset_p + ocp.dims.np)))
+        self.offset_p += ocp.dims.np
+        p_list.append(yref)
+        self._index_map['yref_in_p_nlp'].append(list(range(self.offset_p, self.offset_p + ny)))
+        self.offset_p += ny
```
FreyJo: This stuff should go to the `p` case of `_append_node`.
Ideally, all logic regarding the ordering of `p_node` is in one place.

interfaces/acados_template/acados_template/acados_casadi_ocp_solver.py
```diff
@@ -489,23 +557,26 @@ def _get_cost_node(self, i, N_horizon, xtraj_node, utraj_node, ptraj_node, sl_no
...
+                    ca.vertcat(ocp.model.p, ptraj_node[0][dims.np:dims.np+dims.ny]),
                     dims.ns_0, cost.zl_0, cost.Zl_0, cost.zu_0, cost.Zu_0)
         elif i < N_horizon:
             return (xtraj_node[i],
                     utraj_node[i],
                     ptraj_node[i],
                     sl_node[i],
                     su_node[i],
-                    ocp.get_path_cost_expression(),
+                    self.get_cost_expression(ocp, ptraj_node[i][dims.np:dims.ny+dims.np], i),
```
FreyJo: Here, there is  logic on the order within `p`. Better use the `index*` map instead that you added in this PR.

interfaces/acados_template/acados_template/acados_casadi_ocp_solver.py
```diff
@@ -489,23 +557,26 @@ def _get_cost_node(self, i, N_horizon, xtraj_node, utraj_node, ptraj_node, sl_no
...
                     dims.ns, cost.zl, cost.Zl, cost.zu, cost.Zu)
         else:
             return (xtraj_node[-1],
                     [],
                     ptraj_node[-1],
                     sl_node[-1],
                     su_node[-1],
-                    ocp.get_terminal_cost_expression(),
+                    self.get_cost_expression(ocp, ptraj_node[-1][dims.np:(dims.nx if dims.ny else 0)+dims.np], i),
+                    ca.vertcat(ocp.model.p, ptraj_node[-1][dims.np:(dims.nx if dims.ny else 0)+dims.np]),
```
FreyJo: Also here. Plus, I don't know why there is an `nx`. I think it should only depend on `ny`. Am I missing something?

interfaces/acados_template/acados_template/acados_casadi_ocp_solver.py
Pandatheon: I directly separated the nodes and values formulation of `p` into a different section with a function `_append_node_and_value_for_params`, I hope this organization makes sense.

PR #1622, opened 2025-09-06
 Ros2 Node Generation

ArgoJ:
A generic Ros2 Humble Node generation for a simple MPC including generic messages. 
Create your `AcadosOcp` and add the `AcadosOcpRosOptions` class to the ocp variable `ros_opts` as below:
```python
from acados_template import AcadosOcpRosOptions, AcadosOcp

ocp = AcadosOcp()
...

ocp.ros_opts = AcadosOcpRosOptions()
ocp.ros_opts.package_name = "ocp_pkg" 
```

That's it. Two packages are created next to the `c_generated_code` folder. One which includes the interface for the messages (here `ocp_pkg_interface`) to communicate with the main package, and the main package with MPC functionality (here `ocp_pkg`).
All weights and constraints are implemented as node parameters, which can be changed during runtime. The change of `ts` only changes the control loop timer and not yet the `ts` of the OCP solver. The references, OCP parameters as well as state can be updated via the given message which are included in the interface package.

Look at the generated `Redme.md` in the main folder to see how to build, start and test the generated packages.

This MPC is triggered by a timer, other versions like topic trigger are not implemented yet.
Other architecture types like lifecycle nodes are also not implemented yet as well as services or actions (other form of messages).

Similarly, a ROS2 node for the `AcadosSimSolver` can be generated.
Additionally, some improvements to the CI workflow are in this PR.


ArgoJ:
I found an issue, again where the pendulum example will not build because I left in the parameter msg as well as 2 typos where i didnt use the {{ model.name }} but the actual name that ive testet with.

interfaces/acados_template/acados_template/c_templates_tera/ros2/ocp_interface_templates/package.in.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
+    <name>{{ ros_opts.package_name }}_interface</name>
+    <version>0.1.0</version>
+    <description>A Generic Interface for the acados solver</description>
+    <maintainer email="acados.ros@email.com">Acados Ros</maintainer>
```
FreyJo: Can we not add an email? Or instead use an existing one?
e.g feel free to add yours

interfaces/acados_template/acados_template/c_templates_tera/ros2/ocp_node_templates/CMakeLists.in.txt
```diff
@@ -0,0 +1,107 @@
...
+    $<INSTALL_INTERFACE:include>
+)
+# Link Libraries
+target_link_libraries({{ ros_opts.node_name }} 
+    ${ACADOS_GENERATED_LIB}
+    ${ACADOS_LIB_DIR}/libacados.so
+    ${ACADOS_LIB_DIR}/libblasfeo.so
+    ${ACADOS_LIB_DIR}/libhpipm.so
+    ${ACADOS_LIB_DIR}/libosqp.so
+    ${ACADOS_LIB_DIR}/libqpOASES_e.so
```
FreyJo: OSQP and qpOASES are only optionally installed, so we should check if they need to be linked against. Similar to how it is done in https://github.com/acados/acados/blob/main/interfaces/acados_template/acados_template/c_templates_tera/CMakeLists.in.txt

interfaces/acados_template/acados_template/c_templates_tera/ros2/ocp_node_templates/CMakeLists.in.txt
```diff
@@ -0,0 +1,107 @@
...
+if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  add_compile_options(-Wall -Wextra -Wpedantic)
+endif()
+
+# --- ROS DEPENDENCIES ---
+find_package(ament_cmake REQUIRED)
+find_package(rclcpp REQUIRED)
+find_package(std_msgs REQUIRED)
+find_package({{ ros_opts.package_name }}_interface REQUIRED)
+find_package(OpenMP REQUIRED)
```
FreyJo, line 19: OpenMP is optional, so we should only require this if actually needed, similar to how it is done in https://github.com/acados/acados/blob/main/interfaces/acados_template/acados_template/c_templates_tera/CMakeLists.in.txt

interfaces/acados_template/acados_template/c_templates_tera/ros2/ocp_node_templates/CMakeLists.in.txt
```diff
@@ -0,0 +1,107 @@
...
+    ${ACADOS_INCLUDE_PATH}/qpOASES_e
+)
+
+ament_export_libraries(
+    acados_ocp_solver_{{ name }}
+    ${ACADOS_LIB_DIR}/libacados.so
+    ${ACADOS_LIB_DIR}/libblasfeo.so
+    ${ACADOS_LIB_DIR}/libhpipm.so
+    ${ACADOS_LIB_DIR}/libosqp.so
+    ${ACADOS_LIB_DIR}/libqpOASES_e.so
```
FreyJo: also here

interfaces/acados_template/acados_template/c_templates_tera/ros2/ocp_node_templates/CMakeLists.in.txt
```diff
@@ -0,0 +1,107 @@
...
+# --- EXPORTS ---
+ament_export_include_directories(
+    "include"
+    ${ACADOS_INCLUDE_PATH}
+    ${ACADOS_INCLUDE_PATH}/acados
+    ${ACADOS_INCLUDE_PATH}/acados_c
+    ${ACADOS_INCLUDE_PATH}/blasfeo/include
+    ${ACADOS_INCLUDE_PATH}/hpipm/include
+    ${ACADOS_INCLUDE_PATH}/osqp
+    ${ACADOS_INCLUDE_PATH}/qpOASES_e
```
FreyJo: I think adding empty/ non existing include dirs is not an issue. So this should be fine.

ArgoJ:
one python interface ctest failure. I dont know why. I changed the workflow structure a little to maintain reusability in different jobs. Ros2 workflow works now.

ArgoJ:
so finally fixed. every test works.

.github/workflows/core_build.yml
```diff
@@ -0,0 +1,81 @@
+name: Reusable Core Build
+
+on:
+  workflow_call:
+    # Wir definieren einen Output auf Workflow-Ebene.
+    # So kann der aufrufende Workflow den Wert empfangen.
```
FreyJo: Please use english comments.

.github/workflows/full_build.yml
```diff
@@ -9,266 +9,206 @@ on:
...
-    - name: Store include directory (/include)
-      uses: actions/upload-artifact@v4
-      with:
-        name: include
-        if-no-files-found: error
-        path: ${{runner.workspace}}/acados/include/
-        compression-level: 0
-        overwrite: true
+  call_core_build:
+    uses: ./.github/workflows/core_build.yml
```
FreyJo, line 14: I see that it might make sense to structure this into a separate file.
However, especially given this is already a large PR, it would have been better to avoid unrelated changes.

.github/workflows/full_build.yml
```diff
@@ -9,266 +9,206 @@ on:
...
 
+    - name: Rerun failed tests with verbose output
+      if: failure()
+      working-directory: ${{ github.workspace }}/build
+      shell: bash
+      run: |
+        echo "--- CTest failed. Re-running all failing tests with verbose output: ---"
+        source ${{ github.workspace }}/acadosenv/bin/activate
+        ctest -C Release --rerun-failed -V
+        exit 1
```
FreyJo, line 51: Did this help to get better output?
Does it provide more info than the `--output-on-failure` option above?

.github/workflows/full_build.yml
```diff
@@ -400,49 +352,55 @@ jobs:
...
     steps:
-    - uses: actions/checkout@v4
+    - name: Checkout repository
+      uses: actions/checkout@v4
       with:
-        submodules: 'recursive'
+          submodules: 'recursive'
+    
+    - name: Install System Dependencies
+      uses: ./.github/actions/setup-system-dependencies
```
FreyJo: why is this added here? I think it is not needed.

.github/workflows/full_build.yml
```diff
@@ -464,79 +422,83 @@ jobs:
...
-
-    - uses: actions/checkout@v4
+    - name: Checkout repository
+      uses: actions/checkout@v4
       with:
-        submodules: 'recursive'
+          submodules: 'recursive'
+    
+    - name: Install System Dependencies
+      uses: ./.github/actions/setup-system-dependencies
```
FreyJo: Probably, not needed. If so, please remove.

.github/workflows/ros2_build.yml
```diff
@@ -0,0 +1,82 @@
...
+        WS=${{ github.workspace }}/ros_pendulum_ws
+        mkdir -p $WS/src
+        rsync -a . $WS/src/pendulum_on_cart
+
+        cd $WS
+        rosdep update || true
+        rosdep install --from-paths src --ignore-src -r -y || true
+
+        colcon build --packages-select pendulum_on_cart pendulum_on_cart_interface --event-handlers console_direct+
+        source install/setup.bash
```
FreyJo: I think it would make sense to divide this step into 3: 1) Python export 2) Ros compilation 3) Ros run

interfaces/acados_template/acados_template/c_templates_tera/ros2/ocp_interface_templates/CMakeLists.in.txt
```diff
@@ -0,0 +1,40 @@
...
+  {%- endif %}
+)
+
+# set(SRV_FILES
+#   "srv/SolveOCP.srv"
+# )
+
+# set(ACTION_FILES
+#   "srv/SolveOCP.srv"
+# )
```
FreyJo: are these comments useful / needed? If not, please remove.

interfaces/acados_template/acados_template/c_templates_tera/ros2/ocp_node_templates/config.in.hpp
```diff
@@ -0,0 +1,199 @@
...
+    {%- if dims.nbx_e > 0 %}
+    std::array<double, {{ model.name | upper }}_NBXN> lbx_e{};
+    std::array<double, {{ model.name | upper }}_NBXN> ubx_e{};
+    {%- endif %}
+    {%- if dims.nbu > 0 %}
+    
+    // Input Bounds
+    std::array<double, {{ model.name | upper }}_NBU> lbu{};
+    std::array<double, {{ model.name | upper }}_NBU> ubu{};
+    {%- endif %}
```
FreyJo: Is the code for updating the cost weights and constraint bounds fully functional / tested?
I see that the values are basically written into the memory when code generating the ROS wrapper.
However, this is what is already done when we create the acados solver itself and there the workflow of going from values in Python via code generation to the templated solver is quite well tested.
So, I am not sure if the functions added here to overwrite those values provide more an additional source of error or actually add functionality.
Can you please clarify?

interfaces/acados_template/acados_template/acados_ocp.py
```diff
@@ -114,6 +115,7 @@ def __init__(self, acados_path=''):
         self.__p_global_values = np.array([])
         self.__problem_class = 'OCP'
         self.__json_file = "acados_ocp.json"
+        self.__ros_opts: Optional[AcadosOcpRos] = None
```
FreyJo: I think renaming `AcadosOcpRos` to `AcadosOcpRosOpts` or `AcadosOcpRosOptions` would be good.

interfaces/acados_template/acados_template/ros2/__init__.py
```diff
@@ -0,0 +1,2 @@
+from .ocp_node import AcadosOcpRos
+from .utils import ArchType, ControlLoopExec
```
FreyJo: I think the init file here can just be an empty line.
Instead the parent init needs to import all this to expose it through the Python package.
Maybe you can test by using the options in the example explicitly.

interfaces/acados_template/acados_template/ros2/utils.py
```diff
@@ -0,0 +1,115 @@
...
+class ArchType(str, Enum):
+    NODE = "node"
+    LIFECYCLE_NODE = "lifecycle_node"
+    ROS2_CONTROLLER = "ros2_controller"
+    NAV2_CONTROLLER = "nav2_controller"
+    
+    
+class AcadosRosBase:
+    def __init__(self):
+        self._package_name: str = "acados_base"
```
FreyJo: I suggest to rename this to `AcadosRosOptsBase` and put this and the OCP specific opts into one file, `opts.py`.
There could be a Ros Builder, or a Ros node class to interact with the Ros node via Python.

interfaces/acados_template/acados_template/acados_ocp.py
```diff
@@ -1496,7 +1574,9 @@ def to_dict(self) -> dict:
         # convert acados classes to dicts
         for key, v in ocp_dict.items():
             if isinstance(v, (AcadosModel, AcadosOcpDims, AcadosOcpConstraints, AcadosOcpCost, AcadosOcpOptions, ZoroDescription)):
-                ocp_dict[key]=dict(getattr(self, key).__dict__)
+                ocp_dict[key] = dict(getattr(self, key).__dict__)
+            if isinstance(v, AcadosOcpRos) and v is not None:
```
FreyJo: I think the `None` check is redundant here.

acados_install.sh
FreyJo, line 1: I think this is your private install file, probably added accidentally.
Please remove from the PR. You can rename to `private_acados_install.sh` to have it gitignored.

.github/workflows/full_build.yml
FreyJo: But the suggested structure overall makes sense!

interfaces/acados_template/acados_template/c_templates_tera/ros2/ocp_node_templates/node.in.cpp
```diff
@@ -0,0 +1,619 @@
...
+void {{ ClassName }}::solver_status_behaviour(int status) {
+    if (status == ACADOS_SUCCESS) {
+        this->get_input(u0_.data(), 0);
+        this->publish_input(u0_, status);
+        {%- if solver_options.nlp_solver_type == "SQP_RTI" %}
+        first_solve_ = false;
+        this->prepare_rti_solve();
+        {%- endif %}
+    } else {
+        this->publish_input(u0_default_, status);
```
FreyJo: I did not see what `u0_default_` is set to. I think it would be better to always return the `u0` from the solver. E.g. if the solver has returned max. number of iterations, the solution can still be useful.
Maybe one would additionally want to return the residuals to check how good the solution is.

interfaces/acados_template/acados_template/c_templates_tera/ros2/ocp_node_templates/node.in.cpp
```diff
@@ -0,0 +1,619 @@
...
+    int status = this->ocp_solve();
+    {%- endif %}
+    this->solver_status_behaviour(status);
+}
+
+void {{ ClassName }}::solver_status_behaviour(int status) {
+    if (status == ACADOS_SUCCESS) {
+        this->get_input(u0_.data(), 0);
+        this->publish_input(u0_, status);
+        {%- if solver_options.nlp_solver_type == "SQP_RTI" %}
```
FreyJo: Nice that you also implemented RTI with split preparation and feedback phase. It seems to be done correctly!

ArgoJ:
Today or tomorrow I will also add a proper test for each generated node. Already done but I need to push it to main and fork it, I also want to add a GitHub action that does this testing in a sort of function, so that for future tests on this, we don't need to write all of this. Then the integration test is more seamless and you can also test the package with colcon test after build. It's nothing special but it's generic and always included in the package. It tests node parameter getting and setting as well as publishing and subscription

ArgoJ:
Note: For some reason the pull request MATLAB on windows test works here while it does not in my forked main. I didn't change anything related to Matlab or acados itself, but keep an eye on it please. Maybe it's also just false indexing by GitHub or whatever.

FreyJo:
I don't know what is going with the actions on your fork. Please just look at the runs here for now.
Just let me know when I should review again. But there is no rush, I am also quite busy this week.

.github/workflows/full_build.yml
ArgoJ: Yes this helped because the logs only show a quater and potentially not the one with the issue inside due to 4 threads.

.github/workflows/full_build.yml
ArgoJ: If you like I can also change the full build into smaller chunks and add them together in one gib build that should be more easy readable.

.github/workflows/full_build.yml
ArgoJ: Yeah I'll try to run it without. I found it as a fix for some problem I don't remember.

.github/workflows/ros2_build.yml
ArgoJ: Yes already did this

interfaces/acados_template/acados_template/c_templates_tera/ros2/ocp_interface_templates/CMakeLists.in.txt
ArgoJ: Not usefull for now but the idea was that it's already there if someone wants to implement this part but I can remove it

interfaces/acados_template/acados_template/c_templates_tera/ros2/ocp_interface_templates/package.in.xml
ArgoJ: Initially I had it as an option but I think nobody will use this, maybe I am wrong but yeah I can use my email or yours?

interfaces/acados_template/acados_template/c_templates_tera/ros2/ocp_node_templates/config.in.hpp
ArgoJ: Theoretically I can leave this out and always call the acados function when I want to get these values. The problem is that I think the setting of these values is a little harder with the parameters. Either way I will try to have a look at it.

interfaces/acados_template/acados_template/ros2/utils.py
ArgoJ: Uff directly In python I don't know of this would work that easy. I would say this is not the priority now

interfaces/acados_template/acados_template/c_templates_tera/ros2/ocp_node_templates/node.in.cpp
ArgoJ: Okay that's a good insight I don't know. I will change that

ArgoJ:
ups that was not correct sorry you dont need to review again xD

interfaces/acados_template/acados_template/c_templates_tera/ros2/ocp_node_templates/node.in.cpp
ArgoJ: it is set to all 0, the idea was that if the solver fails i dont want it to do something unexpected but actually one can solve this in a mapper node that we need anyways for the topics to keep it safe.

interfaces/acados_template/acados_template/ros2/utils.py
ArgoJ: AcadosRosBase should be the base also for other things like the sim solver or the mapper node. I want to keep these in individual filese, therefore i need one file where all the relevant utils are inside. If you would like it in another way, so that i put everything in one file i can do that. I just think its harder to navigate through it :D

interfaces/acados_template/acados_template/c_templates_tera/ros2/ocp_node_templates/CMakeLists.in.txt
ArgoJ: Where do i get the openmp flag from? the json does not have such flag right? is it just the with batch functinoality? I thought I can just use open mp and it always runs with it :( Always did it wrong I guess.

ArgoJ:
Can you look into the matlab testing problem? Now it is also here in this test, even though i only deleted my acados install sh

ArgoJ:
I had implemented a caching of the core build as long as no cachanges were made in the submodules, acados folder and cmakelists. If you would to keep this have a look at 
commit 6ce32a5d271aa00cf127ce118c11b809e7cd301f
 - took 1 minute less with the build
 
 Caching ros2 humble did not work because of a lot of missing dependencies elsewise.

interfaces/acados_template/acados_template/ros2/__init__.py
ArgoJ: Left it in, for future work when i implement different arch types control loop executions and stuff. So that I dont clutter the main init.

interfaces/acados_template/acados_template/c_templates_tera/ros2/ocp_node_templates/config.in.hpp
ArgoJ: Fixed, I deleted all configs that are already in the acados solver. Downside is that I cannot log all parameters at once but I think it is still a better practice.

ArgoJ:
so simulation is also added, the mapper is still missing. Maybe you can give a quick comment. My idea was that i dont render it with acados ocp or acados sim but in a seperate class that is completely off the other ones, because that would generate unrelated deps, in acados ocp or sim.

FreyJo:
I don't know what the "mapper" is supposed to do.
Maybe it is better to do that in a follow-up PR?
If you agree, I will do a second round of review here tomorrow.

ArgoJ:
Yes I can do a follow up pr sorry😬
The mapper is an extra node that is basically a middle man between the ocp and the simulator node. My plan is to just give it both jsons and create it out of this. For the ocp and the simulator the states are often times different to each other, that's why I kept the messages unique for both. This way I can still maintain a flexible code base, but I always need a message map from ocp to sim.

E.g.:
$x_{0,ocp}$ &rarr; $x_{4,sim}$
$x_{1,ocp}$ &rarr; $x_{5,sim}$
$x_{2,ocp}$ &rarr; $x_{7,sim}$
....

Same goes for the input

.github/workflows/core_build.yml
```diff
@@ -0,0 +1,109 @@
...
```
FreyJo: The comment about german comments is not addressed here.

.github/workflows/core_build.yml
```diff
@@ -0,0 +1,109 @@
...
+          ${{ github.workspace }}/include
+        key: ${{ runner.os }}-${{ env.BUILD_TYPE }}-acados-${{ steps.submods.outputs.sha }}-${{ hashFiles(
+            '.gitmodules',
+            'CMakeLists.txt',
+            'cmake/**/*.cmake',
+            'acados/**'
+          ) }}
+
+    - name: Install System Dependencies
+      uses: ./.github/actions/setup-system-dependencies
```
FreyJo: why is this added?
It installs BLAS and LAPACK. Those should not be dependencies.
Is this only needed for the ROS2 build? If so, only install it in that action.

.github/workflows/core_build.yml
```diff
@@ -0,0 +1,109 @@
...
+      with:
+        path: |
+          ${{ github.workspace }}/build
+          ${{ github.workspace }}/lib
+          ${{ github.workspace }}/include
+        key: ${{ runner.os }}-${{ env.BUILD_TYPE }}-acados-${{ steps.submods.outputs.sha }}-${{ hashFiles(
+            '.gitmodules',
+            'CMakeLists.txt',
+            'cmake/**/*.cmake',
+            'acados/**'
```
FreyJo: So if non of this changes, acados is not recompiled but the artifacts are retrieved from cache, right?
That is nice!
I think `interfaces/acados_c` should be added to the list. Then it is complete :)

.github/workflows/ros2_build.yml
```diff
@@ -0,0 +1,52 @@
...
+    branches-ignore:
+      - 'doc*'
+      - 'wip*'
+  pull_request:
+    branches:
+      - '*'
+
+jobs:
+  call_core_build:
+    uses: ./.github/workflows/core_build.yml
```
FreyJo: If the `ros2_humble_test` was just a job inside `full_build`, we only need to call `call_core_build` once. Now it is called twice.
I would thus move this job to the `full_build` file. Can you do that? Or is there a strong reason to not do that?

examples/acados_python/pendulum_on_cart/ros2/example_ros_minimal_ocp.py
```diff
@@ -0,0 +1,151 @@
...
+    # set prediction horizon
+    ocp.solver_options.tf = Tf
+
+    # Ros stuff
+    ocp.ros_opts = AcadosOcpRosOptions()
+    ocp.ros_opts.package_name = "pendulum_on_cart_ocp"
+
+    export_code = script_dir / 'generated_ocp'
+    ocp.code_export_directory = str(export_code / "c_generated_code")
+    ocp_solver = AcadosOcpSolver(ocp, json_file = str(export_code / 'acados_ocp.json'))
```
FreyJo: Would be better to use `os.path.join` here.

examples/acados_python/pendulum_on_cart/ros2/example_ros_minimal_sim.py
```diff
@@ -0,0 +1,53 @@
...
+    sim.solver_options.num_steps = 3
+    sim.solver_options.newton_iter = 3 # for implicit integrator
+    sim.solver_options.collocation_type = "GAUSS_RADAU_IIA"
+
+    sim.ros_opts = AcadosSimRosOptions()
+    sim.ros_opts.package_name = "pendulum_on_cart_sim"
+
+    export_code = script_dir / 'generated_sim'
+    sim.code_export_directory = str(export_code / "c_generated_code")
+    acados_integrator = AcadosSimSolver(sim, json_file=str(export_code / 'acados_sim.json'))
```
FreyJo: Would be better to use `os.path.join` here.

interfaces/acados_template/acados_template/ros2/utils.py
```diff
@@ -0,0 +1,126 @@
...
+    NAV2_CONTROLLER = "nav2_controller"
+
+
+class AcadosRosBaseOptions:
+    def __init__(self):
+        self.__package_name: str = "acados_base"
+        self.__node_name: str = "acados_base_node"
+        self.__namespace: str = ""
+        self.__archtype: str = ArchType.NODE.value
+        self.__control_loop_executor: str = ControlLoopExec.TIMER.value
```
FreyJo, line 26: change the typehints here to `ControlLoopExec`, and `ArchType`?
also in the properties below

FreyJo (commented):
Few minor comments. From my side this can be merged, when they are addressed.

.github/workflows/core_build.yml
ArgoJ: Whoops forgot to delete it.

.github/workflows/core_build.yml
ArgoJ: Same here forgot it

ArgoJ: Same here forgot it

.github/workflows/core_build.yml
ArgoJ: Yes exactly. There should be all folders or files that have any impact on the build of acados.

.github/workflows/ros2_build.yml
ArgoJ: I can do this I think

examples/acados_python/pendulum_on_cart/ros2/example_ros_minimal_ocp.py
ArgoJ: Ah yeah this was just for me because my setup worked easier with it to test locally. I will fix this

interfaces/acados_template/acados_template/ros2/utils.py
ArgoJ: No I explicitly didn't use the types because I thought no one likes to import so much stuff and then I have it directly as a string for the jsons and so on. That's why I also used the value of the enum. But if you think it is a better way to always use the enum directly and don't allow strings I can do this

FreyJo: ok, makes sense :)

PR #1627, opened 2025-09-10
interfaces for terminal equality state constraints

yf-gao:


sandmaennchen:
Hi @yf-gao you need to add the new fields also in MATLAB to have the tests pass

PR #1631, opened 2025-09-14
From json methods for ocp and sim

ArgoJ:
A classmethod from which we can simply set the values from the generated json (`from_json()`). Also includes functions for subclasses `from_dict()`.

```python
from acados_template import AcadosOcp, AcadosSim

ocp = AcadosOcp.from_json(file_name)
sim = AcadosSim.from_json(file_name)
```

Then the standard stuff can be done, like putting it into the solvers.

PS: this is just a suggestion and not mandatory, also not for the acados internal Ros2 Node generation implementation. I can also do this with just dict calls, but I thought this functionallity may be elsewhere also useful.

### DIFF
--- a/.github/workflows/core_build.yml
+++ b/.github/workflows/core_build.yml
@@ -59,6 +59,7 @@ jobs:
             'CMakeLists.txt',
             'cmake/**/*.cmake',
             'acados/**',
+            'interfaces/CMakeLists.txt',
             'interfaces/acados_c/**'
           ) }}
 


interfaces/acados_template/acados_template/acados_sim.py
```diff
@@ -306,6 +306,76 @@ def with_batch_functionality(self, with_batch_functionality):
...
+    def from_dict(cls, data: dict, *, strict: bool = False, allow_none: bool = True) -> "AcadosSimOptions":
+        """
+        Create an `AcadosSimOptions` instance from a dictionary.
+        
+        Parameters
+        ----------
+        data : dict
+            Dictionary containing simulation options entries.
+        strict : bool, optional
+            If True, raise KeyError for unknown keys; otherwise, unknown keys are ignored. Default True.
```
FreyJo: Default not documented correctly.

examples/acados_python/tests/test_ocp_sim_from_json.py
```diff
@@ -0,0 +1,38 @@
...
+    ocp.json_file = str(ocp.json_file).replace('.json', '_out.json')
+    ocp.make_consistent()
+    ocp_dict = ocp.to_dict()
+    pprint(ocp_dict)
+    
+    ocp.dump_to_json()
+    
+    
+def test_sim_json(file_name: str | Path):
+    sim = AcadosSim.from_json(file_name)
```
FreyJo: Where is the json file originally created?
Maybe you run this after some other test.
Maybe this will be part of a Github actions file later.

interfaces/acados_template/acados_template/acados_sim.py
FreyJo: also for OCP.

examples/acados_python/tests/test_ocp_sim_from_json.py
```diff
@@ -0,0 +1,38 @@
...
+    
+    json_file = str(file_name).replace('.json', '_out.json')
+    sim.make_consistent()
+    sim_dict = sim.to_dict()
+    pprint(sim_dict)
+    
+    sim.dump_to_json(json_file)
```
FreyJo: Or the test could: dump to json, load from json, dump again to another file and then check that the two json files match.

sandmaennchen:
Hi Josua, great that you are working on this! I will have a closer look next week (also at my try at implementing this a while ago). I just remember that `has_x0` was a bit of a problem in my implementation as this property does not have a setter. Might be relevant for you as well not sure. 
I'll get back to you next week!

ArgoJ:
Hey, for the has x0, I do actually exactly the what the initial setting of hasx0 does. In my test case it works and logically it should also work but, I haven't wrote a test or multiple tests on this yet.

sandmaennchen:
And do we really need to be able to load the CasADi expressions?
Because (de)serialization of CasADi expressions is a bit more involved than what we do at the moment (I think)

See
https://github.com/casadi/casadi/blob/3c11fa702530c018fb2a7afc028b466bd4afc04c/test/python/misc.py#L350
https://github.com/casadi/casadi/issues/2560

sandmaennchen:
Ah now I understand your approach. The expressions are needed as you want to call `make_consistent` which is where `has_x0` is then set correctly.

I think this is the way to go, but then we should make sure that the serialization is really working

ArgoJ:
Yeah one drawback might be the handling of the serialisation in the first place. So you're right I need to put in a check to validate that the expression is indeed correct. Another additional option is to handle the serialisation differently as you currently do. If I remember correct only the SX is currently serialised even though casadi provides support for SX, MX and DM. Therefore I could just change this to nested try catch blocks. This would at least make sure that serialisation is always tried.

sandmaennchen:
Yes at the moment this function handles "serialization". 
https://github.com/acados/acados/blob/51b379b620b073bb7fce6c77eafcedcd15b04bcc/interfaces/acados_template/acados_template/utils.py#L364

But I dont think this is working even for SX (you cannot serialize an individual SX expression, you also need to know about the variables). 
I guess this should go into separate `serialize` and `deserialiue` functions in `AcadosModel`. And maybe this should be a separate PR 😜

ArgoJ:
So I now added the true or at least tried the option to avoid free casadi variables.

sandmaennchen:
Could you have a look at this PR https://github.com/acados/acados/pull/1639? I think that's the way CasADi provides serialization of expressions

PR #1632, opened 2025-09-16
Acados Ros mapper for OCP to SIM messages

ArgoJ:
This is an addition to the previous ros pull request. It includes some changes regarding multithreading, a generated dir to specify the package directory, while I also added a `json_file` field to `AcadosSim`. This PR also includes some formating of the ros things. 
However, the main part is the **ros message mapper* for all sorts of messages. This map can automatically be created for the ocp and sim, by just passing the instances to it like so:

```python
from acados_template import AcadosOcp, AcadosSim, AcadosOcpSolver, AcadosSimSolver
from acados_template.ros2 import RosTopicMapper, AcadosOcpRosOptions, AcadosSimRosOptions

ocp = AcadosOcp()
...
ocp.ros_opts = AcadosOcpRosOptions()
ocp.ros_opts.package_name = "pendulum_on_cart_ocp"
ocp_solver = AcadosOcpSolver(ocp)

sim = AcadosSim()
...
sim.ros_opts = AcadosSimRosOptions()
sim.ros_opts.package_name = "pendulum_on_cart_sim"
sim_solver = AcadosSimSolver(sim)

# The ros_opts for ocp and sim have to be set for the mapper
ros_mapper = RosTopicMapper.from_instances(ocp_solver, sim_solver)
ros_mapper.package_name = "sim_ocp_mapper"

ros_mapper.generate()
```

This generates all the relevant packages for ocp, sim, interfaces and mapper. 
Note: For this functionality, the casadi symetrics that correspond to each other need to be named the same. 

However, one can also use this mapper to build a direct state message to a standard format like Twist, Pose or whatever, like so:

```python
from acados_template import AcadosOcp, AcadosSim, AcadosOcpSolver, AcadosSimSolver
from acados_template.ros2 import RosTopicMapper, AcadosOcpRosOptions, AcadosSimRosOptions
from acados_template.ros2.default_msgs import GEOMETRY_MSGS_TWIST

ocp = AcadosOcp()
...
ocp.ros_opts = AcadosOcpRosOptions()
ocp.ros_opts.package_name = "ocp"
AcadosOcpSolver(ocp)

# The ros_opts for ocp and sim have to be set for the mapper
# --- SETUP MESSAGES --- 
# We create all messages here that the mapper should subscribe and publish.
# e.g. /pose.position.x -> /ocp_state.x[0] 
#      /ocp_control.u[0] -> /cmd_vel.linear.x
in_ocp_ctrl  = build_default_control(ocp, direction_out=False)

[...]

PR #1634, opened 2025-09-17
Minor fixes for ROS2 node template

tpoignonec:
Some propositions related to the generation of ROS2 nodes (see PR#1622):
- Add env-hooks to add Acados libs to `LD_LIBRARY_PATH` automatically in ROS2 workspaces
- Fix the setter for the period that is currently called twice at startup and has potentially undefined behaviors without verbose. For instance, if the parameter `Ts` is set first, the timer is started even if the solver is not yet ready.
- Upon a period change (ROS2 parameter), also change the integration period.

On this last point, should it be done for the OCP solver also?


ArgoJ:
Hey, I would suggest keeping the ts of the timer seperated from the ts of the ocp. Otherwise I would change the parameter name to {{ ros_opts.node_name }}.solver_options.Tsim. Then its clear that this directly corresponds to the acados ocp.

tpoignonec:
> Hey, I would suggest keeping the ts of the timer seperated from the ts of the ocp. Otherwise I would change the parameter name to {{ ros_opts.node_name }}.solver_options.Tsim. Then its clear that this directly corresponds to the acados ocp.

Only for the OCP solver right ?
The  way I understand it, the SIM solver Tsim and the period of the timer are the same thing.

ArgoJ:
Yes that's correct.

FreyJo:
Thanks a lot @tpoignonec for checking the implementation and creating this PR!
Can you update the sim according to the discussion?
Then, if there are no further comments by @ArgoJ, I would merge this.

ArgoJ:
Sounds nice to me

tpoignonec:
@ArgoJ FYI, I removed the prefixed node name in the parameters.

Before:

```bash
$ ros2 param dump pendulum_on_cart_sim_node

/pendulum_on_cart_sim_node:
  ros__parameters:
    pendulum_on_cart_sim_node:
      solver_options:
        Tsim: 0.1
    qos_overrides:
      /parameter_events:
        publisher:
          depth: 1000
          durability: volatile
          history: keep_last
          reliability: reliable
    start_type_description_service: true
    use_sim_time: false
```

After

```bash
/pendulum_on_cart_sim_node:
  ros__parameters:
    qos_overrides:
      /parameter_events:
        publisher:
          depth: 1000
          durability: volatile
          history: keep_last
          reliability: reliable
    solver_options:
      Tsim: 0.1
    start_type_description_service: true
    use_sim_time: false
```

tpoignonec:
LGTM

ArgoJ:
Well I think this is okay, but then due to consistency I have to change it everywhere else. Maybe you could move the parameter renaming to a seperate PR? and maybe do this after my latest PR got accepted, elsewise I think merging will be awfull. Stick to the previous implementation until then. But thanks for the work, I think this makes it cleaner. So, good idea :D

FreyJo:
Sorry @tpoignonec, I went ahead merging #1632 first.
Now there are two conflicts here. I would appreciate if you could resolve them.
Then, I will merge.

ArgoJ:
Theoretically you can also add the param name changes you suggested. Else I will do a pr on that tomorrow.

tpoignonec:
It should be mergable now.
I also changed the parameters names.

PR #1639, opened 2025-09-22
Python: Serialization of CasADi expression in `AcadosModel`

sandmaennchen:
Add `serialize`, `deserialize` and `from_dict`, `to_dict` methods to `AcadosModel`.

TODO: Test should be improved

### DIFF
--- a/interfaces/acados_template/acados_template/acados_model.py
+++ b/interfaces/acados_template/acados_template/acados_model.py
@@ -1006,10 +1006,10 @@ class AcadosModel():
 
         for k, _ in inspect.getmembers(type(self), lambda v: isinstance(v, property)):
             v = getattr(self, k)
-            if not isinstance(v, (ca.SX, ca.MX)):
-                model_dict[k] = v
+            if isinstance(v, (ca.SX, ca.MX)):
+                model_dict[k] = repr(v) # only for debugging
             else:
-                model_dict[k] = "" # TODO: maybe add some human-readable string representation for debugging
+                model_dict[k] = v
 
         model_dict['serialized_expressions'], model_dict['expression_names'] = self.serialize()
 
@@ -1041,8 +1041,8 @@ class AcadosModel():
                 warnings.warn(f"Attribute {attr} not in dictionary.")
             else:
                 try:
-                    # check whether value is not the empty list
-                    if not (isinstance(value, list) and not value):
+                    # check whether value is not the empty list and not a CasADi symbol/expression
+                    if not (isinstance(value, list) and not value) and not attr in expression_names:
                         setattr(model, attr, value)
                 except Exception as e:
                     Exception("Failed to load attribute {attr} from dictionary:\n" + repr(e))
--- a/interfaces/acados_template/acados_template/acados_multiphase_ocp.py
+++ b/interfaces/acados_template/acados_template/acados_multiphase_ocp.py
@@ -372,6 +372,11 @@ class AcadosMultiphaseOcp:
 
 
     def dump_to_json(self) -> None:
+        """
+        Dumps the OCP description in JSON format to AcadosOcp.json_file.
+
+        NOTE: Please make sure to call `AcadosOcp.make_consistent()` before dumping to json.
+        """
         ocp_nlp_dict = self.to_dict()
         with open(self.json_file, 'w') as f:
             json.dump(ocp_nlp_dict, f, default=make_object_json_dumpable, indent=4, sort_keys=True)


sandmaennchen:
@ArgoJ what do you think?

interfaces/acados_template/acados_template/acados_model.py
```diff
@@ -964,5 +965,79 @@ def reformulate_with_polynomial_control(self, degree: int) -> None:
...
+        model = self()
+
+        # loop over all properties
+        for attr, _ in inspect.getmembers(type(model), lambda v: isinstance(v, property)):
+
+            value = model_dict.get(attr)
+
+            if value is None:
+                warnings.warn(f"Attribute {attr} not in dictionary.")
+            else:
```
FreyJo, line 1042: This now warns for all symbolic entries, that are later loaded via `serialized_expressions`.
Can we avoid this?

interfaces/acados_template/acados_template/acados_model.py
```diff
@@ -964,5 +965,79 @@ def reformulate_with_polynomial_control(self, degree: int) -> None:
...
+        model_dict = {}
+
+        for k, _ in inspect.getmembers(type(self), lambda v: isinstance(v, property)):
+            v = getattr(self, k)
+            if not isinstance(v, (ca.SX, ca.MX)):
+                model_dict[k] = v
+
+        model_dict['serialized_expressions'], model_dict['expression_names'] = self.serialize()
+
+        return model_dict
```
ArgoJ, line 1016: this looks like the format of the model json then is different right? So its not anymore the model json with all the properties in a list or string format but rather a section serialized expressions and expression names with entrys in it?

interfaces/acados_template/acados_template/acados_model.py
```diff
@@ -964,5 +965,79 @@ def reformulate_with_polynomial_control(self, degree: int) -> None:
...
+
+
+    @classmethod
+    def from_dict(self, model_dict: dict) -> 'AcadosModel':
```
ArgoJ: due to convention maybe change the self to `cls`

interfaces/acados_template/acados_template/acados_model.py
```diff
@@ -964,5 +965,79 @@ def reformulate_with_polynomial_control(self, degree: int) -> None:
 
     def _has_custom_hess(self) -> bool:
         return not (is_empty(self.cost_expr_ext_cost_custom_hess_0) and
-                    is_empty(self.cost_expr_ext_cost_custom_hess) and 
+                    is_empty(self.cost_expr_ext_cost_custom_hess) and
                     is_empty(self.cost_expr_ext_cost_custom_hess_e))
+
+
+    def serialize(self) -> Tuple[str, List[str]]:
```
ArgoJ, line 972: Move this function to a utils function, that any class instance that include casadi expressions can potentially use it? Ignore this coment, if it is intended that only the model contains casadi expressions

interfaces/acados_template/acados_template/acados_model.py
```diff
@@ -964,5 +965,79 @@ def reformulate_with_polynomial_control(self, degree: int) -> None:
...
+        for k, _ in inspect.getmembers(type(self), lambda v: isinstance(v, property)):
+            v = getattr(self, k)
+            if isinstance(v, (ca.SX, ca.MX)):
+                serializer.pack(v)
+                expression_names.append(k)
+
+        return serializer.encode(), expression_names
+
+
+    def deserialize(self, s: str, expression_names: list) -> None:
```
ArgoJ, line 989: same here as with serialize

interfaces/acados_template/acados_template/acados_model.py
FreyJo: yes. This is because it is not possible to serialize individual casadi expressions.

.github/workflows/full_build.yml
```diff
@@ -129,6 +129,7 @@ jobs:
         source ${{ github.workspace }}/acadosenv/bin/activate
         cd ${{ github.workspace }}/examples/acados_python/tests
         python test_rti_sqp_residuals.py
+        python test_serialize.py
```
ArgoJ, line 132: If possible and usefull in your experience maybe move this to the CMakeLists.txt in the folder interfaces, [here](https://github.com/acados/acados/blob/main/interfaces/CMakeLists.txt)

interfaces/acados_template/acados_template/acados_model.py
ArgoJ: i did it but the code looks messy af. But this makes it cleaner anyways also in the json. The upside is that we then don't need to serialize the expressions but can make them human readable again to control it again. So this makes it way cleaner. I can add a PR that reverts this serialization change in the utils back to __str__. This way this is really and entirely seperated, at least if I understand this correctly

ArgoJ:
oh yes this looks so much cleaner damn :D
I haven't tested it, but I think it looks really good.

ArgoJ:
btw when you merged this into main, I will change my from json approach of the other classes reagarding to this approach here.

interfaces/acados_template/acados_template/acados_model.py
sandmaennchen: this is indeed intentional, as only the model contains casadi expressions after calling make_consistent

.github/workflows/full_build.yml
sandmaennchen: cleanup of the tests is on the way, will likely get rid of CMakeLists.txt

FreyJo: Done in https://github.com/acados/acados/pull/1643

PR #1640, opened 2025-09-23
`AcadosOcpSolver.get_residuals()`: change default to `recompute=False`, fixes AS-RTI, breaking!

FreyJo:
- add warning for AS-RTI, as recomputing residual breaks the reuse of the linearization which is essential for AS-RTI to work correctly.
- test AS-RTI with `get_residuals()`

### DIFF
--- a/interfaces/acados_template/acados_template/acados_ocp_solver.py
+++ b/interfaces/acados_template/acados_template/acados_ocp_solver.py
@@ -35,6 +35,7 @@ import os
 import shutil
 import sys
 import time
+import warnings
 
 from ctypes import (POINTER, byref, c_char_p, c_double, c_int, c_bool,
                     c_void_p, cast)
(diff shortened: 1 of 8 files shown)


interfaces/acados_template/acados_template/acados_ocp_solver.py
```diff
@@ -2359,6 +2367,24 @@ def options_set(self, field_, value_):
         return
 
 
+    def options_get(self, field_):
```
sandmaennchen: Add docstring

interfaces/acados_template/acados_template/acados_ocp_solver.py
```diff
@@ -1709,14 +1710,21 @@ def get_residuals(self, recompute=False):
         Returns an array of the form [res_stat, res_eq, res_ineq, res_comp].
         The residuals has to be computed for SQP_RTI solver, since it is not available by default.
 
+        NOTE: For AS-RTI, recomputing the residual is not supported!
```
sandmaennchen: Not sure if I like the behaviour of this function. Wouldn't it be more intuitive if we would not re-evaluate the residuals for `recompute=False` even for standard RTI?

For sure the docstring could be improved

sandmaennchen:
and how exactly does recomputing the residuals break AS-RTI?

PR #1461, opened 2025-03-12
Python: `detect_cost_type`

josipkh:
This PR adds the option of defining the cost type as `AUTO` also in the Python interface, similar to [the MATLAB implementation](https://github.com/acados/acados/blob/main/interfaces/acados_matlab_octave/detect_cost_type.m).
This enables an automatic translation of a generic cost into a linear least-squares form (if possible).

PR #1647, opened 2025-09-29
Update `detect_cost_type`

josipkh:
A small follow-up on #1461, allowing for states defined as casadi.MX and adding checks for global parameters in the MATLAB interface.


josipkh:
@sandmaennchen thanks for the correction!
do you maybe have any idea why the test is failing?
I ran the same example locally and it seems to work fine...

FreyJo:
The failing test is unrelated. It also failed on https://github.com/acados/acados/pull/1634

josipkh:
I see, thanks for the confirmation :)

PR #1664, opened 2025-10-13
Control horizon utility function

Alefunk:
Added a utility function to create a MPOCP with control horizon from a regular OCP. The function takes the original OCP `ocp` and the number of shooting nodes in the control horizon `Nc` as function arguments. The remaining number of shooting nodes without control can be reduced by setting the optional `Nr` argument to a smaller value. They will be distributed equally on the remaining horizon. If you choose a small value, it might be necessary to increase the number of integrator steps of the last phase to keep the same integrator accuracy. If one provides `Nc = N_horizon`, the original OCP will be returned. In many cases, the returned MPOCP can be used as in-place replacement for the original OCP.

The control horizon MPOCP is generated from the original OCP, a discrete 1-step transition OCP and an augmented OCP, where the state vector is extended by the controls, which are constant (`u_dot = 0`) on the remaining horizon.

First tests have shown that the execution time can be improved by using the condensed formulation with a small number of shooting nodes in the control horizon. Using the sparse formulation is not recommended as we add more states on the remaining horizon. Highly nonlinear or inherently unstable systems might suffer from control performance loss when using small control horizons, which can also lead to worse computing performance and longer execution times (more SQP steps necessary to solve).

If you use this function, remember to take care of the discrete 1-step transition phase, when e.g. setting the reference on the horizon. Instead of iterating over the whole range, you need to leave out the transition stage at `k = Nc`:
```
for k in [i for i in range(N) if i != Nc]:
    ocp_solver.cost_set(k, 'yref', yref)
ocp_solver.cost_set(N, 'yref', yref_e)
```

This PR is in draft stage as it is lacking an example.


FreyJo:
Hi Alex, the PR looks already very good.
Can you create a corresponding example soon?
No hurry though.

Alefunk:
Hey Jonathan, sorry for the delay. I created an example which compares the execution times of an ocp with and without control horizon. I used the already present double integrator example from the mocp_transition_example.

The generated plot compares the traditional uncondensed approach (N=Nc) and compares it to a fully condensed ocp with control horizon Nc = 5 (instead of 20). I didnt change any other parameters. Tracking performance is almost the same (but degrades for even smaller control horizons). 

It seems like when using full condensing for both ocps the difference is almost negligible but that probably depends on the model complexity.

grafik

examples/acados_python/mocp_transition_example/control_horizon_example.py
```diff
@@ -0,0 +1,155 @@
...
+    # Simulate closed loop (Control Horizon < Prediction Horizon)
+    for i in range(Nsim):
+        simU[i,:] = mocp_solver.solve_for_x0(simX[i, :])
+        time_tot[i] = 1e3 * mocp_solver.get_stats('time_tot')
+        simX[i+1, :] = sim_solver.simulate(simX[i, :], simU[i,:])
+
+    # Plot final results
+    axs[0].plot(simT, simX[:-1,0], label='Nc < N', color='tab:blue')
+    axs[1].plot(simT, simX[:-1,1], label='Nc < N', color='tab:blue')
+    axs[2].plot(simT, simU, label='Nc < N', color='tab:blue')
```
FreyJo: could you make a steps/stairs plot for the control please? See the other example in this folder.

interfaces/acados_template/acados_template/mpc_utils.py
```diff
@@ -664,3 +665,90 @@ def J_to_idx(J):
                 f'J_to_idx: J matrices can only contain 1s, got J({i + 1}, {this_idx}) = {J[i, this_idx]}')
         idx.append(this_idx[0])
     return np.array(idx)
+
+
+def create_ocp_with_control_horizon(ocp: AcadosOcp, Nc: int, Nr: int=None) -> Union[AcadosOcp, AcadosMultiphaseOcp]:
```
FreyJo, line 665: I think making `Nr` non-optional would remove ambiguity and simply the code.
Alternatively, just input `Nc` and `Np`?

interfaces/acados_template/acados_template/mpc_utils.py
```diff
@@ -664,3 +665,90 @@ def J_to_idx(J):
...
+    If Nr == 0 an AcadosOcp is returned.
+
+    Otherwise, an AcadosMultiphaseOcp is returned.
+
+    Note: An intermediate stage is added, which has no explicit reference set.
+    The control value that is applied on the remaining horizon is held constant and is a decision variable on the transition stage.
+    To set reference values, iterate over the horizon using :code:`rng = [i for i in range(N) if i != Nc]`.
+
+    :param ocp: regular ocp that will be extended with an control horizon
+    :param Nc: number of shooting nodes in control horizon
```
FreyJo, line 678: In the current implementation `Nc+1` is the number of controls.
`Nc` discrete control live in the first phase, 1 lives on the transition stage.

I think we should assert `Nc > 1` and have `Nc-1`  stages in the first phase, then we have `Nc` discrete controls as expected.
Alternatively document it here.

interfaces/acados_template/acados_template/mpc_utils.py
```diff
@@ -664,3 +665,90 @@ def J_to_idx(J):
...
+    aug_ocp.model.x = ca.vertcat(ocp.model.x, ocp.model.u)
+    aug_ocp.model.u = ca_symbol('u_aug', 0, 0)
+    if ocp.solver_options.integrator_type == 'ERK':
+        aug_ocp.model.f_expl_expr = ca.vertcat(ocp.model.f_expl_expr, ca.SX.zeros(nu))
+    elif ocp.solver_options.integrator_type == 'IRK':
+        udot = ca_symbol('udot', nu)
+        aug_ocp.model.xdot = ca.vertcat(ocp.model.xdot, udot)
+        aug_ocp.model.f_impl_expr = ca.vertcat(ocp.model.f_impl_expr, udot)
+    elif ocp.solver_options.integrator_type == 'DISCRETE':
+        aug_ocp.model.disc_dyn_expr = ca.vertcat(ocp.model.disc_dyn_expr, ocp.model.u)
```
FreyJo, line 719: I pushed some fixes for the non ERK case. Hope this is correct. Maybe an additional test would be good. Or at least review my code please :)

interfaces/acados_template/acados_template/mpc_utils.py
```diff
@@ -664,3 +665,90 @@ def J_to_idx(J):
...
+    aug_ocp.constraints.ubx_0 = np.array([])
+    # remove input constraints, not relevant in last phase:
+    aug_ocp.constraints.lbu = np.array([])
+    aug_ocp.constraints.ubu = np.array([])
+    aug_ocp.constraints.idxbu = np.array([])
+
+    # create MOCP
+    mocp = AcadosMultiphaseOcp(N_list=[Nc, 1, Nr])
+    mocp.solver_options = ocp.solver_options
+    mocp.solver_options.time_steps = np.array(Nc * [Ts] + [0] + Nr * [Ts*(Np-Nc)/Nr])
```
FreyJo: If we change the API of this function as above, this makes more sense, as (Np-Nc) == Nr.

FreyJo (commented):
Hi Alex,
thanks for the update!
I pushed some small corrections and reviewed the code carefully.
Could you check the comments please?
Regarding the example, I think it would be good to have the value of `Nc` explicitly in the plot, and to add one variant where the performance degrades.
Otherwise, very good!

interfaces/acados_template/acados_template/mpc_utils.py
Alefunk: I believe the concept of `Nr` (remaining shooting nodes in prediction horizon) is not widely known in conjunction with using a control horizon. That's why I kept it as an optional extension for the users who want to dig a little deeper. Nonetheless, I can change it to non-optional of course if you wish.

`Np` is already part of the original ocp formulation and would cause confusion if required here again I would say.

interfaces/acados_template/acados_template/mpc_utils.py
Alefunk: Good catch! I added an edge case for `Nc=1` and removed the additional control. To keep consistency with the time steps and total number of shooting nodes, I added one additional shooting node on the remaining horizon.

interfaces/acados_template/acados_template/mpc_utils.py
Alefunk: Thanks, I checked the code and tested the IRK case in the example and it works fine.

I just had the problem that the dimension consistency checker is now throwing an assertion if the user is for example defining f_expl_expr although using IRK which I would suspect to happen often (because often one defines all the variables in case you want to switch quickly in between the different integrator settings).

I thus would suggest to bring the `is_empty` checks back again.

FreyJo: yes, checking `is_empty` and augmenting all kinds of dynamics provided is a good idea then.
Please do that!

interfaces/acados_template/acados_template/mpc_utils.py
FreyJo: If you don't see a reason to have `Nr != (Np - Nc)`, I would really like to avoid the additional logic. `Np` can still be taken from `N_horizon` and one just additionally inputs `Nc`.

FreyJo (approved):

PR #1674, opened 2025-10-28
rm `coding: future_fstrings`

lalten:
https://github.com/acados/acados/pull/802 originally introduced `# -*- coding: future_fstrings -*-` to support Python 3.5

Python 3.5 has an EOL date of 2020-09-30 (https://devguide.python.org/versions/)

Acados requires Python >= 3.8 so `future_fstrings` really isn't needed anymore.
https://github.com/acados/acados/blob/5847c74fe774137a5fb31c9791bf39aa980b8b05/interfaces/acados_template/setup.py#L37

The line causes `SyntaxError: unknown encoding: future_fstrings` in my setup (Python 3.10) so it would be nice to just remove it


sandmaennchen:
I'm fine with removing the declaration, but I don't see why this should raise a syntax error. Do you have `future_fstrings` installed? I'm on python 3.10.12 and don't have any issues

lalten:
> I'm fine with removing the declaration, but I don't see why this should raise a syntax error. Do you have `future_fstrings` installed? I'm on python 3.10.12 and don't have any issues

No, I don't have it installed. I checked and that also does fix the issue, but I think this is the better solution.

PR #1254, opened 2024-09-19
Move Globalization functionality into modules, rename related options

david0oo:
This PR introduces a new globalization module. This means that globalization of an algorithm is now handled within the new module. There are three different globalization strategies available at this point: `FIXED_STEP`, `MERIT_BACKTRACKING`, `FUNNEL_L1PEN_LINESEARCH`. All algorithms use either a fixed step size or a line search.

The PR includes the following features:
- options that are related to the globalization have a new prefix `globalization_`. These options include:
    - `alpha_min`  --> `globalization_alpha_min` 
    - `alpha_reduction`  --> `globalization_alpha_reduction` 
    - `use_SOC`  --> `globalization_use_SOC` 
    - `line_search_use_sufficient_descent`  --> `globalization_line_search_use_sufficient_descent`
    - `full_step_dual`  --> `globalization_full_step_dual`
    - `eps_sufficient_descent`  --> `globalization_eps_sufficient_descent`
    - `nlp_solver_step_length` -> `globalization_fixed_step_length`
- All options related to the funnel globalization also got a `globalization_` prefix
- All globalization strategies have a function 'find_acceptable_iterate' that performs the globalization and then also does the step update
- The printing is handled within the globalization as well since for different globalization strategies we would like to print different stats

PR #1679, opened 2025-11-04
Convexify: fix correct_dual_sol to not restore original data

FreyJo:
Plus: fix `ext_qp_res` in RTI, broken in https://github.com/acados/acados/pull/1254

### DIFF
--- a/examples/acados_python/pendulum_on_cart/ocp/example_hessian_regularization.py
+++ b/examples/acados_python/pendulum_on_cart/ocp/example_hessian_regularization.py
@@ -5,28 +5,6 @@ import numpy as np
 from acados_template import AcadosOcp, AcadosOcpSolver
 from pendulum_model import export_pendulum_ode_model
 
-def get_iterate_hessian(ocp_solver: AcadosOcpSolver, N) -> list:
-    """
-    Obtain exact Hessian from the current iterate of the ocp_solver of every stage
-    and form the block diagonal Hessian matrix.
-    """
-    exact_Hessian_list = []
-    for i in range(N+1):
-        hess_block = ocp_solver.get_hessian_block(i)
-        exact_Hessian_list.append(hess_block)
-    return exact_Hessian_list
-
-def check_positive_definite(whole_Hessian, N):
-    '''
-    check each stage Hessian positive definiteness of whole matrix by cholosky decomposition
-    '''
-    for i in range(N+1):
-        stage_Hessian = whole_Hessian[i]
-        try:
-            np.linalg.cholesky(stage_Hessian)
-        except np.linalg.LinAlgError:
-            assert False, f"Stage {i} Hessian is not positive definite. {stage_Hessian=}"
-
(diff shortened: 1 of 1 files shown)


Pandatheon:
For the test, the expected test result is shown below:
图片
However, the current output is as follows:
图片
where the `res_stat` does not align with the expected test result.

FreyJo:
Just looked into `ocp_nlp_sqp`, and external QP residual evaluation (`ocp_qp_res_compute`) happens after the "correct dual" part.
The purpose of `ext_qp_res` is to be able to check, if the QP solver does the job of solving the QP to the required tolerance.
The QP it solves is the fully processed one, i.e. regularized and scaled, so I think we should change this behavior of `ext_qp_res`.
Additional functionality for assessing the QP residual of 1) the scaled QP 2) the non-regularized QP could be implemented additionally.

I suggest:
1) making a separate PR and change the behavior of `ext_qp_res` to regard the fully processed QP.
2) adding/modifying a test here which shows what the PR fixed, i.e. a positive definite Hessian is obtained in Python.

FreyJo:
QP residual fix already merged in #1681

examples/acados_python/pendulum_on_cart/ocp/example_hessian_regularization.py
```diff
@@ -5,6 +5,28 @@
 from acados_template import AcadosOcp, AcadosOcpSolver
 from pendulum_model import export_pendulum_ode_model
 
+def get_iterate_hessian(ocp_solver: AcadosOcpSolver, N) -> list:
+    """
+    Obtain exact Hessian from the current iterate of the ocp_solver of every stage
```
sandmaennchen: The doc string is a bit confusing. It's not neccessarily the exact hessian and WE also do not return the full block diagonal hessian

examples/acados_python/pendulum_on_cart/ocp/example_hessian_regularization.py
```diff
@@ -5,6 +5,28 @@
...
+    and form the block diagonal Hessian matrix.
+    """
+    exact_Hessian_list = []
+    for i in range(N+1):
+        hess_block = ocp_solver.get_hessian_block(i)
+        exact_Hessian_list.append(hess_block)
+    return exact_Hessian_list
+
+def check_positive_definite(whole_Hessian, N):
```
sandmaennchen: Ist this needed? Could ne added to qp diagnostics?

sandmaennchen (approved):

examples/acados_python/pendulum_on_cart/ocp/example_hessian_regularization.py
FreyJo: Better just use qp diagnostic and check the eigenvalues obtained there IMO

Pandatheon:
Basically just change the `nlp_solver_type` in the original test from SQP to SQP_RTI

PR #1705, opened 2025-11-21
Add option `anderson_activation_threshold`

FreyJo:
The option allows to not use Anderson accelerations while the infinity norm of the current KKT residual is above `anderson_activation_threshold`.
This allows to mitigate slower convergence of the Anderson accelerated variant in an initial regime as show-cased in the example in https://github.com/acados/acados/pull/1703
The option can be set on the fly in MATLAB and Python.
[Pollock2021] suggests using values in the order of 1.

### Plots with the new option:
convergence__GAUSS_NEWTON_1e-08_furuta
convergence_slack_EXACT_0 1_furuta
convergence_slack_EXACT_1e-08_furuta
convergence_slack_GAUSS_NEWTON_1e-08_furuta

### Documentation of `anderson_activation_threshold`
```
Only relevant if with_anderson_acceleration == True.
Anderson accelerations are performed whenever the infinity norm of the KKT residual is < `anderson_activation_threshold `.

If the KKT residual norm is larger than `anderson_activation_threshold `, no Anderson acceleration is performed.

In the language of [Pollock2021, Sec. 5.1], this corresponds to specifying an "initial regime", consisting of iterates with KKT residual norm > `anderson_activation_threshold ` and an (pre-)asymptotic regime, where the residual norm is <= `anderson_activation_threshold`.
In the initial regime, no Anderson acceleration is performed, i.e. depth $m=0$.
In the (pre-)asymptotic regime, Anderson acceleration with depth $m=1$ is performed.

Type: float
Default: 1e1
```

### Reference
[Pollock2021] Anderson acceleration for contractive and noncontractive operators, Sara Pollock, IMA Journal of Numerical Analysis, 2021

### DIFF
--- a/interfaces/acados_template/acados_template/c_templates_tera/matlab_templates/acados_mex_set.in.c
+++ b/interfaces/acados_template/acados_template/c_templates_tera/matlab_templates/acados_mex_set.in.c
@@ -669,6 +669,13 @@ void mexFunction(int nlhs, mxArray *plhs[], int nrhs, const mxArray *prhs[])
         double qp_mu0 = (double) value[0];
         ocp_nlp_solver_opts_set(config, opts, "qp_mu0", &qp_mu0);
     }
+    else if (!strcmp(field, "anderson_activation_threshold"))
+    {
+        acados_size = 1;
+        MEX_DIM_CHECK_VEC(fun_name, field, matlab_size, acados_size);
+        double anderson_activation_threshold = (double) value[0];
+        ocp_nlp_solver_opts_set(config, opts, "anderson_activation_threshold", &anderson_activation_threshold);
+    }
     else if (!strcmp(field, "qp_print_level"))
     {
         if (!(plan->ocp_qp_solver_plan.qp_solver == FULL_CONDENSING_HPIPM || plan->ocp_qp_solver_plan.qp_solver == PARTIAL_CONDENSING_HPIPM))


interfaces/acados_template/acados_template/acados_ocp_options.py
```diff
@@ -786,14 +787,33 @@ def as_rti_iter(self):
     @property
     def with_anderson_acceleration(self):
         """
-        Determines if Anderson acceleration is performed.
+        Determines if algorithm uses Anderson accelerations.
+        Only depth one is supprted.
```
sandmaennchen: typo

acados/ocp_nlp/ocp_nlp_globalization_fixed_step.c
```diff
@@ -166,6 +166,7 @@ int ocp_nlp_globalization_fixed_step_find_acceptable_iterate(void *nlp_config_,
     ocp_nlp_workspace *nlp_work = nlp_work_;
     ocp_nlp_opts *nlp_opts = nlp_opts_;
     ocp_nlp_globalization_fixed_step_opts *opts = nlp_opts->globalization;
+    // ocp_nlp_globalization_fixed_step_memory *glob_mem = nlp_mem->globalization;
```
david0oo, line 169: Not sure why this is added?

FreyJo: That was from a previous implementation, where I kept doing Anderson accelerations if the iterate once goes below the threshold.
I think the comment doesn't hurt here.

interfaces/acados_template/acados_template/acados_ocp_options.py
```diff
@@ -786,14 +787,33 @@ def as_rti_iter(self):
...
 
+    @property
+    def anderson_activation_threshold(self):
+        """
+        Only relevant if with_anderson_acceleration == True.
+        Anderson accelerations are performed whenever the infinity norm of the KKT residual is < `anderson_activation_threshold `.
+
+        If the KKT residual norm is larger than `anderson_activation_threshold `, no Anderson acceleration is performed.
+
+        In the language of [Pollock2021, Sec. 5.1], this corresponds to specifying an "initial regime", consisting of iterates with KKT residual norm > `anderson_activation_threshold ` and an (pre-)asymptotic regime, where the residual norm is <= `anderson_activation_threshold`.
```
sandmaennchen: Maybe add full reference

PR #1709, opened 2025-11-25
ERK optional forward parameter sensitivities `S_p` (end-to-end)

ivandoornen:
## Summary

This PR adds optional forward parameter sensitivities \(S_p\) to the simulation layer, with ERK support for propagating them, exposes per-stage parameter sensitivities at the OCP level, and makes the feature usable from the MATLAB interface and generated code.

For an integrator step
$x_{n+1} = f(x_n, u_n, p_n),$
we define the forward parameter sensitivities
$S_p = \frac{\partial x_{n+1}}{\partial p}, \quad S_p \in \mathbb{R}^{n_x \times n_{p}},$

---

### Common sim API (`acados/sim/sim_common.[ch]`)
- Add optional parameter inputs to `sim_in`:
  - `p` (stagewise, length `np`)
- Add `S_p` to `sim_out` with layout `[nx x np]` (column-major).
- Add `sens_forw_p` option to `sim_opts` (default: `false`).

### ERK integrator (`acados/sim/sim_erk_integrator.[ch]`)
- Add `np` to ERK dims.
- Add a new external function hook: `expl_vde_forw_p` for parameter-sensitivity propagation.
- When `sens_forw && sens_forw_p && np> 0`:
  - propagate `S_p` across stages via `expl_vde_forw_p`
  - write results to `sim_out->S_p`

### IRK / lifted-IRK / GNSF compatibility
- Update dims/plumbing in `sim_irk_integrator`, `sim_lifted_irk_integrator`, and `sim_gnsf` so they remain compatible with the extended sim API.
- These integrators do not compute `S_p` yet.

### OCP dynamics (`acados/ocp_nlp/ocp_nlp_dynamics_cont.c/h`)
- Extend the continuous dynamics memory with a BLASFEO matrix `S_p` of size `[nx_{k+1} × np]`.
- Implement `dims_get(..., "np", ...)` to return `np` based on the underlying sim dims (`np`).
- In the dynamics evaluation callback:
  - If parameter sensitivities are enabled in the simulator options, pack `sim_out->S_p` into the cached BLASFEO `S_p`.
  - Otherwise, zero `S_p` to keep memory contents deterministic.
- Provide `memory_get(..., "S_p", ...)` to export the cached `S_p(k)` to a dense buffer.

### C interface (`interfaces/acados_c/ocp_nlp_interface.c`)
- Add a dimension query for `"S_p"`:
  - `ocp_nlp_qp_dims_get_from_attr(..., "S_p", dims_out) -> [nx_{k+1}, np]`
- Add a getter for `"S_p"`:
  - `ocp_nlp_get_at_stage(solver, k, "S_p", S_p_out)` for stages `k = 0..N-1`
  - Terminal stage is handled explicitly (warning / empty).

---

### examples/acados_matlab_octave/test/test_sens_forw_p.m (new)
- Adds a finite-difference validation test for ERK parameter sensitivities:
  - enables `sens_forw_p='true'` and compares `S_p` against finite differences
  - also checks `Sx` / `Su` FD consistency

---

### interfaces/acados_matlab_octave/AcadosOcpOptions.m
- Adds option property `sens_forw_p` (default `false`) so OCP creation can request param sensitivities in dynamics.

### interfaces/acados_matlab_octave/AcadosSimOptions.m
- Adds `sens_forw_p` (default `false`).

### interfaces/acados_matlab_octave/AcadosSim.m
- Extends option validation: `opts.sens_forw_p` must be logical.

### interfaces/acados_matlab_octave/acados_sim_opts.m (legacy sim interface)
- Adds legacy string option `sens_forw_p`.

### interfaces/acados_matlab_octave/setup_AcadosSim_from_legacy_sim_description.m
- When upgrading legacy sim descriptions, forwards `sens_forw_p` if present (else sets to `false`).

---

### interfaces/acados_matlab_octave/generate_c_code_explicit_ode.m
- Selects the effective parameter symbol:
  - use `model.p` if present, else an empty CasADi vector (typed SX/MX).
- If `np > 0`, creates a new CasADi function:
  - `{{model.name}}_expl_vde_forw_p`
  - computes `vdeP = f_p + A * Sp` (with `A = ∂f/∂x`, `f_p = ∂f/∂p`)
  - signature uses inputs `{x, Sp, u, p}` and outputs `{vdeP}`.
  - This is the function the ERK integrator calls when `sens_forw_p` is enabled.

---

### interfaces/acados_matlab_octave/sim_create_ext_fun.c
- Extends the external function struct with `dyn_expl_vde_forw_p` pointer storage.

### interfaces/acados_matlab_octave/sim_get.c
- Reads `np` from `sim_dims`.
- Adds `sim.get('S_p')` returning an `[nx x np]` MATLAB matrix.

### interfaces/acados_matlab_octave/ocp_get.c
- Adds `ocp.get('S_p')`:
  - `ocp.get('S_p')` returns a `1xN` cell array, each cell `[nx_{k+1} x np]`
  - `ocp.get('S_p', k)` returns a single `[nx_{k+1} x np]` matrix
- Layout note: returned matrices are MATLAB-native (column-major, leading dimension `nx_{k+1}`).

---

### interfaces/acados_template/.../model.in.h
- Declares prototypes for `*_expl_vde_forw_p` (guarded by `np>0`).

### interfaces/acados_template/.../acados_sim_solver.in.h / .in.c
- Adds a capsule pointer `sim_expl_vde_forw_p`.
- Allocates/maps `sim_expl_vde_forw_p` when parameters exist.
- Registers external function with the sim model via `model_set("expl_vde_forw_p", ...)`.
- Sets sim option `"sens_forw_p"` only if user enabled it and `np>0`.

### interfaces/acados_template/.../acados_solver.in.h / .in.c
- Adds `expl_vde_forw_p` array (ERK stages) when parameters exist.
- Registers it into the dynamics via `ocp_nlp_dynamics_model_set_external_param_fun(...,"expl_vde_forw_p",...)`.
- Sets solver option `"dynamics_sens_forw_p"` at all stages when enabled (and parameters exist).
- Frees the allocated `expl_vde_forw_p` array during teardown.

---

## Behavior / compatibility

- Off by default: `sens_forw_p = false`.
- `S_p` is only propagated when `sens_forw && sens_forw_p && np> 0`.
- The additional `S_p` workspace in ERK is only used when the above condition holds.
- If `np = 0` or `sens_forw_p = false` , existing behavior is unchanged.

External function requirement:
- If `sens_forw_p` is enabled (and `sens_forw` is enabled) with `np> 0`, the ERK requires an externally supplied `expl_vde_forw_p`.

OCP exposure:
- If sensitivities are enabled, `S_p(k)` is cached in the OCP dynamics memory and can be retrieved via the new `"S_p"` getter.
- If sensitivities are not enabled, `S_p(k)` is returned as zero (deterministic memory contents).
- `S_p_out` is returned as a dense matrix with leading dimension `nx_{k+1}` (i.e. column-major).

Backward compatibility:
- Default is unchanged (`sens_forw_p = false` everywhere).
- No additional memory or external functions are used unless `sens_forw_p` is explicitly enabled.
- When parameters are absent, generated code auto-disables `sens_forw_p` even if requested.

---

## Follow-up plan (roles + dependencies)

1. zoRO / Simulink integration
   - Wire `S_p` into the zoRO workflow (tightening/backoff terms that depend on parameter uncertainty).
   - Update custom update templates and Simulink S-function integration / payload streaming as needed.

### DIFF
--- a/acados/ocp_nlp/ocp_nlp_dynamics_cont.c
+++ b/acados/ocp_nlp/ocp_nlp_dynamics_cont.c
@@ -165,7 +165,7 @@ void ocp_nlp_dynamics_cont_dims_set(void *config_, void *dims_, const char *fiel
     {
         ocp_nlp_dynamics_cont_set_nu1(config_, dims_, value);
     }
-	else if (!strcmp(field, "np"))
+    else if (!strcmp(field, "np"))
     {
         ocp_nlp_dynamics_cont_set_np(config_, dims_, value);
     }
@@ -435,7 +435,7 @@ void *ocp_nlp_dynamics_cont_memory_assign(void *config_, void *dims_, void *opts
 
     // fun
     assign_and_advance_blasfeo_dvec_mem(nx1, &memory->fun, &c_ptr);
-   
+
     assert((char *) raw_memory +
                ocp_nlp_dynamics_cont_memory_calculate_size(config_, dims, opts_) >=
            c_ptr);
@@ -592,10 +592,10 @@ void ocp_nlp_dynamics_cont_memory_get(void *config_, void *dims_, void *mem_, co
     {
         sim->memory_get(sim, dims->sim, mem->sim_solver, field, value);
     }
-	else if (!strcmp(field, "S_p"))
-	{
-		sim->memory_get(sim, dims->sim, mem->sim_solver, "S_p", value);
-	}
+    else if (!strcmp(field, "S_p"))
+    {
+        sim->memory_get(sim, dims->sim, mem->sim_solver, "S_p", value);
+    }
     else
     {
         printf("\nerror: ocp_nlp_dynamics_cont_memory_get: field %s not available\n", field);
--- a/acados/sim/sim_erk_integrator.c
+++ b/acados/sim/sim_erk_integrator.c
@@ -97,9 +97,9 @@ void sim_erk_dims_set(void *config_, void *dims_, const char *field, const int *
     {
         dims->np = *value;
     }
-	else if (!strcmp(field, "np_global"))
+    else if (!strcmp(field, "np_global"))
     {
-		// np_global dimension not needed
+        // np_global dimension not needed
     }
     else
     {
@@ -126,10 +126,10 @@ void sim_erk_dims_get(void *config_, void *dims_, const char *field, int *value)
     {
         *value = 0;
     }
-    else if (!strcmp(field, "np")) 
-	{ 
-		*value = dims->np; 
-	}
+    else if (!strcmp(field, "np"))
+    {
+        *value = dims->np;
+    }
     else
     {
         printf("\nerror: sim_erk_dims_get: dim type not available: %s\n", field);
@@ -165,8 +165,8 @@ void *sim_erk_model_assign(void *config, void *dims, void *raw_memory)
     model->expl_vde_for = NULL;
     model->expl_vde_adj = NULL;
     model->expl_ode_hes = NULL;
-	model->expl_vde_for_p = NULL;
-	
+    model->expl_vde_for_p = NULL;
+
     return model;
 }
 
@@ -192,9 +192,9 @@ int sim_erk_model_set(void *model_, const char *field, void *value)
     {
         model->expl_ode_hes = value;
     }
-	else if (!strcmp(field, "expl_vde_forw_p"))
+    else if (!strcmp(field, "expl_vde_forw_p"))
     {
-        model->expl_vde_for_p = value; 
+        model->expl_vde_for_p = value;
     }
     else
     {
@@ -294,7 +294,7 @@ void sim_erk_opts_initialize_default(void *config_, void *dims_, void *opts_)
     opts->sens_forw = true;
     opts->sens_adj = false;
     opts->sens_hess = false;
-	opts->sens_forw_p = false;
+    opts->sens_forw_p = false;
     opts->cost_computation = false;
 
     opts->output_z = false;
@@ -340,7 +340,7 @@ acados_size_t sim_erk_memory_calculate_size(void *config, void *dims, void *opts
 
     acados_size_t size = sizeof(sim_erk_memory);
 
-    if (opts->sens_forw && opts->sens_forw_p )
+    if ( opts->sens_forw_p )
     {
         size += erk_dims->nx * erk_dims->np * sizeof(double);
     }
@@ -356,13 +356,13 @@ void *sim_erk_memory_assign(void *config, void *dims, void *opts_, void *raw_mem
 
     sim_erk_memory *mem = (sim_erk_memory *) c_ptr;
     c_ptr += sizeof(sim_erk_memory);
-	
+
     mem->S_p = NULL;
 
     sim_erk_dims *erk_dims = (sim_erk_dims *) dims;
     sim_opts *opts = (sim_opts *) opts_;
 
-    if (opts->sens_forw && opts->sens_forw_p )
+    if ( opts->sens_forw_p )
     {
         mem->S_p = (double *) c_ptr;
         c_ptr += erk_dims->nx * erk_dims->np * sizeof(double);
@@ -424,10 +424,10 @@ void sim_erk_memory_get(void *config_, void *dims_, void *mem_, const char *fiel
         sim_erk_dims *dims = (sim_erk_dims *) dims_;
         int nx = dims->nx;
         int np = dims->np;
-		
-		// If there are no parameters, S_p is an empty [nx x 0] matrix.
-		if (np == 0)
-			return;
+
+        // If there are no parameters, S_p is an empty [nx x 0] matrix.
+        if (np == 0)
+            return;
 
         if (mem->S_p == NULL)
         {
@@ -462,11 +462,11 @@ acados_size_t sim_erk_workspace_calculate_size(void *config_, void *dims_, void 
     int nx = dims->nx;
     int nu = dims->nu;
     int nf = opts->num_forw_sens;                  // existing (Sx,Su) cols
-	if (!opts->sens_forw) nf = 0;
-	
-	int np   = dims->np;
-	int nf_p = (opts->sens_forw && opts->sens_forw_p) ? np : 0;
-	
+    if (!opts->sens_forw && !opts->sens_forw_p) nf = 0;
+
+    int np   = dims->np;
+    int nf_p = opts->sens_forw_p ? np : 0;
+
     int nX = nx * (1 + nf + nf_p);  // x + [Sx,Su] + [S_p]
     int nhess = (nf + 1) * nf / 2;
     int num_steps = opts->num_steps;  // number of steps
@@ -511,15 +511,15 @@ static void *sim_erk_cast_workspace(void *config_, sim_erk_dims *dims, sim_opts 
 {
     int ns = opts->ns;
 
-	int nx = dims->nx;
-	int nu = dims->nu;
-	int nf = opts->num_forw_sens;
-	if (!opts->sens_forw) nf = 0;
+    int nx = dims->nx;
+    int nu = dims->nu;
+    int nf = opts->num_forw_sens;
+    if (!opts->sens_forw && !opts->sens_forw_p) nf = 0;
 
-	int np   = dims->np;
-	int nf_p = (opts->sens_forw && opts->sens_forw_p) ? np : 0;
+    int np   = dims->np;
+    int nf_p = opts->sens_forw_p ? np : 0;
 
-	int nX = nx * (1 + nf + nf_p);
+    int nX = nx * (1 + nf + nf_p);
 
     int nhess = (nf + 1) * nf / 2;
     int num_steps = opts->num_steps;  // number of steps
@@ -624,7 +624,7 @@ void sim_erk_set_external_fun_workspaces(void *config_, void *dims_, void *opts_
     external_function_set_fun_workspace_if_defined(model->expl_vde_for, workspace_);
     external_function_set_fun_workspace_if_defined(model->expl_vde_adj, workspace_);
     external_function_set_fun_workspace_if_defined(model->expl_ode_hes, workspace_);
-	external_function_set_fun_workspace_if_defined(model->expl_vde_for_p, workspace_);
+    external_function_set_fun_workspace_if_defined(model->expl_vde_for_p, workspace_);
 }
 
 
@@ -688,31 +688,31 @@ int sim_erk(void *config_, sim_in *in, sim_out *out, void *opts_, void *mem_, vo
         exit(1);
     }
 
-	int nf = opts->num_forw_sens;
-	if (!opts->sens_forw) nf = 0;
+    int nf = opts->num_forw_sens;
+    if (!opts->sens_forw && !opts->sens_forw_p) nf = 0;
 
-	/* ------------------ parameter-sensitivity (S_p) support ------------------ */
-	/* parameter dimension for S_p uses only stagewise np */
-	int np   = dims->np;
-	int nf_p = (opts->sens_forw && opts->sens_forw_p) ? np : 0;
-	
-	if (nf_p > 0 && mem->S_p == NULL)
-	{
-		printf("sim ERK: parameter sensitivities requested but not available in memory.\n");
-		exit(1);
-	}
+    /* ------------------ parameter-sensitivity (S_p) support ------------------ */
+    /* parameter dimension for S_p uses only stagewise np */
+    int np   = dims->np;
+    int nf_p = opts->sens_forw_p ? np : 0;
 
-	/* layout:
-	 * forw_traj has length nX = nx * (1 + nf + nf_p)
-	 * [ x (nx) | S_forw (nx*nf) | S_p (nx*nf_p, optional) ]
-	 * rhs_forw_in additionally stores u at the end: [ ... | u (nu) ]
-	 */
-	const int off_x  = 0;
-	const int off_Sx = nx;          // only meaningful if sens_forw
-	const int off_Su = nx + nx*nx;  // only meaningful if sens_forw
-	const int off_Sp = nx + nx*nf;  // only meaningful if nf_p > 0
-	const int nX     = nx * (1 + nf + nf_p);
-	const int off_u  = nX;
+    if (nf_p > 0 && mem->S_p == NULL)
+    {
+        printf("sim ERK: parameter sensitivities requested but not available in memory.\n");
+        exit(1);
+    }
+
+    /* layout:
+     * forw_traj has length nX = nx * (1 + nf + nf_p)
+     * [ x (nx) | S_forw (nx*nf) | S_p (nx*nf_p, optional) ]
+     * rhs_forw_in additionally stores u at the end: [ ... | u (nu) ]
+     */
+    const int off_x  = 0;
+    const int off_Sx = nx;          // only meaningful if sens_forw
+    const int off_Su = nx + nx*nx;  // only meaningful if sens_forw
+    const int off_Sp = nx + nx*nf;  // only meaningful if nf_p > 0
+    const int nX     = nx * (1 + nf + nf_p);
+    const int off_u  = nX;
 
     int nhess = (nf + 1) * nf / 2;
 
@@ -735,24 +735,24 @@ int sim_erk(void *config_, sim_in *in, sim_out *out, void *opts_, void *mem_, vo
     double *adj_tmp = work->out_adj_tmp;
     double *adj_traj = work->adj_traj;
     double *rhs_adj_in = work->rhs_adj_in;
-	
+
     double *xn = out->xn;
     double *S_forw_out = out->S_forw;
     double *S_adj_out = out->S_adj;
     double *S_hess_out = out->S_hess;
 
-	ext_fun_arg_t ext_fun_type_in[5];
-    void *ext_fun_in[5];					
+    ext_fun_arg_t ext_fun_type_in[5];
+    void *ext_fun_in[5];
     ext_fun_arg_t ext_fun_type_out[3];
     void *ext_fun_out[3];
 
     ext_fun_arg_t expl_vde_type_in[4];
     void *expl_vde_in[4];
-	
+
     ext_fun_arg_t expl_vde_type_out[3];
     void *expl_vde_out[3];
 
-    if (opts->sens_forw)
+    if (opts->sens_forw || opts->sens_forw_p)
     {  // simulation + forward sensitivities
         expl_vde_type_in[0] = COLMAJ;
         expl_vde_in[0] = rhs_forw_in + off_x;   // x: nx
@@ -787,7 +787,7 @@ int sim_erk(void *config_, sim_in *in, sim_out *out, void *opts_, void *mem_, vo
 
     // initialize integrator variables
     for (i = 0; i < nx; i++) forw_traj[off_x + i] = x[i];  // x0
-    if (opts->sens_forw)
+    if (opts->sens_forw || opts->sens_forw_p)
     {
         for (i = 0; i < nx * nf; i++) forw_traj[off_Sx + i] = S_forw_in[i];  // sensitivities
     }
@@ -796,7 +796,7 @@ int sim_erk(void *config_, sim_in *in, sim_out *out, void *opts_, void *mem_, vo
     {
         for (i = 0; i < nx * nf_p; i++)
             forw_traj[off_Sp + i] = 0.0;
-    }	
+    }
     for (i = 0; i < nu; i++) rhs_forw_in[off_u + i] = u[i];  // controls
 
     for (istep = 0; istep < num_steps; istep++)
@@ -825,7 +825,7 @@ int sim_erk(void *config_, sim_in *in, sim_out *out, void *opts_, void *mem_, vo
             }
 
             acados_tic(&timer_ad);
-            if (opts->sens_forw)
+            if (opts->sens_forw || opts->sens_forw_p)
             {  // simulation + forward sensitivities
                 // forward VDE evaluation
                 expl_vde_out[0] = K_traj + s * nX + off_x;   // fun: nx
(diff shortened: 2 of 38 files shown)


ivandoornen:
Quick question on review workflow: I have PR2–PR4 prepared as follow-ups (OCP exposure → MATLAB/templates → zoRO/Simulink). Would you prefer I wait with PR2 until PR1 is approved/merged, or should I open the follow-up PRs already as dependent PRs for parallel review?

ivandoornen:
PR2–PR4 are available for early review on my fork and will be opened here after PR1 is reviewed.

acados/sim/sim_common.c
```diff
@@ -80,14 +80,19 @@ acados_size_t sim_in_calculate_size(void *config_, void *dims)
 
     acados_size_t size = sizeof(sim_in);
 
-    int nx, nu, nz;
+    int nx, nu, nz, np = 0, np_global = 0;
```
FreyJo: Initialization of to 0 not necessary as getter will overwrite it later anyway.
Comments regarding "optional" can be removed, as also "nu" and "nz" can be 0.

acados/sim/sim_erk_integrator.c
```diff
@@ -63,6 +63,8 @@ void *sim_erk_dims_assign(void *config_, void *raw_memory)
     dims->nx = 0;
     dims->nu = 0;
     dims->nz = 0;
+    dims->np = 0;
+    dims->np_global = 0;
```
FreyJo: These initializations should also be done in the other integrators.

ivandoornen:
Hi,  
Thank you for reviewing my code!  

> Somehow, I cannot make the CI tests run on the PR as usual. Did you run them? Otherwise I can try some more, probably pushing as a maintainer of acados helps.

I was also surprised it did not run the CI tests, like it did with my last (now closed) PRs, even though they normally seem to be triggered upon opening a pull request. I thought perhaps a maintainer needed to trigger them, but apparently that is not the case either. I did run `test_all_examples.m` locally, but I do not think I can run or trigger the rest of the CI myself.

> I think that `p_eff` introduces some ambiguity and we should rather just support either p or p_global and later think about how to generalize it to the other one.

I see your point that `p_eff` might be a bit ambiguous. I could either change it to support only `p` or only `p_global`, or define a separate `sens_forw_p` and `sens_forw_p_global` so it is explicit which one is used. Could you let me know which option you would prefer?

> In particular, the function that you need to call (`expl_vde_for_p`) needs to be defined in the interfaces to evaluate the derivative of some specific parameter. Also, this definition should be part of this PR.

I split the series of PRs up into 4, with the first being the sim side, the second the OCP side, the third the MATLAB interfaces and templates, and the last the zoRO and Simulink implementation. In the third PR I modified `generate_c_code_explicit_ode.m` to generate `expl_vde_for_p`. The other 3 PRs are already visible on my fork. Since I implemented the interface side in PR 3, would you prefer that I move those interface and template changes into this PR to keep everything consolidated?

> Do you have a test for evaluating parametric sensitivities of ERK? I think it would be great to add one here.

In the third PR I added a test script that compares the parametric sensitivities of ERK against a finite difference approach for `pendulum_on_cart_model_with_param()` (similar to the existing `test_sens_*m` scripts). Since the test requires the MATLAB interface, I only added it to the third PR.

FreyJo:
> I see your point that p_eff might be a bit ambiguous. I could either change it to support only p or only p_global, or define a separate sens_forw_p and sens_forw_p_global so it is explicit which one is used. Could you let me know which option you would prefer?

Defining it separately would be preferred in my opinion.
Alternatively, implementing only one of the two (the one that you need) would be fine as well.

Regarding the separation of the PRs, I think it would be better to pick the changes form https://github.com/ivandoornen/acados/pull/3 that are not specific to the OCP already here, such that the integrator functionality can already be tested.

But we can also go ahead with the PR structure you have to avoid more git trouble for you.

ivandoornen:
For my project, I don’t necessarily need stage-varying parameters. For now, I’ll modify it so that it supports only `p_global`, and perhaps add support for stagewise `p` in a future update.

Could you clarify whether acados supports using both `p` and `p_global` at the same time? In the original `generate_c_code_explicit_ode.m`, there was only a single parameter vector `p`. Was that intended to represent a concatenation such as `p = [p; p_global]`?

If it’s not too much trouble, I would prefer to continue working within the current PR structure. You can either check out my `PR3` branch (which is built on top of PRs 1 and 2) to verify the evaluation test, or I can open those PRs here for you to review. Let me know what you prefer!

FreyJo:
Yes, acados does support `p` and `p_global` at the same time.
However, `p_global` is added as an input to the function definitions later in the `GenerateContext`.
Another limitation of `p_global` is that they have not been added to the integrator (yet).
Thus, sensitivities with respect to `p_global` cannot be tested with the standalone integrator interface.

FreyJo:
I think it would be best to implement both sensitivities wrt p and p_global separately in ERK to be able to test both.
Alternatively, we need to add a similar test through the OCP solver interface.

Regarding the PR structure: I suggest also adding the changes from your PRs 2 and 3 directly.
This allows to actually test the new functionality when its added and also simplify the work for you.
Right now, I would need to test a checkout a different branch in order to test the added feature.

ivandoornen:
I tried wiring an extra parameter argument into the ERK sim VDE calls, but I’m not entirely sure it’s actually being used. From what I see in the parametric CasADi wrappers, the parameter seems to come from `set_param(...)` / `set_param_pointer(...)`, and the parameter input is skipped during `evaluate()`. So just passing `p_global` in the input list might have no effect unless I’m missing something or I change how the external functions are wrapped. If that’s the right interpretation, I can revert the changes where I pass `p`/`p_global` explicitly and instead add/use a dedicated wrapper for the new `expl_vde_for_p` function.

Given that (and your note that `p_global` isn’t in the integrator yet), I’m a bit worried that proper `p_global` sensitivity support in the standalone integrator may be more involved than I expected. For now I’m leaning towards supporting `S_p` for `p` only, and putting all parameters I want sensitivities for into `p`, while keeping the rest in `p_global`.

ivandoornen:
> Regarding the PR structure: I suggest also adding the changes from your PRs 2 and 3 directly. This allows to actually test the new functionality when its added and also simplify the work for you. Right now, I would need to test a checkout a different branch in order to test the added feature.

Just to confirm: would you prefer me to open PR2 and PR3 here as stacked PRs on top of PR1, or squash them into a single PR so everything needed for testing is in one place?

FreyJo:
> For now I’m leaning towards supporting S_p for p only, and putting all parameters I want sensitivities for into p, while keeping the rest in p_global.

This is fine for me!
Indeed extending integrator to use `p_global` is a bit more involved.
The parameters (or variants of them) are set in the general solver wrapper.
So, yes, easier if you just do p, also for testing.

I meant squashing them into one PR, so you can just apply the commits on the branch of this PR.

ivandoornen:
> I meant squashing them into one PR, so you can just apply the commits on the branch of this PR.

I have now squashed them into this PR. It has become a bit messy, so if you would like I could also create a fresh PR.

ivandoornen:
I guess it will keep failing unless I implement the Python side as well. I can mirror the MATLAB changes into Python and push them here as well? Perhaps I should at this point also change to draft or open a new PR ones everything is implemented.

Also, what is the Python equivalent of test_all_examples.m? I set-up a WSL such that I can test Python side as well before this CI.

FreyJo:
All properties in the problem formulation objects should be the same in MATLAB and Python.
For the solver interactions, you can also implement them in just one of the interfaces.

ivandoornen:
I see my last commit resolved the CI issues. At this point you should have all files in one place, and you can verify that S_p is correctly computed using the test_sens_frow_p.m script

FreyJo:
Amazing! I am a bit busy this week but will review it by Monday at the latest.

ivandoornen:
I plan to use the diff_drive code you have been using in previous zoRO related publications to test my parametric uncertainty approach. Since that is all Python based, I think it is easier for me to add the S_p support to Python as well. Do you prefer adding that to this PR as well, or create a new one?

FreyJo:
> I think it is easier for me to add the S_p support to Python as well. Do you prefer adding that to this PR as well, or create a new one?

You can also add it here!

acados/ocp_nlp/ocp_nlp_dynamics_cont.c
```diff
@@ -379,6 +394,13 @@ acados_size_t ocp_nlp_dynamics_cont_memory_calculate_size(void *config_, void *d
     size +=
         config->sim_solver->memory_calculate_size(config->sim_solver, dims->sim, opts->sim_solver);
 
+    // buffer for parameter sensitivities S_p (nx1 x np_eff)
```
FreyJo: Please replace `np_eff` with `np` everywhere.

acados/ocp_nlp/ocp_nlp_dynamics_cont.c
```diff
@@ -577,6 +609,18 @@ void ocp_nlp_dynamics_cont_memory_get(void *config_, void *dims_, void *mem_, co
     {
         sim->memory_get(sim, dims->sim, mem->sim_solver, field, value);
     }
+	else if (!strcmp(field, "S_p"))
+	{
+		int np = 0;
+		config->sim_solver->dims_get(config->sim_solver, dims->sim, "np", &np);
+		if (np > 0)
+		{
+			// layout: row-major double[nx1 * np_eff]
```
FreyJo: I think it is column-major, also compare your comment below.

acados/sim/sim_irk_integrator.h
```diff
@@ -46,7 +46,7 @@ typedef struct
     int nx;
     int nu;
     int nz;
-
+    int np; // total number of stage wise parameters
```
FreyJo: I find this comment is rather confusing, rather remove it.

acados/sim/sim_lifted_irk_integrator.h
```diff
@@ -44,6 +44,7 @@ typedef struct
     int nx;
     int nu;
     int nz;
+    int np; // total number of stage wise parameters
```
FreyJo: I find this comment is rather confusing, rather remove it.

interfaces/acados_template/acados_template/acados_ocp_solver_pyx.pyx
FreyJo, line 1: Was not really necessary to add it also in Cython, but thanks!

acados/sim/sim_common.c
```diff
@@ -135,6 +138,7 @@ sim_in *sim_in_assign(void *config_, void *dims, void *raw_memory)
     // assign doubles
     assign_and_advance_double(nx, &in->x, &c_ptr);
     assign_and_advance_double(nu, &in->u, &c_ptr);
+    assign_and_advance_double(np, &in->p, &c_ptr);
```
FreyJo: Adding `in->p` is actually not needed as far as I see.
Am I missing something?
Otherwise, please remove it.

acados/sim/sim_common.c
```diff
@@ -275,16 +289,18 @@ acados_size_t sim_out_calculate_size(void *config_, void *dims)
...
     config->dims_get(config_, dims, "nu", &nu);
     config->dims_get(config_, dims, "nz", &nz);
+    config->dims_get(config_, dims, "np", &np);
 
     int NF = nx + nu;
     size += sizeof(sim_info);
 
     size += nx * sizeof(double);                // xn
     size += nx * NF * sizeof(double);           // S_forw
+    size += nx * np * sizeof(double);       // S_p (parameter sens, may be 0)
```
FreyJo: Most new memory assignments and all potentially costly operations should be conditional on the option: `sens_forw_p`.
In particular, since often np >> nx, assigning nx * np memory can increase the memory footprint significantly.

Here in `sim_out`, there is the issue that the assign functions do not depend on `opts`.
I would mitigate this, by not including `S_p` in `sim_out` and instead directly get it from the integrator memory `sim_erk_memory_get`.

examples/acados_matlab_octave/test/test_sens_forw_p.m
```diff
@@ -0,0 +1,203 @@
...
+sim_opts.set('num_stages', num_stages);
+sim_opts.set('num_steps', num_steps);
+sim_opts.set('newton_iter', newton_iter);
+sim_opts.set('method', method);
+sim_opts.set('sens_forw', sens_forw);
+sim_opts.set('sens_forw_p', sens_forw_p);   
+sim_opts.set('jac_reuse', jac_reuse);
+
+%% acados sim
+sim_solver = acados_sim(sim_model, sim_opts);
```
FreyJo, line 100: was there a reason to use the legacy interface here?
(the "modern" way is to use `AcadosSim` and `AcadosSimSolver`).
But it is also fine like this.

interfaces/acados_template/acados_template/c_templates_tera/acados_sim_solver.in.c
```diff
@@ -292,6 +308,13 @@ int {{ model.name }}_acados_sim_create({{ model.name }}_sim_solver_capsule * cap
     sim_opts_set({{ model.name }}_sim_config, {{ model.name }}_sim_opts, "sens_hess", &tmp_bool);
     tmp_bool = {{ solver_options.output_z }};
     sim_opts_set({{ model.name }}_sim_config, {{ model.name }}_sim_opts, "output_z", &tmp_bool);
+    // enable S_p only if user asks for it *and* parameters exist
+    tmp_bool = {{ solver_options.sens_forw_p | default(value="false") }};
+    if (tmp_bool)
+    {
+        tmp_bool = ({{ dims.np }} > 0 );
+    }
```
FreyJo: Is there a reason to not allow this for an empty parameter vector (np = 0)?
It should just work IMO.
So, I would remove the extra check.

interfaces/acados_template/acados_template/c_templates_tera/acados_sim_solver.in.c
```diff
@@ -144,7 +144,12 @@ int {{ model.name }}_acados_sim_create({{ model.name }}_sim_solver_capsule * cap
     capsule->sim_expl_vde_forw = (external_function_param_{{ model.dyn_ext_fun_type }} *) malloc(sizeof(external_function_param_{{ model.dyn_ext_fun_type }}));
     capsule->sim_vde_adj_casadi = (external_function_param_{{ model.dyn_ext_fun_type }} *) malloc(sizeof(external_function_param_{{ model.dyn_ext_fun_type }}));
     capsule->sim_expl_ode_fun_casadi = (external_function_param_{{ model.dyn_ext_fun_type }} *) malloc(sizeof(external_function_param_{{ model.dyn_ext_fun_type }}));
-
+	{% if dims.np > 0 %}
+		capsule->sim_expl_vde_forw_p = (external_function_param_{{ model.dyn_ext_fun_type }} *) malloc(sizeof(external_function_param_{{ model.dyn_ext_fun_type }}));
+	{% else %}
+		capsule->sim_expl_vde_forw_p = NULL;
+	{% endif %}
+
```
FreyJo: I think all newly added ifs should be checking the new option rather than the dimension `np`.

interfaces/acados_template/acados_template/c_templates_tera/acados_solver.in.c
```diff
@@ -2471,6 +2481,21 @@ static void {{ model.name }}_acados_create_set_opts({{ model.name }}_solver_caps
     free(sim_method_jac_reuse);
   {%- endif %}
 
+{%- if solver_options.integrator_type == "ERK" %}
+    // Enable parameter forward sensitivities S_p at all stages for ERK,
+    // but only if user enabled it *and* parameters exist.
+    bool dynamics_sens_forw_p_val = {{ solver_options.sens_forw_p | default(value="false") }};
```
FreyJo: The interfaces ensure that all options have valid values, so the ` | default(value="false") ` should not be needed.
In particular, I think they might cause issues with older versions of tera renderer.

interfaces/acados_template/acados_template/casadi_function_generation.py
```diff
@@ -311,14 +312,21 @@ def generate_c_code_explicit_ode(context: GenerateContext, model: AcadosModel, m
...
 
     fun_name = model_name + '_expl_vde_adj'
     context.add_function_definition(fun_name, [x, lambdaX, u, p], [adj], model_dir, 'dyn')
 
     if generate_hess:
         fun_name = model_name + '_expl_ode_hess'
-        context.add_function_definition(fun_name, [x, Sx, Sp, lambdaX, u, p], [adj, hess2], model_dir, 'dyn')
+        context.add_function_definition(fun_name, [x, Sx, Su, lambdaX, u, p], [adj, hess2], model_dir, 'dyn')
+
+    # NEW: param-direction forward VDE
```
FreyJo: don't write `NEW` in comments, this outdates quickly.

interfaces/acados_template/acados_template/casadi_function_generation.py
```diff
@@ -311,14 +312,21 @@ def generate_c_code_explicit_ode(context: GenerateContext, model: AcadosModel, m
...
+    if np > 0:
+        Sp = symbol('Sp', nx, np)
+        vdeP = ca.jacobian(f_expl, p) + ca.jtimes(f_expl, x, Sp)   # f_p + A*Sp
```
FreyJo: This function generation should also be conditional on the new option.
The derivative of model equations wrt parameters can be arbitrarily complicated, or not even exists.
E.g. if splines are within the model.
So, we should only create and compile those functions if needed.

examples/acados_matlab_octave/test/test_sens_forw_p.m
ivandoornen: I aimed to keep the new test script consistent with existing scripts (e.g., `test_sens_hess.m` and `test_sens_forw.m`). Since those rely on the legacy interface, I chose to use it here as well.

ivandoornen:
I believe I have implemented all your comments. Let me know if you have any further comments!

FreyJo:
The MATLAB test is failing with a segfault. Can you fix it?
I will check the PR later this week.

ivandoornen:
> The MATLAB test is failing with a segfault. Can you fix it? I will check the PR later this week.

I think it should be fixed now!

acados/sim/sim_erk_integrator.c
```diff
@@ -337,6 +356,17 @@ void *sim_erk_memory_assign(void *config, void *dims, void *opts_, void *raw_mem
 
     sim_erk_memory *mem = (sim_erk_memory *) c_ptr;
     c_ptr += sizeof(sim_erk_memory);
+	
+    mem->S_p = NULL;
+
+    sim_erk_dims *erk_dims = (sim_erk_dims *) dims;
+    sim_opts *opts = (sim_opts *) opts_;
+
+    if (opts->sens_forw && opts->sens_forw_p )
```
FreyJo: why do you check for both? I would have just checked `opts->sens_forw_p`.

acados/sim/sim_erk_integrator.c
```diff
@@ -651,35 +735,33 @@ int sim_erk(void *config_, sim_in *in, sim_out *out, void *opts_, void *mem_, vo
...
+	
     double *xn = out->xn;
     double *S_forw_out = out->S_forw;
     double *S_adj_out = out->S_adj;
     double *S_hess_out = out->S_hess;
 
-    ext_fun_arg_t ext_fun_type_in[5];
-    void *ext_fun_in[5];
+	ext_fun_arg_t ext_fun_type_in[5];
+    void *ext_fun_in[5];
```
FreyJo: Minor comment on formatting:
- There are a few places where some trailing spaces are added
- We prefer indenting with spaces
I think we can still merge as it is, but we will likely be reformat it accordingly.

interfaces/acados_template/acados_template/c_templates_tera/acados_solver.in.c
```diff
@@ -2471,6 +2481,16 @@ static void {{ model.name }}_acados_create_set_opts({{ model.name }}_solver_caps
     free(sim_method_jac_reuse);
   {%- endif %}
 
+{%- if solver_options.integrator_type == "ERK" %}
```
FreyJo: I would remove the if here and add sanity checks that `sens_forw_p` can only be True if ERK is used in `make_consistent` in the high level interfaces.

FreyJo (commented):
Looks good!
Can you answer the remaining comments?

FreyJo:
I can also do the remaining minor changes and push them here before merging.
I think the only question I still have is on the condition:
`if (opts->sens_forw && opts->sens_forw_p )`

ivandoornen:
> I can also do the remaining minor changes and push them here before merging. I think the only question I still have is on the condition: `if (opts->sens_forw && opts->sens_forw_p )`

Fine with me! It is also no problem for me to address them myself. 

I check both because the calculation of `S_p` is structurally nested inside the `if (opts->sens_forw)` block in the integrator.

Mathematically, $\dot{S}_p = f_x S_p + f_p$ requires the state Jacobian $f_x$. If `sens_forw` is false, the integrator takes the "simulation only" path (`expl_ode_fun`), which doesn't evaluate derivatives. We need `sens_forw=true` to switch to the VDE path (`expl_vde_for`) where the Jacobian is available to drive the `S_p` propagation.

FreyJo:
Ok, thanks for the explanation.
I would then rather replace the corresponding
`if (opts->sens_forw)` checks with
`if (opts->sens_forw || opts->sens_forw_p)`.
Such that $S_p$ is always computed when `sens_forw_p` is true.
And potentially $S_x$ is computed despite `sens_forw = False`, just as a by product.
Does that make sense?

Also thanks for taking care of the remaining points.
Please just make a non-forced push on top, such that I only need to review the new commits.

ivandoornen:
> Ok, thanks for the explanation. I would then rather replace the corresponding `if (opts->sens_forw)` checks with `if (opts->sens_forw || opts->sens_forw_p)`. Such that S p is always computed when `sens_forw_p` is true. And potentially S x is computed despite `sens_forw = False`, just as a by product. Does that make sense?

I see, so essentially you would prefer that whenever `opts->sens_forw_p` you take the VDE path, regardless of  `opts->sens_forw`? Makes sense to me!

FreyJo:
Exactly. And also go the VDE path if `opts->sens_forw` is True of course.

ivandoornen:
I think I have now implemented your last comments. I also scanned through all files and removed any trailing whitespaces and changed indents to spaces.

PR #1718, opened 2025-12-01
Python: load OCP from JSON

sandmaennchen:


interfaces/acados_template/acados_template/acados_ocp.py
```diff
@@ -2570,3 +2566,23 @@ def create_default_initial_iterate(self) -> AcadosOcpIterate:
...
+                field_dict = dict.get(field)
+
+                if field_dict is not None:
+                    setattr(ocp, field, type(getattr(ocp, field)).from_dict(field_dict))
+                else:
+                    raise Exception(f"Failed to load OCP from json. Field {field} is not provided.")
+            else:
+                setattr(ocp, field, dict.get(field))
+
+        # ocp.make_consistent()
```
sandmaennchen, line 2587: I think we need to call `make_consistent` here otherwise `has_x0` is not set correctly. I think we should include a test that compares the json files directly to test this

examples/acados_python/tests/load_ocp.py
```diff
@@ -0,0 +1,28 @@
...
+
+    # load json, store options in object
+    with open(json_file, 'r') as f:
+        acados_ocp_json = json.load(f)
+    acados_ocp_json['json_file'] = os.path.abspath(json_file)
+
+    warnings.filterwarnings("ignore", message=".*not in dictionary.*", category=UserWarning)
+    ocp = AcadosOcp().from_dict(acados_ocp_json)
+
+    ocp.render_templates()
```
sandmaennchen: why do you call this here?

examples/acados_python/tests/load_ocp.py
```diff
@@ -0,0 +1,28 @@
...
+
+def create_solver_from_json(json_file):
+
+    # load json, store options in object
+    with open(json_file, 'r') as f:
+        acados_ocp_json = json.load(f)
+    acados_ocp_json['json_file'] = os.path.abspath(json_file)
+
+    warnings.filterwarnings("ignore", message=".*not in dictionary.*", category=UserWarning)
+    ocp = AcadosOcp().from_dict(acados_ocp_json)
```
sandmaennchen: ` AcadosOcp.from_dict(acados_ocp_json)`

examples/acados_python/tests/load_ocp.py
FreyJo: I wanted to test this separately. Removed.

examples/acados_python/tests/load_ocp.py
FreyJo: done

interfaces/acados_template/acados_template/acados_ocp.py
FreyJo: `make_consistent` does work.
In the test I create a solver with `solver = AcadosOcpSolver(ocp)`.
One use case of the functionality is to debug errors in the template rendering process.
If there are issues with dumped casadi expressions, `make_consistent` might fail.
But one can still debug the rendering without consistent casadi model definitions.

sandmaennchen: I just think that people might be confused if they load from json and then dump to json again and the two jsons are not the same (if you do not call make consistent in between loading and dumping)

sandmaennchen: but maybe we just add this as a note to the docstring and we are all good

FreyJo (approved):

Issue #1684, opened 2025-11-10
Refactor example


sandmaennchen:
should use N=0

- [x] https://github.com/acados/acados/blob/main/examples/acados_python/convex_problem_globalization_needed/convex_problem_globalization_necessary.py
- [x] https://github.com/acados/acados/tree/main/examples/acados_python/hock_schittkowsky
- [x] `examples/acados_python/inconsistent_qp_linearization/inconsistent_qp_linearization_test.py`
- [x] `examples/acados_python/non_ocp_nlp/adaptive_eps_reg_test.py`
- [x] `examples/acados_python/non_ocp_nlp/maratos_test_problem.py`
- [x] `examples/acados_python/tests/armijo_test.py`

PR #1719, opened 2025-12-01
Python: Refactor example with N=0

Pandatheon:
This PR aims to adress this issue https://github.com/acados/acados/issues/1684.

### DIFF
--- a/examples/acados_python/inconsistent_qp_linearization/inconsistent_qp_linearization_test.py
+++ b/examples/acados_python/inconsistent_qp_linearization/inconsistent_qp_linearization_test.py
@@ -106,9 +106,13 @@ def test_byrd_omojokun_qps():
     # feasibility QP solution should be (d = -10, s = 3.98)
     # here should be a test of the bounds
 
-    # nominal QP solution should be d= -10
+    # nominal QP solution should be d= -10 with N = 1
+    # nominal QP solution should be d= -20 with N = 0 
     d = iter1.x_traj[0] - iter0.x_traj[0]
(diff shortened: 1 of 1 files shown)


sandmaennchen:
Looks good! Did you check the behaviour before and after the change? Do you get the same number of iterations?

examples/acados_python/inconsistent_qp_linearization/inconsistent_qp_linearization_test.py
```diff
@@ -106,9 +106,13 @@ def test_byrd_omojokun_qps():
     # feasibility QP solution should be (d = -10, s = 3.98)
     # here should be a test of the bounds
 
-    # nominal QP solution should be d= -10
+    # nominal QP solution should be d= -10 with N = 1
+    # nominal QP solution should be d= -20 with N = 0
```
sandmaennchen, line 110: why is this the case?

FreyJo: I also can't explain it just from what I know and reading the test problem definition.
I think it would be good to check as it might reveal some bug or unexpected cost scaling in the relaxed QP.
I can do that soon.

FreyJo: I just checked both relaxed QPs. The difference come from the fact, that with N=1, a quadratic cost term is added for two stage costs, while for N=0, this is only done for one.

PR #1722, opened 2025-12-04
Docs: minor updates

josipkh:
This PR proposes a few minor changes:
- removed an obsolete line and duplicate file in MATLAB environment setup
- fix a few typos in Python docstrings and MATLAB warning text
- fix the [QP solver status tables](https://docs.acados.org/python_interface/index.html#acados_template.acados_ocp_options.AcadosOcpOptions.qp_solver)
- add basic docstrings for iterate-getting functions
- add Clarabel as a possible QP solver in the problem formulation pdf
- add a success note to the MATLAB install script

### DIFF
--- a/examples/acados_matlab_octave/acados_env_variables_windows.m
+++ b/examples/acados_matlab_octave/acados_env_variables_windows.m
@@ -35,10 +35,8 @@ example_dir = fileparts(which('acados_env_variables_windows'));
 acados_dir = fullfile(example_dir, '..', '..');
 casadi_dir = fullfile(acados_dir, 'external', 'casadi-matlab');
 matlab_interface_dir = fullfile(acados_dir, 'interfaces', 'acados_matlab_octave');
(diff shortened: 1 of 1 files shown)


sandmaennchen:
Thanks @josipkh! Why do you move the matlab install script? This might be a breaking change for some people ...

josipkh:
Hi Katrin :)
I thought it would be good to avoid duplicate files so I removed `acados_env_variables_windows` file from the `examples` folder and left the one in `interfaces`. It's also possible the other way around if you would prefer that.
In the MATLAB install script, I just changed some printing.

FreyJo:
Hi Josip,
I agree with all changes, but would like to keep the `acados_env_variables_windows` in the examples folder.
Users are more encouraged to start their MATLAB acados session there and the duplication does hurt less than potentially breaking some user experience.
Can you just revert this change please?
Then it can be merged.
Thanks for the contribution!

josipkh:
Of course, I brought back the deleted file (not sure if you meant in a different way).
Thank you for the quick feedback!

PR #1726, opened 2025-12-15
Windows: Compile shared and static libs, simultaneous use of Python and MATLAB interface

FreyJo:
Document the workflow to use Python and MATLAB interfaces simultaneously on Windows.
The MATLAB interface uses static libraries on Windows and the Python interface has to be used with shared libs.
If one wants to use both interfaces without recompiling acados in between, the following workflow was tested.

In MATLAB, run the following script
```
acados_install_shared_and_static_windows()
```
This internally uses `acados_install_windows`.
It compiles first the shared libraries and moves them into `<acados_root>/bin`.
Then, it compiles the static libraries, they are located in `<acados_root>/lib`.

### DIFF
--- a/interfaces/acados_template/acados_template/acados_multiphase_ocp.py
+++ b/interfaces/acados_template/acados_template/acados_multiphase_ocp.py
@@ -175,8 +175,7 @@ class AcadosMultiphaseOcp:
             """Path to where acados library is located"""
         self.acados_lib_path.replace(os.sep, '/')
 
-        self.acados_include_path = os.path.join(acados_path, 'include').replace(os.sep, '/') # the replace part is important on Windows for CMake
-        """Path to acados include directory (set automatically), type: `string`"""
+        self.__acados_include_path = os.path.join(acados_path, 'include').replace(os.sep, '/')
 
         self.shared_lib_ext = get_shared_lib_ext()
 
@@ -195,6 +194,10 @@ class AcadosMultiphaseOcp:
         self.simulink_opts = None
         """Options to configure Simulink S-function blocks, mainly to activate possible Inputs and Outputs."""
 
+    @property
+    def acados_include_path(self):
+        """Path to acados include directory (set automatically), type: `string`"""
+        return self.__acados_include_path
 
     @property
     def parameter_values(self):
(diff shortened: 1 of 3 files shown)


interfaces/acados_template/acados_template/acados_multiphase_ocp.py
```diff
@@ -161,12 +164,20 @@ def __init__(self, N_list: list):
...
+        # acados paths
+        if acados_path is None:
+            acados_path = get_acados_path()
+
+        if acados_lib_path is not None:
+            self.acados_lib_path = acados_lib_path
+        else:
+            self.acados_lib_path = os.path.join(acados_path, 'lib')
+            """Path to where acados library is located"""
+        self.acados_lib_path.replace(os.sep, '/')
```
sandmaennchen, line 176: this should be a read only property

interfaces/acados_template/acados_template/acados_ocp.py
```diff
@@ -79,14 +79,26 @@ class AcadosOcp:
...
             acados_path = get_acados_path()
 
+        if acados_lib_path is not None:
+            self.acados_lib_path = acados_lib_path
+        else:
+            self.acados_lib_path = os.path.join(acados_path, 'lib')
+            """Path to where acados library is located"""
+        self.acados_lib_path.replace(os.sep, '/')
+
+        self.acados_include_path = os.path.join(acados_path, 'include').replace(os.sep, '/') # the replace part is important on Windows for CMake
```
sandmaennchen: this should be a read only property

interfaces/acados_template/acados_template/acados_sim.py
```diff
@@ -333,20 +333,33 @@ class AcadosSim:
...
             acados_path = get_acados_path()
+
+        if acados_lib_path is not None:
+            self.acados_lib_path = acados_lib_path
+        else:
+            self.acados_lib_path = os.path.join(acados_path, 'lib')
+            """Path to where acados library is located"""
+        self.acados_lib_path.replace(os.sep, '/')
+
+        self.acados_include_path = os.path.join(acados_path, 'include').replace(os.sep, '/') # the replace part is important on Windows for CMake
```
sandmaennchen: this should be a read only property

PR #1740, opened 2026-01-07
MATLAB/python: rename attributes in iterate

sandmaennchen:
This PR renames the attributes in the `AcadosOcpIterate` class from using the `_traj` suffix (e.g., `x_traj`, `u_traj`) to just the field names (e.g., `x`, `u`) across both Python and MATLAB interfaces in order to improve naming consistency.

### DIFF
--- a/interfaces/acados_template/acados_template/acados_ocp_iterate.py
+++ b/interfaces/acados_template/acados_template/acados_ocp_iterate.py
@@ -137,25 +137,25 @@ class AcadosOcpIterate:
 
         if x_traj is not None:
             self.x = x_traj
-            warnings.warn("Parameter 'x_traj' is deprecated, use 'x' instead.", DeprecationWarning)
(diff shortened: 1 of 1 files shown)


interfaces/acados_template/acados_template/acados_ocp_iterate.py
```diff
@@ -115,24 +117,100 @@ class AcadosOcpIterate:
...
+        self.u = u
+        self.z = z
+        self.sl = sl
+        self.su = su
+        self.pi = pi
+        self.lam = lam
+
+        if x_traj is not None:
+            self.x = x_traj
+            warnings.warn("Parameter 'x_traj' is deprecated, use 'x' instead.", DeprecationWarning)
```
FreyJo: I recently learned that one should add `stacklevel=2` for deprecation warnings, such that the warning points to the caller of the deprecated function.

PR #1741, opened 2026-01-07
sim: IRK support parameter sensitivity S_p

ivandoornen:
## Summary

This PR extends the existing `sens_forw_p` feature (forward parameter sensitivities `S_p`) to the IRK integrator (implicit dynamics), in addition to the already supported ERK path. It adds the required implicit Jacobian w.r.t. parameters, plumbs it through generated code/templates, and updates tests so that both ERK and IRK are verified against finite differences.

For one simulation step
$$x_{n+1} = \Phi(x_n, u_n, p),$$
we expose
$$S_p = \frac{\partial x_{n+1}}{\partial p} \in \mathbb{R}^{n_x \times n_p}.$$

---

## Math / implementation notes (IRK)

For IRK, the step is defined implicitly via stage variables $K$ through a residual equation
$G(K, x, u, p) = 0,$
and the state update depends on these stages:
$$x_{n+1} = x_n + h \sum_{i=1}^{s} b_i k_i .$$

Differentiating the implicit system gives
$\frac{\partial G}{\partial K}\frac{\partial K}{\partial p} + \frac{\partial G}{\partial x} S_p + \frac{\partial G}{\partial p} = 0,$
so at each step we solve the linear system
$\frac{\partial G}{\partial K}\frac{\partial K}{\partial p}= -\Big(\frac{\partial G}{\partial x} S_p + \frac{\partial G}{\partial p}\Big),$
and then update $S_p$ through the IRK state update relation.

This PR adds the missing implicit Jacobian $\partial f/\partial p$ hook (as `impl_dae_jac_p`) and uses it to form the right-hand side above.

---

## What changed (high level)

- **IRK integrator**: adds `np` support, allocates/caches `S_p`, and computes it when `sens_forw_p=true`.
- **New external function hook** for implicit models: `impl_dae_jac_p`.
- **Codegen / templates / MATLAB MEX plumbing** updated so `impl_dae_jac_p` is generated (CasADi) and wired into sim + OCP creation when requested.
- **Interface checks** updated so `sens_forw_p` is accepted for both ERK and IRK (MATLAB + Python templates).

---

## Tests

- MATLAB test `test_sens_forw_p.m` now runs the FD verification for ERK and IRK.
- Python tests `test_sens_forw_p.py` and `test_ocp_sens_p.py` now cover ERK and IRK, checking `S_p` against finite differences (sim) and validating OCP `S_p` consistency.

### DIFF
--- a/acados/sim/sim_common.h
+++ b/acados/sim/sim_common.h
@@ -57,7 +57,6 @@ typedef enum
     IMPL_ODE_FUN_JAC_X_XDOT,
     IMPL_ODE_JAC_X_XDOT_U,
     IMPL_ODE_FUN_JAC_X_XDOT_U,
-    IMPL_DAE_JAC_P,
     IMPL_ODE_HESS,
     // gnsf
     PHI_FUN,
--- a/acados/sim/sim_irk_integrator.c
+++ b/acados/sim/sim_irk_integrator.c
@@ -713,7 +713,7 @@ acados_size_t sim_irk_workspace_calculate_size(void *config_, void *dims_, void 
         size += 1 * blasfeo_memsize_dmat(nK, nx + nu);  // dK_dxu
         size += 1 * blasfeo_memsize_dmat(nx, nx + nu);  // S_forw
         size += nK * sizeof(int);  // ipiv
-	}
+    }
     else
     {
         size += steps * blasfeo_memsize_dmat(nK, nx + nu);      // dG_dxu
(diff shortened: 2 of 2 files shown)


acados/sim/sim_irk_integrator.c
```diff
@@ -665,7 +713,7 @@ acados_size_t sim_irk_workspace_calculate_size(void *config_, void *dims_, void
         size += 1 * blasfeo_memsize_dmat(nK, nx + nu);  // dK_dxu
         size += 1 * blasfeo_memsize_dmat(nx, nx + nu);  // S_forw
         size += nK * sizeof(int);  // ipiv
-    }
+	}
```
FreyJo: Minor: formatting changed.

acados/sim/sim_irk_integrator.c
FreyJo: I think that this block is not needed if only sensitivities wrt p are needed.
So, we can add an `if (sens_forw || sens_hess)` around it.

acados/sim/sim_irk_integrator.c
```diff
@@ -1565,6 +1660,42 @@ int sim_irk(void *config_, sim_in *in, sim_out *out, void *opts_, void *mem_, vo
...
+
+                    // dK_dp += df/dp
+                    blasfeo_dgead(nx + nz, dims->np, 1.0, df_dp, 0, 0, dK_dp, ii * (nx + nz), 0);
+                }
+
+                // solve linear system
+                blasfeo_drowpe(nK, ipiv_ss, dK_dp);
+                blasfeo_dtrsm_llnu(nK, dims->np, 1.0, dG_dK_ss, 0, 0, dK_dp, 0, 0, dK_dp, 0, 0);
+                blasfeo_dtrsm_lunn(nK, dims->np, 1.0, dG_dK_ss, 0, 0, dK_dp, 0, 0, dK_dp, 0, 0);
+
```
FreyJo, line 1696: you can use the unpacked `np` instead of `dims->np`.

acados/sim/sim_common.h
FreyJo, line 67: I think this whole enum is not used at all. You can try to remove it.

PR #1742, opened 2026-01-08
interfaces: Expose zoRO P_k matrices and enhance Simulink I/O

ivandoornen:
## Summary

This PR exposes the zoRO uncertainty matrices $P_k \in \mathbb{R}^{n_x \times n_x}$ consistently across the generated solver C API, Python, and MATLAB interfaces, and improves the Simulink OCP block I/O for a smoother zoRO workflow.

In zoRO, $P_k$ represents the propagated state uncertainty used for tightening. This commit does not change the underlying zoRO math, it only standardizes how the resulting $P_k$ matrices are accessed and streamed.

---

### Unified exposure of zoRO $P_k$ matrices (`zoRO_Pk_mats`)
The field `zoRO_Pk_mats` is retrievable via:

- **C interface (generated solver API):**
  - `{{ model.name }}_acados_get_zoRO_Pk_matrices(capsule, P_out, P_out_len)` exports all $P_k$ matrices into a flat output buffer.
  - Layout is column-major blocks: $[P_0(:);\;P_1(:);\;\dots;\;P_N(:)]$ with block size `nx*nx`.

- **Python:**
  - `AcadosOcpSolver.get(k, "zoRO_Pk_mats")` allocates a `(nx, nx)` matrix (Fortran-ordered) and retrieves $P_k$.
  - Internally, Python calls the generated symbol `{{name}}_acados_get_zoRO_Pk_matrices`, then slices out the requested stage.
  - If the solver library does not provide the zoRO getter symbol, Python raises an error indicating that `zoRO_Pk_mats` is unavailable in this build.

- **MATLAB/MEX:**
  - `ocp.get_zoRO_Pk_matrices()` returns a `1x(N+1)` cell array of `[nx x nx]` matrices.
  - `ocp.get_zoRO_Pk_matrices(k)` returns a single `[nx x nx]` matrix for stage `k`.
  - This is implemented via a dedicated generated MEX entry point `acados_mex_get_zoRO_Pk_{{ name }}` (only present when custom update / zoRO is enabled).

---

### External payload input
- Adds an optional inport `zoRO_payload` that forwards an external payload buffer to `custom_update_function`.
- Supports **unwired payload ports** by treating them as `NULL/0` length and falling back to build time uncertainty.

### Direct $P_k$ output ports
- Adds an optional output `zoRO_Pk` that exposes the concatenation of all $P_k$ matrices for $k = 0,\dots,N$.
- Implements a helper `{{ model.name }}_acados_get_zoRO_Pk_matrices(...)` to flatten and export the internal zoRO buffer into the Simulink output.

---

## Behavior / compatibility

- Off by default: new Simulink options `inputs.zoRO_payload` and `outputs.zoRO_Pk_matrices` default to `0`.
- Payload size is validated when streaming is enabled; invalid payload sizes are reported as an error.
- The legacy shared I/O data workflow is still supported: streaming matrices via the single `data` payload and retrieving uncertainty outputs from the same buffer continues to work unchanged.

### DIFF
--- a/interfaces/acados_matlab_octave/AcadosOcp.m
+++ b/interfaces/acados_matlab_octave/AcadosOcp.m
@@ -1815,7 +1815,9 @@ classdef AcadosOcp < handle
 
             if ~isempty(self.solver_options.custom_update_filename)
                 template_list{end+1} = {fullfile(matlab_template_path, 'acados_mex_custom_update.in.c'), ['acados_mex_custom_update_', self.name, '.c']};
-                template_list{end+1} = {fullfile(matlab_template_path, 'acados_mex_get_zoRO_Pk.in.c'), ['acados_mex_get_zoRO_Pk_', self.name, '.c']};
+                if ~isempty(self.zoro_description)
+                    template_list{end+1} = {fullfile(matlab_template_path, 'acados_mex_get_zoRO_Pk.in.c'), ['acados_mex_get_zoRO_Pk_', self.name, '.c']};
+                end
             end
 
             % append headers
--- a/interfaces/acados_template/acados_template/acados_ocp_solver.py
+++ b/interfaces/acados_template/acados_template/acados_ocp_solver.py
@@ -423,13 +423,10 @@ class AcadosOcpSolver:
 
         # zoRO getter (only present for zoRO builds)
         self.__zoro_getter = None
-        try:
+        if acados_ocp_json.get('zoro_description'):
             self.__zoro_getter = getattr(self.__shared_lib, f"{self.name}_acados_get_zoRO_Pk_matrices")
             self.__zoro_getter.argtypes = [c_void_p, POINTER(c_double), c_int]
             self.__zoro_getter.restype  = c_int
-        except AttributeError:
-            self.__zoro_getter = None
-
 
         getattr(self.shared_lib, f"{self.name}_acados_set_p_global_and_precompute_dependencies").argtypes = [c_void_p, POINTER(c_double), c_int]
         getattr(self.shared_lib, f"{self.name}_acados_set_p_global_and_precompute_dependencies").restype = c_int
--- a/interfaces/acados_template/acados_template/c_templates_tera/matlab_templates/acados_mex_get_zoRO_Pk.in.c
+++ b/interfaces/acados_template/acados_template/c_templates_tera/matlab_templates/acados_mex_get_zoRO_Pk.in.c
@@ -4,6 +4,28 @@
  * This file is part of acados.
  *
  * The 2-Clause BSD License
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ * this list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.;
  */
 
 // system
--- a/interfaces/acados_template/acados_template/c_templates_tera/matlab_templates/acados_solver_sfun.in.c
+++ b/interfaces/acados_template/acados_template/c_templates_tera/matlab_templates/acados_solver_sfun.in.c
@@ -376,7 +376,7 @@ static void mdlInitializeSizes (SimStruct *S)
     {%- set n_outputs = n_outputs + 1 %}
   {%- endif %}
 
-  {% if simulink_opts.outputs.zoRO_P_matrices == 1%}
+  {% if simulink_opts.outputs.zoRO_Pk_matrices == 1%}
     {%- set n_outputs = n_outputs + 1 %}
   {% endif %}
 
@@ -718,7 +718,7 @@ static void mdlInitializeSizes (SimStruct *S)
     ssSetOutputPortVectorDimension(S, {{ i_output }}, {{ np_total }});
   {%- endif -%}
 
-  {%- if simulink_opts.outputs.zoRO_P_matrices == 1 %}
+  {%- if simulink_opts.outputs.zoRO_Pk_matrices == 1 %}
     {%- set i_output = i_output + 1 %}
     ssSetOutputPortVectorDimension(S, {{ i_output }}, {{ dims_0.nx * (solver_options.N_horizon+1) * dims_0.nx }} );
   {%- endif %}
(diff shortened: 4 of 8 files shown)


interfaces/acados_template/acados_template/custom_update_templates/custom_update_function_zoro_template.in.c
```diff
@@ -1511,18 +1531,24 @@ int custom_update_function({{ model.name }}_solver_capsule* capsule, double* dat
...
 {%- else %}
+    if (streaming > 0)
+    {
     {%- if zoro_description.output_riccati_t %}
         data[custom_mem->offset_P_out] = time_riccati;
     {%- endif %}
+    }
 {%- endif %}
 
     return 1;
```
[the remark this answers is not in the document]
ivandoornen: Thanks co-pilot, this was actually implemented in one of my commits to my own dev branch but accidentally got dropped after reordering my commit history :)

acados/ocp_nlp/ocp_nlp_common.h
```diff
@@ -476,6 +476,9 @@ typedef struct ocp_nlp_memory
     struct blasfeo_dvec *dyn_fun;
     struct blasfeo_dvec *dyn_adj;
 
+    // zoRO optional cache (set by custom_update code if enabled)
+    struct blasfeo_dmat *zoro_Pk_mats; // NULL if unavailable
```
FreyJo: I don't see where this memory is set. I didn't find it in the custom update function.

FreyJo: I don't think we should actually put any zoRO stuff in the acados core.
Right now, the zoRO update is pretty clearly separated from the acados core.
All zoRO stuff lives in the custom update and associated memory.
I see that it makes assessing things from the algorithm maybe more tedious, as we cannot use the acados interface structure, but instead everything goes through the custom update data.
However, I would really like to keep this separation.

FreyJo: How those matrices are retrieved in Simulink is independent of this code. But somehow the Python wrapper uses this additional change in the core, which seems not even functional, see the first comment.

ivandoornen: Thanks for the review @FreyJo!

1. **Missing hook:** The code that sets `nlp_mem->zoro_Pk_mats` was accidentally dropped when I reordered commits. I have put it back with the latest commit.

2. **Architectural concern:** I get the motivation to keep zoRO isolated from the acados core. From my side, I see two reasonable options:

**Option A (current approach):**  
Keep the thin `zoro_Pk_mats` pointer in `ocp_nlp_memory`. The core still has no zoRO logic, it only stores a pointer that `custom_update` sets during init. This gives users a consistent `get(stage, "zoRO_Pk_mats")` API that matches other fields like `x`, `u`, `S_p`.

**Option B (full isolation):**  
Remove `zoro_Pk_mats` from the core entirely. Since the PR already generates `{{ model.name }}_acados_get_zoRO_P_matrices()` for Simulink, Python/MATLAB could call that directly instead of going through `ocp_nlp_get_at_stage`. That would keep all zoRO code in `custom_update`, with a slightly different API pattern.

Let me know which direction you prefer. I’m happy to refactor toward Option B if you want to strictly sperate zoRO.

ivandoornen: However, I’d like to support zoRO with the full SQP solver in another future PR. At the moment, users have to manually loop SQP iterations and insert `custom_update` calls in between.

My proposal adds a generic callback hook in `ocp_nlp_sqp.c`:

    typedef int (*ocp_nlp_sqp_custom_update_fun)(void *ctx, double *data, int data_len);

The hook is called after `ocp_nlp_approximate_qp_matrices()` and before  
`ocp_nlp_approximate_qp_vectors_sqp()`.

I couldn't find a cleaner alternative that avoids core changes while enabling per-iteration callbacks. What do you think?

FreyJo: I am strongly in favor for option B.
Many of the SQP features do not work correctly if custom update is called in between.
In particular, residual computations and globalization come to mind here.
The burden of looping outside is not too much IMO and makes more clear what is happening.
Additionally, keeping the separation of core and custom update a design choice that we really want to keep.

ivandoornen: Do you have an idea how looping outside would work in a Simulink implementation?

FreyJo: You can just add a for loop around this block.
https://github.com/acados/acados/blob/6b7820f089e4b95498dbd6ae8df76bdd2c5d5178/interfaces/acados_template/acados_template/c_templates_tera/matlab_templates/acados_solver_sfun.in.c#L1312-L1332
Could even be an option in `simulink_opts.zoro_iterations` with default 1.

ivandoornen: Could you take a look if the refactor is more to your liking?

FreyJo:
Could you please revise the PR description to reflect the changes implemented after the refactoring?

ivandoornen:
> Could you please revise the PR description to reflect the changes implemented after the refactoring?

Ah yes sorry, I forgot. Thanks for reminding me, it should now be updated

interfaces/acados_matlab_octave/AcadosOcp.m
```diff
@@ -1815,6 +1815,7 @@ function render_templates(self)
 
             if ~isempty(self.solver_options.custom_update_filename)
                 template_list{end+1} = {fullfile(matlab_template_path, 'acados_mex_custom_update.in.c'), ['acados_mex_custom_update_', self.name, '.c']};
+                template_list{end+1} = {fullfile(matlab_template_path, 'acados_mex_get_zoRO_Pk.in.c'), ['acados_mex_get_zoRO_Pk_', self.name, '.c']};
```
FreyJo: Add `if ~isempty(self.zoro_description)` around this line. As custom update is more generic than zoro.

interfaces/acados_template/acados_template/c_templates_tera/matlab_templates/acados_mex_get_zoRO_Pk.in.c
```diff
@@ -0,0 +1,80 @@
+/*
+ * Copyright (c) The acados authors.
+ *
+ * This file is part of acados.
+ *
+ * The 2-Clause BSD License
+ */
```
FreyJo, line 29: Please use the standard C header, to keep it consistent.
I am open for changing to a more concise one, like here, in the future.

interfaces/acados_template/acados_template/c_templates_tera/matlab_templates/acados_solver_sfun.in.c
```diff
@@ -1318,10 +1338,27 @@ static void mdlOutputs(SimStruct *S, int_T tid)
...
+      data_len = ssGetInputPortWidth(S, {{ zoro_port_index }});
+      if (data_len > 0) {
+    // Simulink guarantees this pointer is contiguous
+        c_data = (double *) ssGetInputPortRealSignal(S, {{ zoro_port_index }});
+      }
+    }
+  {% endif %}
+
+    // After RTI prep (rti_phase=1): ERK produced per-stage A,B and (if enabled) S_p.
+    // custom_update will fetch S_p via ocp_nlp_get_at_stage and add S_p Sigma_p S_p^T.
```
FreyJo, line 1358: Comment is too specific here, please remove.

interfaces/acados_template/acados_template/c_templates_tera/matlab_templates/acados_solver_sfun.in.c
```diff
@@ -1318,10 +1338,27 @@ static void mdlOutputs(SimStruct *S, int_T tid)
...
+  {% if simulink_opts.inputs.zoRO_payload == 1 %}
+    // Only compiled if the extra port was created at build time
+    if (ssGetInputPortConnected(S, {{ zoro_port_index }})) {
+      data_len = ssGetInputPortWidth(S, {{ zoro_port_index }});
+      if (data_len > 0) {
+    // Simulink guarantees this pointer is contiguous
+        c_data = (double *) ssGetInputPortRealSignal(S, {{ zoro_port_index }});
+      }
+    }
+  {% endif %}
```
FreyJo: Indent should be done such that rendered C code is correct with 4 spaces.

interfaces/acados_template/acados_template/c_templates_tera/matlab_templates/make_mex.in.m
```diff
@@ -76,6 +76,7 @@
         'acados_mex_set_{{ name }}' ...
 {%- if solver_options.custom_update_filename != "" %}
         'acados_mex_custom_update_{{ name }}' ...
+        'acados_mex_get_zoRO_Pk_{{ name }}' ...
```
FreyJo, line 80: also here, should be a different if related to zoro

interfaces/acados_template/acados_template/custom_update_templates/custom_update_function_zoro_template.in.c
```diff
@@ -1173,7 +1173,8 @@ static void uncertainty_propagate_and_update(ocp_nlp_solver *solver, ocp_nlp_in
     struct blasfeo_dmat *K_mat = &custom_mem->K_mat;
 
 {%- if zoro_description.feedback_optimization_mode != "CONSTANT_FEEDBACK" %}
-    K_mat = &custom_mem->uncertainty_matrix_buffer[0];
+    K_mat = &custom_mem->riccati_K_buffer[0];
+    // I believe uncertainty_matrix_buffer[0]; should be riccati_K_buffer[0]; to make it consistent with middle stages
```
FreyJo: Agree. Comment can be removed then. @fmesserer fyi

interfaces/acados_template/acados_template/acados_ocp_solver.py
```diff
@@ -423,6 +421,16 @@ def __init__(self, acados_ocp: Union[AcadosOcp, AcadosMultiphaseOcp, None], json
         getattr(self.shared_lib, f"{self.name}_acados_update_params").argtypes = [c_void_p, c_int, POINTER(c_double), c_int]
         getattr(self.shared_lib, f"{self.name}_acados_update_params").restype = c_int
 
+        # zoRO getter (only present for zoRO builds)
+        self.__zoro_getter = None
+        try:
```
FreyJo: Probably better to use an if on existence of `zoro_description`.

FreyJo (commented):
Looks good!
I added some minor comments.
Please revise the description removing mentioned changes to the C interface which have been removed while working on this PR.

ivandoornen:
> Please revise the description removing mentioned changes to the C interface which have been removed while working on this PR.

I think any traces have now been removed, right?

PR #1743, opened 2026-01-08
Simulink: fix example

josipkh:
This PR fixes a Simulink example in the `getting_started` folder (not tested on CI), which was failing due to a dimension mismatch.
I also slightly modified the `test_all_examples.m` script to avoid some warnings when running locally on Windows.


sandmaennchen (approved):
thanks! 👍

Issue #1209, opened 2024-08-20
Using generated CMake in Matlab on Linux


FreyJo:
On my machine, the change in #1198 causes the acados Matlab OCP examples to fail with the error:
```
Error using acados_template_mex.compile_ocp_shared_lib
Generating buildsystem failed.
Got status 1, result: cmake: /usr/local/MATLAB/R2023b/bin/glnxa64/libcurl.so.4: no version information available (required by cmake)
cmake: /usr/local/MATLAB/R2023b/sys/os/glnxa64/libstdc++.so.6: version `GLIBCXX_3.4.30' not found (required by cmake)
cmake: /usr/local/MATLAB/R2023b/sys/os/glnxa64/libstdc++.so.6: version `GLIBCXX_3.4.29' not found (required by cmake)
cmake: /usr/local/MATLAB/R2023b/sys/os/glnxa64/libstdc++.so.6: version `GLIBCXX_3.4.29' not found (required by /lib/x86_64-linux-gnu/libjsoncpp.so.25)
```
It seems to be related to specific binaries and version incompatibilities of Matlab.
Note that the CMake way works for me in Octave, although both cmake commands are invoked by a `system` call.

JD123473:
As a workaround, i renamed the library as $\texttt{libstdc++.so.6.bak}$ and everything worked just fine.

But as stated [here](https://stackoverflow.com/a/78941489), you may need to actually replace it with your system's library.

FreyJo:
It should be addressed in https://github.com/acados/acados/pull/1807 which preloads the system libraries before calling CMake.
To be merged soon.

PR #1807, opened 2026-03-25
Enable CMake for Linux MATLAB interface

h3rl:
## Enable CMake for Linux MATLAB interface

On Linux, `compile_ocp_shared_lib` and `compile_sim_shared_lib` were hardcoded to use `make`. This PR introduces a new `ACADOS_USE_CMAKE` environment variable that allows Linux users to opt into CMake instead.

### Changes

- `compile_ocp_shared_lib`: changed `if isunix` to `if isunix && ~isenv("ACADOS_USE_CMAKE")` so that when `ACADOS_USE_CMAKE` is set, the existing CMake path is used instead of `make`.
- `compile_sim_shared_lib`: same change.

### Motivation

This resolves the build failure described in #1209, where MATLAB's `LD_LIBRARY_PATH` causes CMake to pick up outdated versions of `libstdc++` and `libcurl`. With the new variable and the following `env.sh` additions, Linux users can build with CMake and benefit from `ccache` for significantly faster incremental builds:

```bash
export LD_PRELOAD=/usr/lib/x86_64-linux-gnu/libstdc++.so.6:/usr/lib/x86_64-linux-gnu/libcurl.so.4
export ACADOS_USE_CMAKE=1
export CMAKE_CXX_COMPILER_LAUNCHER=ccache
export PATH="/usr/lib/ccache:$PATH"
export CMAKE_BUILD_PARALLEL_LEVEL=$(nproc)
export CC=/usr/lib/ccache/gcc
export CXX=/usr/lib/ccache/g++
```

The default behavior (using `make`) is unchanged for users who do not set `ACADOS_USE_CMAKE`.


FreyJo:
Hi,
nice that you made this work!
I actually could reproduce the MATLAB CMake compilation on my machine.

However, if I export those variables before running MATLAB, I get in the following output in MATLAB:
```
com.jogamp.opengl.GLException: X11GLXDrawableFactory - Could not initialize shared resources for X11GraphicsDevice[type .x11, connection :1, unitID 0, handle 0x0, owner false, ResourceToolkitLock[obj 0x401c7dcf, isOwner false, <595206a, 6fb70df6>[count 0, qsz 0, owner <NULL>]]]
	at jogamp.opengl.x11.glx.X11GLXDrawableFactory$SharedResourceImplementation.createSharedResource(X11GLXDrawableFactory.java:326)
	at jogamp.opengl.SharedResourceRunner.run(SharedResourceRunner.java:297)
	at java.lang.Thread.run(Thread.java:748)
Caused by: java.lang.NullPointerException
	at jogamp.opengl.GLContextImpl.makeCurrent(GLContextImpl.java:688)
	at jogamp.opengl.GLContextImpl.makeCurrent(GLContextImpl.java:580)
	at jogamp.opengl.x11.glx.X11GLXDrawableFactory$SharedResourceImplementation.createSharedResource(X11GLXDrawableFactory.java:297)
	... 2 more
```

Also the plot of the minimal example does not render in the MATLAB GUI.
I don't know how this relates to the `exports`.
Any ideas?

Other than that, I think adding a plain MALTAB option instead of using the `ACADOS_USE_CMAKE` environment variable would be preferred.
Alternatively, it should be renamed to something more specific.

h3rl:
I believe this PR is ready for testing now. It works well on my machine, and I’ve tested the minimal example on Linux with MATLAB. That said, it would be good to do more extensive testing on other platforms before merging.

Can be tested with:
```
ocp_solver = AcadosOcpSolver(ocp, struct('force_cmake',true,'verbose',true));
sim_solver = AcadosSimSolver(ocp, struct('force_cmake',true,'verbose',true));
```

interfaces/acados_matlab_octave/AcadosSimSolver.m
```diff
@@ -311,39 +315,95 @@ function compile_mex_sim_interface_if_needed(obj)
...
+                    libcurl = strtrim(libcurl);
+
+                    preload_libs = {};
+
+                    if status_stdcpp == 0 && exist(libstdcpp, 'file') == 2
+                        preload_libs{end+1} = libstdcpp;
+                    end
+
+                    if status_curl == 0 && exist(libcurl, 'file') == 2
+                        preload_libs{end+1} = libcurl;
```
FreyJo, line 377: Just curious: Do you know why curl is needed?

FreyJo (approved):
Sorry for the late reply.

Testing on my machine and Github actions didnt reveal any issues.
I suggest refactoring the option to have a `force_make` flag instead, such that the default behavior changes to always use CMake. As shown in https://github.com/h3rl/acados/pull/1

What do you think?
Feel free to rebase and cherry-pick the two last commits of https://github.com/h3rl/acados/pull/1 instead, for a cleaner PR history.
Or do you have specific concerns why make should stay the default?

FreyJo:
Looking at this again, I think it is sensible to keep the default `make` as the preload workaround seems more fragile.
I am merging as is.
Thanks for the contribution!

PR #1822, opened 2026-04-20
Python example: solution sensitivities with parametric initial state

FreyJo:
Example can now be run in two variants.
Results are the same

### Classic x0 API
- implemented before
- compute forward sensitivities via `ocp_solver.eval_solution_sensitivity(0, with_respect_to="initial_state") `

## General parametric API
- introduce global parameter for x0 value
- allows using adjoint using `solver.eval_adjoint_solution_sensitivity(seed_u=seed_u, seed_x=None, with_respect_to="p_global")`


sandmaennchen (approved):
great!

PR #1823, opened 2026-04-23
GitHub actions: run docs workflow only in the main repo

josipkh:
Avoids failed actions and manual disabling in forks.


FreyJo (approved):
Thanks!

PR #1824, opened 2026-04-27
RTI animation python plot

FreyJo:


sandmaennchen:
I think the controls should be a step plot, not individual segments

FreyJo:
fixed :heavy_check_mark:

sandmaennchen (approved):

PR #1828, opened 2026-05-07
MATLAB: breaking refactor `AcadosOcpSolver.set()`

sandmaennchen:
- add getter for ny_total
- add flat setter for `y_ref` in MATLAB interface
- new interface to set ranges of stages to the same value

BREAKING: MATLAB interaction with `AcadosOcpSolver`: setting the same value for all stages via `ocp_solver.set(field, value);` does not work anymore. Instead specify a range of dimensions, e.g. to get the same behavior as call above, use: `ocp_solver.set('p', p_value, 0, N_horizon+1);`.
Instead calls of the form `ocp_solver.set(field, value);` will attempt to set field for all stages and assume that value contains a flattened concatenation of the corresponding field for all stages.

This also streamlines the inconsistent behaviour of  `ocp_solver.set('cost_y_ref', value);` which previously did not set the value at the terminal node. Setting  `ocp_solver.set('cost_y_ref_e', value);` now raises a deprecation warning and should be replaced by `ocp_solver.set('cost_y_ref', value, N_horizon);`

### DIFF
--- a/acados/ocp_nlp/ocp_nlp_common.c
+++ b/acados/ocp_nlp/ocp_nlp_common.c
@@ -3487,6 +3487,12 @@ int ocp_nlp_precompute_common(ocp_nlp_config *config, ocp_nlp_dims *dims, ocp_nl
             "nh", &tmp);
         dims->nh_total += tmp;
     }
+    dims->ny_total = 0;
+    for (ii = 0; ii < N+1; ii++)
+    {
+        config->cost[ii]->dims_get(config->cost[ii], dims->cost[ii], "ny", &tmp);
+        dims->ny_total += tmp;
+    }
 
     /* precompute submodules */
     // dyn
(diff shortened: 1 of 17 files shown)


FreyJo (approved):
The "breaking" documentation in the PR description, needs to be fixed before merging.
Ideally also better motivated.
Something like:

**BREAKING**: MATLAB interaction with `AcadosOcpSolver`: setting the same value for all stages via `ocp_solver.set(field, value);` does not work anymore. Instead specify a range of dimensions, e.g. to get the same behavior as call above, use: `ocp_solver.set('p', p_value, 0, N_horizon+1);`.
Instead calls of the form `ocp_solver.set(field, value);` will attempt to set field for all stages and assume that value contains a flattened concatenation of the corresponding field for all stages.

FreyJo (approved):

PR #1835, opened 2026-05-13
Python: Hessian of the value function with respect to the initial state

sandmaennchen:
The Hessian of the optimal value function with respect to the initial state corresponds to the Jacobian of the corresponding multipliers with respect to the initial state. 

What is still missing is the Hessian of the Q function wrt the initial state
second_order_derivative


FreyJo (approved):
Great! 🥇

PR #1836, opened 2026-05-20
fix(acados-template): support multi-output p_global extraction

carlos-edo:
## Summary
- support CasADi extract_parametric() returning one entry per output expression plus symbols/parametric lists
- avoid ValueError in p_global precompute generation for multi-output models

## Context
This fixes acados_template code generation for models that use p_global and generate multiple output expressions, such as the GammaMHE integration exercised from QuantKenobi on Windows/MSVC.

## Validation
- downstream validation in QuantKenobi: uv run pytest tests/test_mhe_sota.py -q
- verified the generated p_global precompute code path no longer aborts with 'too many values to unpack (expected 3)'

### DIFF
--- a/interfaces/acados_template/acados_template/acados_ocp_qp.py
+++ b/interfaces/acados_template/acados_template/acados_ocp_qp.py
@@ -273,6 +273,16 @@ class AcadosOcpQp:
             return True
         return False
 
+    def has_idxs_rev_not_idxs(self) -> bool:
+        # check if idxs_rev can be expressed via idxs
+        for stage in range(self.N + 1):
+            ns = self.__dims.ns[stage]
+            for i in range(ns):
+                idx_rev = np.where(self.idxs_rev[stage] == i)[0]
+                if len(idx_rev) > 1:
+                    return True
+        return False
+
     def has_masks(self) -> bool:
         if any([np.any(mask == 0.0) for mask in self.lbu_mask + self.ubu_mask +
                                         self.lbx_mask + self.ubx_mask +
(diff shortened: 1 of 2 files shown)


FreyJo (commented):
What is QuantKenobi?
From the description, I expect only maybe a change in:
`interfaces/acados_template/acados_template/casadi_function_generation.py`
But the code there seems to be equivalent to what was there before.
Please remove all unrelated changes and describe what is fixed with a minimal example.

carlos-edo:
I rechecked this against upstream/main with CasADi 3.7.2 and I cannot reproduce a valid bug in the current code path. In Python, extract_parametric(outputs, p_global, ...) still returns a 3-tuple (expr_ret, symbols, parametric), where expr_ret is already the list of transformed outputs. The change in this PR therefore does not describe a valid fix; with the current Python binding it actually turns expr_ret into a nested list. I have dropped the unrelated branch history and am closing this PR rather than keeping a non-minimal / non-reproducible patch open. If I can isolate a real upstream issue, I will reopen with a minimal reproducer against upstream/main.

carlos-edo:
Hi Jonathan,

Thanks for the review. You were right: the branch contained unrelated history, and the proposed change was not supported by a minimal upstream reproducer.

I rechecked the issue against upstream/main and could not justify the current patch in casadi_function_generation.py. I therefore reset the branch to upstream/main and closed the PR instead of keeping a noisy or unsupported change open.

I also built a small GenerateContext probe against upstream/main to exercise p_global together with multi-output function definitions in acados_template. With CasADi 3.7.2, I could not reproduce a failing case there, so at this point I do not have a valid minimal upstream bug report yet.

If I can isolate a real failure on top of upstream/main, I will reopen with a clean branch that contains only:
  1. a minimal reproducer,
  2. the minimal fix,
  3. and, if appropriate, a regression test.

Thanks again for flagging this.

PR #1838, opened 2026-05-22
Template: initialize numerical values for `p` and `p_global` from static const arrays

sandmaennchen:
In the ocp and sim solver template, numerical data for `p` and `p_global` is initialized from a static const array instead of assigning each nonzero value individually.

This change significantly reduces compile time for problems with lots of (global) parameters.
In the example `example_p_global.py` with `LARGE_SCALE = True`, this reduces compile time by a factor > 10.

Other
* print timings for external function generation, templated code generation, compilation (if verbose)

### DIFF
--- a/interfaces/acados_template/acados_template/acados_ocp_solver.py
+++ b/interfaces/acados_template/acados_template/acados_ocp_solver.py
@@ -194,6 +194,7 @@ class AcadosOcpSolver:
         """
         code_export_dir = os.path.abspath(code_export_dir)
 
+        t0 = time.time()
         with set_directory(code_export_dir):
             if os.name == 'nt':
                 make_cmd = 'mingw32-make'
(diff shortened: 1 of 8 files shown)


interfaces/acados_template/acados_template/acados_ocp_solver.py
```diff
@@ -210,6 +220,11 @@ def build(code_export_dir, with_cython=False, cmake_builder: CMakeBuilder = None
                     verbose_system_call([make_cmd, 'clean_ocp_shared_lib'], verbose)
                     verbose_system_call([make_cmd, 'ocp_shared_lib'], verbose)
 
+        t1 = time.time()
+
+        if verbose:
+            print(f"Build completed in {1000*(t1-t0):.3f} ms.")
+
```
FreyJo, line 229: Maybe add this in Matlab too?

FreyJo (approved):
If you have timings before and after the changes, on the considered example, would be nice to add to the PR description. Or at least a rough number.

PR #1839, opened 2026-06-02
Python/MATLAB: add check for dynamics expressions, based on the integrator type

josipkh:
If this is ok, I can add the checks in [MATLAB](https://github.com/acados/acados/blob/dc6668f8538e49831f03a36ddf755de6e92d38ea/interfaces/acados_matlab_octave/AcadosOcp.m#L986) as well


sandmaennchen:
Thanks @josipkh! The same checks for Matlab would be great 🚀

sandmaennchen (approved):
thanks!

Issue #1833, opened 2026-05-11
Python: `AcadosCasadiOcpSolver` should return 0 on success


sandmaennchen:
`solve()` and `solver.status` should return acados status values

PR #1844, opened 2026-06-09
Python: Align the return status in `AcadosCasadiOcpSolver`

Pandatheon:
This PR relates to the issue: https://github.com/acados/acados/issues/1833

- Align the return status in IPOPT to the Acados status
- Add an extra test for solver to 
     1. check the status;
     2. prepare a simple test also as a fatrop example


interfaces/acados_template/acados_template/utils.py
```diff
@@ -649,6 +649,18 @@ def status_to_str(status):
     }
     return status_dict.get(status, "UNKNOWN_STATUS")
 
+def str_to_status_ipopt(status_str):
+    str_dict = {
+        "Solve_Succeeded": 0,
+        'Solved_To_Acceptable_Level': 0,
+        "ACADOS_NAN_DETECTED": 1,
```
FreyJo: I guess, IPOPT does not return `ACADOS_NAN_DETECTED`, -> remove.
Also, use consistent '/".
Diverging iterates should maybe rather map to 6: "ACADOS_UNBOUNDED".

FreyJo (approved):

PR #1856, opened 2026-06-25
Python: fix default labels if `u` is empty

sandmaennchen:


sandmaennchen:
The minimal transition example was failing due to this. Should we add it to CI?

FreyJo (approved):
Yes, please add the example to CI

PR #1857, opened 2026-06-25
Matlab: Fix and update wind turbine ocp case

Pandatheon:


examples/acados_matlab_octave/wind_turbine_nx6/example_ocp.m
```diff
@@ -27,361 +27,270 @@
...
+         '(time_lin = %f [ms], time_qp_sol = %f [ms], time_reg = %f [ms])\n'], ...
+        status, sqp_iter, time_ext*1e3, time_tot*1e3, ...
+        time_lin*1e3, time_qp_sol*1e3, time_reg*1e3);
 
 ocp_solver.print('stat');
 
-
-% figures
-
+%% 9. Figures
```
FreyJo: I think the structuring with sections is good. But I would usually not use numbers, as minor changes may result in changing a lot of stuff.

FreyJo (approved):

PR #1865, opened 2026-07-06
Update OSQP to version 1.0

FranekStark:
Following up the discussion in https://discourse.acados.org/t/update-osqp-to-version-1-0-beta/1673/3, this PR updates OSQP to v1.0. 

In addition to minor changes resulting from renaming types, the way OSQP is interfaced with also had to be adjusted:
The old adapter never called `osqp_setup()`. To keep everything inside acados-managed memory, it hand-assembled the entire `OSQPWorkspace` (ADMM iterates, `rho_vec`, scaling, polish structures, …) into the acados memory buffer and initialized it by calling OSQP-private functions (`cold_start`, `scale_data`, `set_rho_vec`, `init_linsys_solver`, …).

In OSQP 1.0, the workspace is private and those functions are no longer exported, so the adapter now uses the supported API with the drawback of memory allocation in the first solver call:
- first solve: `osqp_setup()` (allocates internally)
- subsequent solves: `osqp_update_data_vec` / `osqp_update_data_mat` / `osqp_update_settings`
- teardown: `osqp_cleanup()` via `terminate`

If the internal memory allocation is not acceptable for the embedded use of acados, one would have to think of how to export the private functions. However, I saw that already in the old OSQP adapter the `init_linsys_solver` allocated  the KKT factorization on the heap internally.

## Verification
- `test/ocp_qp/test_qpsolvers.cpp`: all assertions pass
- `examples/c/no_interface_examples/mass_spring_example.c`: OSQP solutions identical to the HPIPM reference in the same run.
- Full `ctest` suite passes.
- Python end-to-end: `examples/acados_python/tests/test_qp_solver.py --qp_solver=PARTIAL_CONDENSING_OSQP` works
## Further points
-  `Makefile.osqp` hand-generates the 0.6 `osqp_configure.h` and compiles the 0.6 source list, so it cannot build OSQP 1.0 as-is. It is not exercised by any CI job (all CI builds use CMake) and was already default-off. Left unchanged here
-  the raw `ocp_qp` C interface (`ocp_qp_solver_destroy`) never invokes the solver's `terminate`, so solvers with internal allocations (OSQP, DAQP, Clarabel) leak here; the NLP interface calls `terminate` correctly. 
- New OSQP 1.0 features (time limit, duality-gap termination, indirect/CG solver) are not exposed as acados options yet; 

I have to say that I'm neither an expert on the internal acados code base nor an expert on OSQP. Even though the tests work for me, it would certainly be a good idea for someone else to take a closer look to the changes.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -26,7 +26,7 @@
 [submodule "external/osqp"]
 	path = external/osqp
 	url = https://github.com/osqp/osqp.git
-	branch = develop
+	branch = master
 [submodule "interfaces/acados_template/tera_renderer"]
 	path = interfaces/acados_template/tera_renderer
 	url = https://github.com/acados/tera_renderer
--- a/acados/ocp_qp/ocp_qp_osqp.c
+++ b/acados/ocp_qp/ocp_qp_osqp.c
@@ -30,6 +30,7 @@
 
 
 #include <assert.h>
+#include <string.h>
 
 // blasfeo
 #include "blasfeo_d_blasfeo_api.h"
@@ -44,14 +45,7 @@
 #include "acados/utils/types.h"
 
 // osqp
-#include "osqp/include/auxil.h"
-#include "osqp/include/constants.h"
-#include "osqp/include/glob_opts.h"
-#include "osqp/include/osqp.h"
-#include "osqp/include/scaling.h"
-#include "osqp/include/types.h"
-#include "osqp/include/util.h"
-#include "osqp/include/lin_sys.h"
+#include "osqp/include/public/osqp.h"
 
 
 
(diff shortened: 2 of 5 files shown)


jmiddelberg:
Nice work. On my machine, when building with ACADOS_WITH_OSQP and ACADOS_UNIT_TESTS, all tests are passed as well :+1:

sandmaennchen:
Hi @FranekStark thanks a lot for this contribution!

Regarding the internal memory allocation, could we move the call to `osqp_setup()` to when the solver is created? Such that we can already assume memory is allocated when calling solve?

FranekStark:
Hi @sandmaennchen, 
I am not too much an expert of the acados internal, but if I understand correctly that would mean, moving the `osqp_setup` into `ocp_qp_osqp_memory_assign`, correct?

I think this is not directly possible. `memory_assign` only knows the dimensions, but `osqp_setup` needs the sparsity pattern of the matrices. And the pattern depends on data that only arrives with the first `qp_in`: the positions of the bounds (`idxb`) and which constraints are soft (`idxs_rev`). Everything else (the RSQ, A, B, C/D blocks) is stored dense anyway, so these two are the only data-dependent parts. This is also the same behavior as before: the 0.6 adapter also allocated at the first solve (the `init_linsys_solver` call in the `first_run` branch). From the second solve on, everything is allocation-free via `osqp_update_data_*`.

Regarding the settings: the options from the problem description are set before `memory_assign`, so they would be available. Settings changed later at runtime are pushed to OSQP via `osqp_update_settings` before every solve, and all settings we currently expose are updatable in OSQP, so they work no matter when the setup runs.
Only OSQP's setup-only settings (scaling, sigma, adaptive_rho) would be frozen at setup time, but none of them are exposed in acados anyway.

FreyJo (approved):
Thanks Franek!
It does look all good :)

PR #1873, opened 2026-07-17
Python: allow calculating sensitivities w.r.t `x_0` wihout `p_global` in the model

josipkh:
It seems like it is currently not possible to calculate solution sensitivities w.r.t. initial state without defining global parameters.
I would suggest to slightly modify the checks which should enable this.

I also replaced a few `assert` commands I introduced earlier with `if-raise` statements ([motivation](https://stackoverflow.com/a/40183030)).

### DIFF
--- a/interfaces/acados_template/acados_template/acados_ocp.py
+++ b/interfaces/acados_template/acados_template/acados_ocp.py
@@ -1263,8 +1263,8 @@ class AcadosOcp:
             suffix = f", got cost_type_e {cost.cost_type_e}."
 
         if opts.with_solution_sens_wrt_params:
-            if dims.np_global == 0:
-                raise ValueError('with_solution_sens_wrt_params is only compatible if global parameters `p_global` are provided. Sensitivities wrt parameters have been refactored to use p_global instead of p in https://github.com/acados/acados/pull/1316. Got emty p_global.')
+            if dims.np_global == 0 and dims.nbx_0 == 0:
+                raise ValueError('with_solution_sens_wrt_params is only compatible if global parameters `p_global` or the initial state are provided. Sensitivities wrt parameters have been refactored to use p_global instead of p in https://github.com/acados/acados/pull/1316. Got empty p_global and initial state.')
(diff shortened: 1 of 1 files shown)


sandmaennchen (approved):

interfaces/acados_template/acados_template/acados_ocp.py
```diff
@@ -1263,8 +1263,8 @@ def make_consistent(self, mocp_info: Optional[dict]=None, verbose: bool=True) ->
             suffix = f", got cost_type_e {cost.cost_type_e}."
 
         if opts.with_solution_sens_wrt_params:
-            if dims.np_global == 0:
-                raise ValueError('with_solution_sens_wrt_params is only compatible if global parameters `p_global` are provided. Sensitivities wrt parameters have been refactored to use p_global instead of p in https://github.com/acados/acados/pull/1316. Got emty p_global.')
+            if dims.np_global == 0 and dims.nbx_0 == 0:
+                raise ValueError('with_solution_sens_wrt_params is only compatible if global parameters `p_global` or the initial state are provided. Sensitivities wrt parameters have been refactored to use p_global instead of p in https://github.com/acados/acados/pull/1316. Got empty p_global and initial state.')
```
FreyJo, line 1267: > It seems like it is currently not possible to calculate solution sensitivities w.r.t. initial state without defining global parameters.
I would suggest to slightly modify the checks which should enable this.

I am not sure why this is the case?
Do you have an example which shows that the current check is too strict?
It would be nice to add a minimal one to this PR.

In principle, computing solution sensitivities with respect to the initial state should be possible without setting `with_solution_sens_wrt_params` as `with_solution_sens_wrt_params` only enables sensitivities with respect to `p_global` and computing sensitivities with respect to x_0 is independent of that option.

josipkh: Hi Jonathan, thank you for the clarification :)
It indeed works fine if I disable `with_solution_sens_wrt_params`.
I somehow assumed that the initial state is treated as a parameter and this flag should always be enabled when calculating sensitivities...

Sorry for the confusion, I will close this PR.

PR #1874, opened 2026-07-19
Align the dense DAQP interface with current version DAQP

darnstrom:
## Summary

This PR updates the DAQP submodule from `b84fe8716` to `66c291f49` (from v0.4.2 to current DAQP master (+v0.8.6)) and adapts the acados dense DAQP interface to the current workspace, soft-constraint, and update APIs.

The main changes are:

- update workspace initialization and calls to the namespaced DAQP API
- fix `warm_start=1` active-set transfer for the current `daqp_update_ldp()` semantics
- pass a BLASFEO-computed Cholesky factor directly to DAQP
- avoid unnecessary Hessian and matrix copies on the DAQP hot-start path
- reconstruct soft slacks robustly, including for zero-row constraints
- compute dense-QP constraint slacks lazily.

## DAQP submodule update

This PR brings the version of the DAQP module to the current master.

For example, the acados interface now initializes the additional fields required by the current `DAQPProblem` and `DAQPWorkspace` structures and uses the current `DAQP_*` constants and `daqp_*` function names.

## Warm and hot starts

The current version `qp->sense` into `work->sense` when `DAQP_UPDATE_sense` is requested and activates the supplied set internally in `daqp_update_ldp()`. The previous acados interface aliased these arrays and explicitly activated constraints after the update, which no longer matches the flow of `daqp_update_ldp()`.

`DAQP_UPDATE_sense` is used for cold and normal warm starts, and the redundant explicit call to `daqp_activate_constraints()` has been removed.

For `warm_start=2`, DAQP reuses `Rinv`, `M` (which depends on the Hessian and constraint matrix), and its active-set factorization after the first matrix setup and updates only `v` and `d`. As before, this hot-start level requires the condensed Hessian and constraint matrices to remain unchanged.

## Hessian and data movement

Since newer version of DAQP supports providing a Cholesky factor of the Hessian rather than the Hessian itself. The interface now:

1. factorizes the BLASFEO Hessian with `blasfeo_dpotrf_l()`
2. packs the resulting upper Cholesky factor in DAQP's expected row-major format

The remaining dense-QP extraction is split from Hessian handling. On a valid hot start, matrix extraction skips `A` and `C` while still refreshing gradients, bounds, masks, indices, and soft-constraint data.

The interface also leaves `qp_info->t_computed = 0`, allowing the existing acados residual routines to compute constraint slacks only when requested.

### DIFF
--- a/acados/dense_qp/dense_qp_daqp.c
+++ b/acados/dense_qp/dense_qp_daqp.c
@@ -53,6 +53,8 @@
 #include "acados/utils/print.h"
 #include "acados/utils/math.h"
 
+#define DAQP_BLASFEO_MEM_ALIGNMENT 64
+
 #include "acados_c/dense_qp_interface.h"
 
 
@@ -175,7 +177,7 @@ static acados_size_t daqp_workspace_calculate_size(int n, int m, int ms, int ns)
     size += sizeof(DAQPWorkspace);
     size += sizeof(DAQPProblem);
 
-    size += n * n * sizeof(c_float); // H
+    size += (n + 1) * n / 2 * sizeof(c_float); // packed Cholesky factor of H
     size += 1 * n * sizeof(c_float); // f
     size += n * (m-ms) * sizeof(c_float); // A
     size += 2 * m * sizeof(c_float); // bupper/blower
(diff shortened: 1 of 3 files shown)


FreyJo (approved):
Looks good! Thanks :)

PR #1882, opened 2026-07-22
[DAQP] Use BLASFEO for QP-to-LDP transformation

darnstrom:
Following up on #1874, this PR simplifies and tightens the dense DAQP interface and moves the LDP pre/post-processing into BLASFEO.

## Summary

Moves the QP-to-LDP transformation and post-solution mapping from DAQP into acados interface, which allows BLASFEO to be used.

While at it, I simplified some of the existing bookkeeping code in the interface.

## Benchmark (old vs. new interface)

Measured with the same generated solver libraries, Release builds, and 1,000 solves for the pendulum case (300 for the larger quadrotor case). Times are medians in microseconds; `QP−xcond` is the DAQP/solver phase excluding condensation.

| Example | Interface | Total | Preparation | Condensation | QP−xcond |
|---|---|---:|---:|---:|---:|
| Pendulum, N=40 | old | 568 | 529 | 16 | 20 |
| Pendulum, N=40 | new | 568 | 534 | 16 | 16 |

| Example | Interface | Total | Preparation | Condensation | QP−xcond |
|---|---|---:|---:|---:|---:|
| Quadrotor, N=50 | old | 5202 | 3700 | 175 | 1317 |
| Quadrotor, N=50 | new | 4124 | 3850 | 171 | 100 |

The new interface has comparable pendulum total time, while the larger quadrotor case shows a substantially lower DAQP phase and total time. Condensation is essentially unchanged, as expected (the improvement comes from the DAQP solve/interface path rather than from xcond.)

### DIFF
--- a/acados/dense_qp/dense_qp_daqp.c
+++ b/acados/dense_qp/dense_qp_daqp.c
@@ -53,6 +53,8 @@
 #include "acados/utils/print.h"
 #include "acados/utils/math.h"
 
+#define DAQP_BLASFEO_MEM_ALIGNMENT 64
+
 #include "acados_c/dense_qp_interface.h"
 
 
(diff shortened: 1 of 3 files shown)


FreyJo (approved):
Nice work, thanks!

PR #1883, opened 2026-07-23
Python: pass verbosity level to `detect_cost_type`

josipkh:
...and suppress non-critical printing if `verbose=False` (assumes that the user is aware of the reformulation).


sandmaennchen (approved):
Thanks!

PR #1890, opened 2026-07-30
Add batch set functionality, modify example to vary dimensions easily.

FreyJo:
For the example in the benchmark setting, time in `eval_adjoint_solution_sensitivity`, I get now:sd
time reset + set seed ~1.2 ms
time C adj solve ~61 ms

Before, time reset + set seed was ~22ms


sandmaennchen (approved):
nice!

PR #1892, opened 2026-07-31
Update tera renderer to v0.2.1, in Python always attempt automatic installation.

FreyJo:
- Python: kwarg `force_download` for `get_tera` is deprecated.


sandmaennchen (approved):
Nice! Should we move all of this to when acados IS installed? At least in Python?

FreyJo:
yes, that is the plan for a follow-up.

PR #1911, opened 2026-08-31
Python: compatibility with CasADi 3.8.1

sandmaennchen:
This PR fixes the acados examples for them to be compatible with CasADi 3.8.1.

- LUT raise an error if number of knots is not a multiple of 2 for efficiency reasons.
- no more SX structs in the examples 
- fix check in `compare_struct_to_json` which previously worked for octave only

- Remove `install_new_casadi_<*>.sh` scripts. We only test with v3.8.1 now.

Note: CasADi 3.8.0 leads to segfaults when casadi+fatrop and acados are used alongside, cf: https://github.com/casadi/casadi/issues/4395


examples/acados_matlab_octave/p_global_example/create_p_global.m
```diff
@@ -41,12 +41,12 @@
     if lut
         if large_scale
             % large scale lookup table
-            knots = {0:200,0:200};
-            coefficient_vals = 0.1*ones(38809, 1);
+            knots = {0:2^8-1,0:2^8-1};
+            coefficient_vals = 0.1*ones((2^8-3)^2, 1);
         else
```
FreyJo, line 46: Do you know why this is needed? If so, please add a comment.

examples/acados_python/chain_mass/solution_sensitivity_example.py
```diff
@@ -28,26 +28,23 @@
...
 """
 Test for solution sensitivities with many parameters.
 """
 
-import os
 import numpy as np
 import casadi as ca
 from casadi import SX, norm_2, vertcat
-from casadi.tools import struct_symSX, entry
-from casadi.tools.structure3 import DMStruct, ssymStruct
```
FreyJo, line 41: I didn't see anything in the casadi release notes on those classes.
Are they removed or broken?

interfaces/acados_template/setup.py
```diff
@@ -55,7 +55,7 @@
     install_requires=[
        'numpy',
        'scipy',
-       'casadi<=3.7.2',
+       'casadi<=3.8',
```
FreyJo, line 58: It seems that not everything works yet. I suggest keeping the  `<=3.7.2` until we addressed everything.
I guess you needed it for testing.

FreyJo (approved):
A short description on the fixes in the PR description would be great.
Also one CI error about non found plot function should be fixed.
Otherwise, this looks good to me!

examples/acados_python/chain_mass/solution_sensitivity_example.py
sandmaennchen: I could not import `casadi.tools.structure3` anymore

PR #1915, opened 2026-09-06
Measured choice of qp_solver_cond_N (partial condensing horizon)

sma1laga:
# Measured choice of `qp_solver_cond_N` (partial condensing horizon)

## What

`qp_solver_cond_N` — the horizon after partial condensing — is set by hand and
defaults to `N` (no condensing). Its best value depends on `N`, `nx`, `nu` and
the machine, and is usually strictly between 1 and N. This PR adds a small
utility that chooses it by measurement, through the existing
`AcadosOcpSolver.update_qp_solver_cond_N`:

```python
solver = AcadosOcpSolver(ocp)        # PARTIAL_CONDENSING_HPIPM
solver.solve_for_x0(x0)
tune_qp_solver_cond_N(solver)        # ~5 cold-start re-solves, sets the best cond_N, restores the iterate
```

and, for loops where an offline pass is not possible, `CondNTuner` with
`before_solve()` / `after_solve()` around each solve (probes on the first
samples, then committed).

## How

A discrete unimodal (bracket) search over `{1, 2, 4, ..., N}`: ~2·log2(log2 N)
probes. Metric: QP time per SQP iteration (the only thing `cond_N` changes;
per iteration so a transient does not bias it). The current value is kept
unless a candidate is faster by more than 10 % — below that, timing noise. No
model of the machine, no tuned performance constant.

Two details that cost a debugging round each and are documented in the module:
a converged problem re-solved from its own iterate does zero SQP iterations
(nothing to time), so each probe starts from a reset; and `reset()` alone puts
the trajectory at zero, singular for models like the chain of masses, so the
cold start sets every stage to the current initial state.

## Measured (chain-of-masses example, `examples/acados_python/chain_mass`, IRK, SQP; single thread)

| nx | N | best cond_N | QP time, default/best | total step time, default/best |
|---:|---:|---:|---:|---:|
| 27 | 40 | 2 | 2.2× | 1.12× |
| 27 | 100 | 4 | 2.6× | 1.35× |
| 39 | 100 | 4 | 3.1× | 1.37× |
| 63 | 40 | full cond. | 4.3× | 1.21× |
| 63 | 100 | 8 | 3.9× | 1.25× |

The tuner's choice is within 7 % of the measured optimum on every shape
tried (three independent runs); tuning costs 0.1–2.5 s once. On the pendulum
of `test_qp_solver.py` with N = 100 the tuner picks `cond_N = 2` after five
probes: 0.26 ms per QP against 1.40 ms at the default, 5.4×. On pure OCP QPs
(no integrator) the default costs 3–13× over the best horizon.

`examples/acados_python/tests/test_cond_N_tuning.py` checks behaviour, not
timings: the chosen horizon is a candidate and is set on the solver; the
solver's iterate is left untouched; the solution at the chosen horizon equals
the solution at the default; a 100 % tolerance changes nothing; the online
tuner commits within one probe per candidate and drives a closed loop to the
same controls as the default; a full-condensing solver is refused.

## Scope

Utility only; no change to defaults or existing behaviour. Requires a
partial-condensing QP solver (`update_qp_solver_cond_N` raises otherwise).
Full condensing (`FULL_CONDENSING_HPIPM`) is a different generated solver and
is not switched to in place; where it is best (short horizons, larger nx), the
tuner's choice of `cond_N = 1` is within a few percent of it.

## Files

- `interfaces/acados_template/acados_template/cond_N_tuning.py` — the utility
- `examples/acados_python/chain_mass/tune_cond_N_example.py` — the example above
- `examples/acados_python/tests/test_cond_N_tuning.py` — behavioural test (pendulum, ~1 min)


interfaces/acados_template/acados_template/cond_N_tuning.py
```diff
@@ -0,0 +1,242 @@
...
+            if u is not None:
+                solver.set(i, 'u', u)
+
+
+def _cold_start(solver, x0):
+    """Reset the iterate, then put every stage at the current initial state.
+    A plain reset leaves the trajectory at zero, which for many models is a
+    singular configuration (the chain of masses: all masses on one point,
+    spring force NaN) and makes the SQP abort before its first QP."""
+    solver.reset()
```
sandmaennchen: here you can use `reset_x_to_x0_bar = True`, then the next to lines are not needed.

examples/acados_python/chain_mass/tune_cond_N_example.py
```diff
@@ -0,0 +1,88 @@
...
+
+if __name__ == '__main__':
+    n_mass = int(sys.argv[1]) if len(sys.argv) > 1 else 8
+    N = int(sys.argv[2]) if len(sys.argv) > 2 else 100
+    solver, integrator, x0 = build(n_mass, N)
+    steps = 25
+    default_tot, default_qp = closed_loop(solver, integrator, x0, steps)     # cond_N = N
+    solver.solve_for_x0(x0_bar=x0)
+    choice = tune_qp_solver_cond_N(solver, verbose=True)
+    tuned_tot, tuned_qp = closed_loop(solver, integrator, x0, steps)
```
sandmaennchen: here I would reset the solver before the second closed loop simulation

examples/acados_python/tests/test_cond_N_tuning.py
```diff
@@ -0,0 +1,142 @@
...
+    ocp.solver_options.nlp_solver_type = 'SQP'
+    ocp.solver_options.nlp_solver_max_iter = 200
+    # The same swing-up as test_qp_solver.py (tf = 1), only discretised finer
+    # so the horizon is long enough for condensing to matter. Stretching tf
+    # instead makes the full-step SQP cycle and hit max_iter.
+    ocp.solver_options.tf = 1.0
+    return AcadosOcpSolver(ocp, json_file=f'{name}.json'), N
+
+
+def trajectory(solver, N):
```
sandmaennchen: here you can use `AcadosOcpFlattenedIterate` instead. It directly implements an `allclose` method

examples/acados_python/tests/test_cond_N_tuning.py
```diff
@@ -0,0 +1,142 @@
...
+    return np.concatenate([solver.get(i, 'x') for i in range(N+1)]+[solver.get(i, 'u') for i in range(N)])
+
+
+def cold_start(solver, N, x0):
+    """Reset, then every stage at x0. A bare reset() leaves the iterate at zero,
+    from which the full-step SQP diverges on this problem (as it does on many)."""
+    solver.reset()
+    for i in range(N+1):
+        solver.set(i, 'x', x0)
```
sandmaennchen: here you can use reset with `reset_x_to_x0_bar = True`

sandmaennchen (commented):
Thanks a lot for this contribution, I have some suggestions to simplify a bit

examples/acados_python/tests/test_cond_N_tuning.py
sma1laga: lbx_0 set before bc the online test loop changes it

sandmaennchen (approved):

PR #1917, opened 2026-09-09
Convexify: restore Hessian symmetry before eigenvalue check

rusty-vibess:
### Summary

`blasfeo_dsyrk_ln_mn` updates only the lower triangle of `RSQrq`, while the subsequent eigenvalue check unpacks `RSQrq` as a full matrix. This can leave stale values in the upper triangle, causing the eigenvalue check to see an inconsistent Hessian and miss negative curvature that should be regularized.

This change mirrors the updated lower triangle into the upper triangle before the eigenvalue check in the full and split regularization paths.

### Test

Added a regression test which constructs a 2x2 control Hessian with one negative eigenvalue after the lower-triangular update. `CONVEXIFY` should detect this negative curvature and regularize the Hessian to a symmetric, positive-definite result.

The regression-only commit `90a825abd` fails because the upper triangle remains stale, leaving the Hessian asymmetric and indefinite. With this change, both full and split regularization pass.

I encountered this while integrating `CONVEXIFY` into an MPCC controller solved with acados SQP-RTI. I'm less familiar with the internals of the convexification module, so I'd appreciate confirmation that restoring symmetry at this point is the intended approach.


giaf:
Yes your findings are correct, symmetry should have been restored with blasfeo_dtrtr_l before calling acados_eigen_decomposition.

sandmaennchen (approved):
thanks! 👍

PR #1918, opened 2026-09-10
Identify and uniquify unused slacks "orphan slacks"

FreyJo:
## Issue
Slacks associated with no constraint (or only masked constraints), we call them *orphan slacks*, can have rather large value, especially in medium tolerance regime, while they should always be zero.

## Details
Slack stationarity condition is:
$\nabla_s \mathcal{L} = z + Z s - \lambda_n + \lambda_{s}$
where $\lambda_n$ is the multiplier of the nominal constraint and $\lambda_{s}$ the multiplier of the slack bound.
- $\lambda_n$ is exactly zero for masked constraints
- $\lambda_s$ can be nonzero depending on complementarity residual.

## OCP modification
Zero values for orphan slack can be achieved robustly by:
- masking lower bounds of orphan slacks
- setting L2 penalty to something positive, we use 1.0
- setting the L1 penalty to 0.0

The example below motivates that orphan slacks should always be handled as explained.
It corresponds to the added test run on `main` before the other changes in this PR.
image

## The new functionality:
1) OCP solver: finds orphan slacks
```
orphan_mask = zeros(2*ns)
n_ineq_nom = nb+ng+nh+nphi
for i in range(n_ineq_nom):
  is = idxs_rev[i]
  if is != -1:
    # check if nominal is not masked: lower, then upper
    if dmask[i]:
      orphan_mask[is] = 1
    if dmask[i+n_ineq_nom]:
      orphan_mask[is+ns] = 1
```

*Now:* orphan_mask[i]==0 <=> slack i is orphan.

2) Constraints: sets masks.
```
d_mask[is] = 0 if slack_bound <= -ACADOS_INFTY else orphan_mask[is]
```
NOTE: here, we don't store previous values, as masks are detected from bounds at NLP solver level.

3) Cost module: set internal slack penalties for NLP solve.
```
z_nlp = z_usr * orphan_mask
Z_nlp = Z_usr
for is in range(2*ns):
    if !orphan_mask[is]:
        Z_nlp[is] = 1.0
```

### Notes on compatibility
`SQP_WITH_FEASIBLE_QP` is where I saw potential issues, however after the following thoughts, I think that it should not be an issue.
SQP_WFQP leaves slacks defined by the user as is. The cost and constraint modules only know about those slacks.
The slacks additionally introduced by the Byrd-Omojokun formulation are not touched by the orphan detection mechanism.
This is fine, as those slacks are also not part of the NLP solution.
@david0oo as this is messing with slacks and `SQP_WITH_FEASIBLE_QP`, you might want to have a look.


FreyJo:
@sandmaennchen I did add the option `orphan_slack_handling` in the C core, but did not interface it.
I saw on some problems that the number of QP iterations was reduced by one a few times, and did never see an increase.
Given the marginal added computation cost, I would thus make orphan handling the default behavior.

I could remove the logic corresponding to having both options in C, leave it as is, or add setters for the option in the interfaces.

acados/ocp_nlp/ocp_nlp_common.c
```diff
@@ -1220,6 +1220,7 @@ void ocp_nlp_opts_initialize_default(void *config_, void *dims_, void *opts_)
     opts->nlp_qp_tol_min_eq = 1e-10;
     opts->nlp_qp_tol_min_ineq = 1e-10;
     opts->nlp_qp_tol_min_comp = 1e-11;
+    opts->orphan_slack_handling = true;
```
sandmaennchen, line 1223: Maybe add a comment here that the option is not interfaced in high level interfaces

acados/ocp_nlp/ocp_nlp_common.c
```diff
@@ -2117,6 +2127,12 @@ ocp_nlp_memory *ocp_nlp_memory_assign(ocp_nlp_config *config, ocp_nlp_dims *dims
         blasfeo_dvecse(nx[i] + nz[i], 0.0, mem->sim_guess+i, 0);
         // printf("sim_guess i %d: %p\n", i, mem->sim_guess+i);
     }
+    // orphan_mask
+    for (i = 0; i <= N; i++)
+    {
+        assign_and_advance_blasfeo_dvec_mem(2 * dims->ns[i], mem->orphan_mask + i, &c_ptr);
+        blasfeo_dvecse(2 * dims->ns[i], 1.0, mem->orphan_mask+i, 0);
+    }
```
sandmaennchen, line 2135: memory could also be created conditionally based on `orphan_slack_handling`

examples/acados_python/non_ocp_nlp/orphan_slack_test.py
```diff
@@ -0,0 +1,177 @@
...
+        plt.grid(True, which='both', ls='--', alpha=0.4)
+        plt.tight_layout()
+        plt.savefig('slack_su_tol_experiment_qp_iters.png', dpi=200)
+
+
+if __name__ == "__main__":
+    with_plots = True
+    slack_su_tol_experiment("soft_constrained_masked_irrelevant_mask_slacks", with_plots=with_plots)
+    slack_su_tol_experiment("soft_constrained_masked_irrelevant", with_plots=with_plots)
+    plt.show()
```
sandmaennchen, line 218: maybe add a comment what exactly is tested here/what the expected behaviour is

sandmaennchen (approved):
Looks good. I would leave it as is with the option, then it is easy to interface if needed

PR #1936, opened 2026-09-17
Use appropriate external function names in IRK

FreyJo:


acados/sim/sim_irk_integrator.c
```diff
@@ -190,22 +190,14 @@ int sim_irk_model_set(void *model_, const char *field, void *value)
     {
         model->impl_ode_fun = value;
     }
-    else if (!strcmp(field, "impl_ode_fun_jac_x_xdot") || !strcmp(field, "impl_dae_fun_jac_x_xdot"))
+    else if (!strcmp(field, "impl_ode_fun_jac_x_xdot_z") || !strcmp(field, "impl_dae_fun_jac_x_xdot_z") || !strcmp(field, "impl_ode_fun_jac_x_xdot") || !strcmp(field, "impl_dae_fun_jac_x_xdot"))
     {
-        // TODO(oj): remove this case and fix dependencies
+        // TODO(oj): deprecate / remove the aliases without _z.
```
apozharski, line 195: I guess deprecating for one release seems reasonable here if this is used externally by someone (which you would know better). I don't _think_ there is a way to do this in a non-obtrusive way except maybe by adding a `ACADOS_DEPRECATED` return flag (which could be a good idea).

apozharski (commented):
One comment w.r.t deprecation, otherwise the diff seems ok to me.

acados/sim/sim_irk_integrator.c
FreyJo: I don't think anyone checked the return status of this, so returning wouldn't be really helpful for outdated code.
We can just print a deprecation warning here, but I would be in favor of just removing it, too much overhead.
But this should be done later, as actually some of the generated code used the old names.

PR #1937, opened 2026-09-17
Fix: nonlinear least-squares cost integration with non-diagonal Hessian

FreyJo:


apozharski (approved):
This looks correct to me, since it was calculating `y^TL^TLy` before 😅

PR #1940, opened 2026-09-18
Factor out initialization routine from `sim_irk`

apozharski:
Unfortunately I can't make this a stacked PR with the rest of the `sim_irk` refactor because this can't be done across forks so for now these will just be annoying separate Draft PRs :upside_down_face: 

This one:
- Factor out initialization from `sim_irk` into `sim_irk_initialize` and `sim_irk_initialize_functions`
- Add `UNPACK_DIMS` macro to reduce the code-size, because each helper will need to do this.


acados/sim/sim_irk_integrator.c
```diff
@@ -1065,15 +1065,152 @@ void sim_irk_compute_z_and_algebraic_sens(sim_irk_dims *dims, sim_opts *opts, si
...
+// Note: These macros should be defined _only_ in this translation unit, so they are not prefixed.
+//       We #undef them at the end of the file in case someone is doing the silly thing of #include-ing source.
+#define UNPACK_DIMS_IRK(dims, opts) \
+int nx = dims->nx;\
+int nu = dims->nu;\
+int nz = dims->nz;\
+int np = dims->np;\
+int nf_p = opts->sens_forw_p ? np : 0;\
+int ns = opts->ns;\
+int nK = (nx + nz) * ns;\
```
FreyJo, line 1077: Does this result in unused warnings if not all of those are used in some helper?

acados/sim/sim_irk_integrator.c
```diff
@@ -1065,15 +1065,152 @@ void sim_irk_compute_z_and_algebraic_sens(sim_irk_dims *dims, sim_opts *opts, si
...
+    ws->impl_ode_type_in[2] = COLMAJ;             // u
+    ws->impl_ode_type_in[3] = BLASFEO_DVEC_ARGS;  // z_i
+    ws->impl_ode_type_in[4] = COLMAJ;							// t
+
+    ws->impl_ode_in[0] = ws->xt;								// 1st input is always xt
+    ws->impl_ode_in[1] = &ws->impl_ode_xdot_in; // 2nd input is part of K[ss],
+                                                // always update impl_ode_xdot_in
+    ws->impl_ode_in[2] = in->u;									// 3rd input is u (always)
+    ws->impl_ode_in[3] = &ws->impl_ode_z_in;    // 4th input is part of Z[ss]
+    ws->impl_ode_in[4] = &ws->t_current;				// current time
```
FreyJo: saw comment alignment was broken in previous PR, but forgot to comment.
I am not a fan of having to adjust whitespaces.
So, would prefer to defaulting to something like 2 spaces then comment.

acados/sim/sim_irk_integrator.c
```diff
@@ -1105,13 +1241,8 @@ int sim_irk(void *config_, sim_in *in, sim_out *out, void *opts_, void *mem_, vo
...
 
-    int nx = dims->nx;
-    int nu = dims->nu;
-    int nz = dims->nz;
-    int np = dims->np;
-    int nf_p = opts->sens_forw_p ? np : 0;
-
-    int nK = (nx + nz) * ns;
+    // TODO(@anton) remove when no longer necessary
+    UNPACK_DIMS_IRK(dims,opts);
```
FreyJo: I prefer space between arguments.

acados/sim/sim_irk_integrator.c
```diff
@@ -1105,13 +1241,8 @@ int sim_irk(void *config_, sim_in *in, sim_out *out, void *opts_, void *mem_, vo
...
     }
 
-    int nx = dims->nx;
-    int nu = dims->nu;
-    int nz = dims->nz;
-    int np = dims->np;
-    int nf_p = opts->sens_forw_p ? np : 0;
-
-    int nK = (nx + nz) * ns;
+    // TODO(@anton) remove when no longer necessary
```
FreyJo, line 1244: So you want to get rid of the macro again?
Because everything will be in one of the helpers?

acados/sim/sim_irk_integrator.c
```diff
@@ -1168,131 +1302,12 @@ int sim_irk(void *config_, sim_in *in, sim_out *out, void *opts_, void *mem_, vo
     int *ipiv_ss;
 
     // parameter sensitivity
-    struct blasfeo_dmat *dK_dp = NULL;
-    struct blasfeo_dmat *df_dp = NULL;
+    struct blasfeo_dmat *dK_dp = ws->dK_dp; // NOTE(@anton) this used to be guarded by an if for some reason???
```
FreyJo: Ok, if guard seems not needed. remove comment then?

acados/sim/sim_irk_integrator.c
```diff
@@ -1340,7 +1355,7 @@ int sim_irk(void *config_, sim_in *in, sim_out *out, void *opts_, void *mem_, vo
         if ( opts->sens_adj || opts->sens_hess )  // store current xn
             blasfeo_dveccp(nx, xn, 0, &xn_traj[ss], 0);
 
-	// do newton iters
+	      // do newton iters
         for (int iter = 0; iter < newton_iter; iter++)
```
FreyJo: comment seemed missaligned before, but now as well. should be indented with `for`.

acados/sim/sim_irk_integrator.c
```diff
@@ -1440,7 +1455,7 @@ int sim_irk(void *config_, sim_in *in, sim_out *out, void *opts_, void *mem_, vo
             }
         } // end newton_iter
 
-	// save k vectors
+	      // save k vectors
         if ( opts->sens_adj || opts->sens_hess )
```
FreyJo, line 1459: comment seemed missaligned before, but now as well. should be indented with `if`.

acados/sim/sim_irk_integrator.c
apozharski: It does, unfortunately. Though interestingly `CodeQL` doesn't catch this :thinking: maybe because GCC is smart enough to output that it comes from a macro? I don't know. I think my current plan of action is to do a pass at the end when the exact helpers are stabilized, and evaluate whether the macro helps readability enough to override the warnings here. I personally lean towards yes but I am happy to defer to your preferences.

acados/sim/sim_irk_integrator.c
apozharski: This todo is for the main `sim_irk` driver, so yes, ideally we won't be unpacking things here at all.

acados/sim/sim_irk_integrator.c
apozharski: Sorry, not sure what you mean. Do you prefer:

```c
while ( true )
{
 // do many things
    for (int ii = 0; ii < 10; ii++)
    {
        do_a_thing();
    }
}
```

I will say I don't know any editor which it is easy to default to this in :sweat_smile:

apozharski: Ignore the above, this was tabs sneaking in and my github settings making things look like everything is fine.

acados/sim/sim_irk_integrator.c
apozharski: One alternate way to reduce code size (which is imo, an important goal for readability) would be something like:
```c
#define NX (dims->nx)
#define NU (dims->nu)
#define NZ (dims->nz)
#define NP (dims->np)
#define NS (opts->ns)
#define NK ((NX + NZ) * NS)

// Helpers and `sim_irk` here

#undef NX
#undef NU
#undef NZ
#undef NP
#undef NS
#undef NK
```
which would remove unused warnings, but may make compiler optimizations worse as it is impossible for the compiler to prove that e.g. `dims->nx` is not aliased, and this would require a de-reference each time.

apozharski: To be clear, this is the micro-est of micro-optimizations, so don't put _too_ much weight on the performance penalty.

FreyJo (approved):

PR #1945, opened 2026-09-21
Cost integration for general nonlinear cost with ERK

FreyJo:
Support exact cost integration for ERK integrators via a new `ERK_WITH_COST` integrator type: the ODE is augmented with a cost state and the cost is integrated together with the dynamics.
In Python and MATLAB, `cost_discretization == 'INTEGRATOR'` with `ERK` and `EXTERNAL` cost now automatically reformulates to `ERK_WITH_COST`. MATLAB cost detection is refactored to allow empty (ny = 0) LLS costs, which this reformulation relies on.

**Motivation / design rationale**
- Dedicated quadrature-state treatment only offers computational benefits for IRK; for ERK with a generic nonlinear cost there is no additional structure to exploit. State augmentation is therefore the simplest and most general approach.
- The dynamics are reformulated as `f_expl_with_cost = vertcat(f_expl, cost_expr)`, and the cost is set to a `LINEAR_LS` cost with `ny = 0`: this avoids generating any cost functions and is computationally the cheapest option.
- Sensitivity handling in the augmented integrator:
  - *Nominal*: `x0_ERK = [x0, 0]`, output `x_out = x_out_ERK[:nx]`.
  - *First order*: standard forward seeds `(eye(nx_rk, nx_rk), zeros(nx_rk, nu))` yield `Sx_ERK = [[Sx, 0], [cost_grad_x, 1]]`, `Su_ERK = [[Su], [cost_grad_u]]` — the cost gradients come for free from the augmented forward sensitivities. (Possible later optimization: use `num_forw_sens = nx + nu = nx_rk - 1 + nu` to avoid computing the trivial block in `Sx_ERK`.)
  - *Second order*: `adj_seed = [adj_seed_dyn, 1]`.
- In the OCP-NLP, the cost module then avoids all computations except slacks and scaling.
- This implementation is orthogonal to the existing `cost_discretization == 'INTEGRATOR'` approach for `CONL` and `NLS` costs, where specialized routines are called inside the integrator. The two implementations coexist: specialized routines are more efficient for structured costs and IRK, but not for ERK with generic cost.

**Implementation**
- C core: Add `CONTINUOUS_WITH_COST` dynamics module `ocp_nlp_dynamics_cont_with_cost`: augments `nx` by one cost state, always computes cost Hessian sensitivities in the sim solver, registers in `ocp_nlp_config_create`.
- Python/MATLAB: new `integrator_type = 'ERK_WITH_COST'`, `AcadosOcp.reformulate_with_erk_with_cost()` converts `EXTERNAL` cost to an empty `LINEAR_LS` cost with `f_expl_expr_with_cost`, `generate_c_code_explicit_ode_with_cost_state` for CasADi codegen; refactor of `generate_c_code_explicit_ode` into shared helper. Integrator `'ERK_WITH_COST'` cannot be used as standalone, only in OCP.
- Refactor `opts_get_add_hess_contribution_ptr` into generic `opts_get` in `ocp_nlp_cost_common` (all cost modules).
- Refactor cost integration tests (`test_cost_integration_euler.py`, `test_ggn_cost_integration.py`), add multi-phase ERK cost integration test.

**Usage**
```python
ocp.solver_options.integrator_type = 'ERK'
ocp.solver_options.cost_discretization = 'INTEGRATOR'
# or direct formulation with augmented system
ocp.solver_options.integrator_type = 'ERK_WITH_COST'
ocp.model.f_expl_expr_with_cost = ca.vertcat(f_expl, cost_dynamics)
```
Requires `hessian_approx == 'EXACT'` and initial cost equal to path cost.

**Note**
Could be more efficient by avoiding one forward seed in integrator, the one corresponding to the cost state.
This will be refactored in a follow-up PR.


acados/ocp_nlp/ocp_nlp_cost_conl.c
```diff
@@ -184,7 +184,7 @@ int ocp_nlp_cost_conl_model_get(void *config_, void *dims_, void *model_,
  ************************************************/
 void ocp_nlp_cost_conl_opts_update(void *config_, void *dims_, void *opts_)
 {
-    ocp_nlp_cost_conl_opts *opts = opts_;
+    // ocp_nlp_cost_conl_opts *opts = opts_;
```
sandmaennchen: remove

acados/ocp_nlp/ocp_nlp_cost_external.c
```diff
@@ -202,7 +202,7 @@ int ocp_nlp_cost_external_model_get(void *config_, void *dims_, void *model_,
 
 void ocp_nlp_cost_external_opts_update(void *config_, void *dims_, void *opts_)
 {
-    ocp_nlp_cost_external_opts *opts = opts_;
+    // ocp_nlp_cost_external_opts *opts = opts_;
```
sandmaennchen: remove

acados/ocp_nlp/ocp_nlp_cost_ls.c
```diff
@@ -301,7 +301,7 @@ int ocp_nlp_cost_ls_model_get(void *config_, void *dims_, void *model_,
 
 void ocp_nlp_cost_ls_opts_update(void *config_, void *dims_, void *opts_)
 {
-    ocp_nlp_cost_ls_opts *opts = opts_;
+    // ocp_nlp_cost_ls_opts *opts = opts_;
```
sandmaennchen: remove

acados/ocp_nlp/ocp_nlp_sqp.c
```diff
@@ -673,7 +673,7 @@ int ocp_nlp_sqp(void *config_, void *dims_, void *nlp_in_, void *nlp_out_,
             }
         }
         // Show input to QP
-        if (nlp_opts->print_level > 3)
+        if (nlp_opts->print_level > 3+nlp_mem->iter)
```
sandmaennchen: not sure if we want to keep this change

acados/ocp_nlp/ocp_nlp_sqp.c
```diff
@@ -690,7 +690,7 @@ int ocp_nlp_sqp(void *config_, void *dims_, void *nlp_in_, void *nlp_out_,
             qp_solver->opts_set(qp_solver, nlp_opts->qp_solver_opts, "warm_start", &nlp_opts->qp_warm_start);
         }
 
-        if (nlp_opts->print_level > 3)
+        if (nlp_opts->print_level > 3+nlp_mem->iter)
```
sandmaennchen: not sure if we want to keep this change

interfaces/acados_c/ocp_nlp_interface.h
```diff
@@ -60,10 +60,10 @@ typedef enum
 } ocp_nlp_solver_t;
 
 
-/// Types of the system dynamics, discrete or continuous time.
 typedef enum
 {
     CONTINUOUS_MODEL,
+    CONTINUOUS_WITH_COST,
```
sandmaennchen: I like `CONTINUOUS_MODEL_WITH_COST`

sandmaennchen (approved):
What is the advantage of having the two usage modes?
I would have expected only this one:
```
ocp.solver_options.integrator_type = 'ERK'
ocp.solver_options.cost_discretization = 'INTEGRATOR'
```

and I think we should also test with slacks

FreyJo:
> What is the advantage of having the two usage modes? I would have expected only this one:
> ```
> ocp.solver_options.integrator_type = 'ERK'
> ocp.solver_options.cost_discretization = 'INTEGRATOR'
> ```
This is indeed the preferred "front-end" and the one demonstrated in the examples.
However, afterwards we need to change the model / formulation in `make_consistent` and need to bring it into an equivalent state, that is dumped to `json` and then used in the templates.
Moreover, we want to be able to load this reformulated OCP back and get equivalent results.
Thus, we also need to have this as an alternative supported front-end. 


> and I think we should also test with slacks

Is already done in `examples/acados_python/tests/test_cost_integration_euler.py`

